PR #4, opened 2016-10-07
MCR-1393 added .iml (intellij module library) to .gitignore

sebhofmann:


.gitignore
```diff
@@ -16,6 +16,9 @@ hs_err_pid*
 *.iws
 # IntelliJ
 /out/
+# mycore addition
+*/*.iml
```
yagee-de: *.iml ought to be enough as .gitignore is also valid in sub directories

yagee-de (approved):

PR #26, opened 2016-10-19
MCR-1408 add redirect to MCRBasketServlet

eckstueck:
- added possibility to send a URL as redirect parameter to MCRBasketServlet

### DIFF
--- a/mycore-base/src/main/java/org/mycore/frontend/basket/MCRBasketServlet.java
+++ b/mycore-base/src/main/java/org/mycore/frontend/basket/MCRBasketServlet.java
@@ -137,6 +137,8 @@ public class MCRBasketServlet extends MCRServlet {
         }
         if (referer != null && "referer".equals(redirect)) {
             res.sendRedirect(res.encodeRedirectURL(referer.toExternalForm()));
+        } else if (redirect != null) {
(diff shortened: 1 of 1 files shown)


mycore-base/src/main/java/org/mycore/frontend/basket/MCRBasketServlet.java
```diff
@@ -135,8 +135,13 @@ public void doGetPost(MCRServletJob job) throws Exception {
...
             return;
         }
-        if (referer != null && "referer".equals(redirect)) {
-            res.sendRedirect(res.encodeRedirectURL(referer.toExternalForm()));
+        if (referer != null) {
+            if ("referer".equals(redirect)) {
+                res.sendRedirect(res.encodeRedirectURL(referer.toExternalForm()));
+            }
+            else {
+                res.sendRedirect(res.encodeRedirectURL(redirect));
```
yagee-de: If a referrer is given but the redirect parameter is empty. What happens?
If the redirect parameter is given but no referrer is available. What happens?

yagee-de (changes requested):

eckstueck:
squashed all commits to a single commit
:shipit: ?

yagee-de (approved):

PR #28, opened 2016-10-20
MCR-1410 fix classification editor save not working

Mewel:


yagee-de:
Target branch must be [mycore-2016.06.0.x](/MyCoRe-Org/mycore/tree/mycore-2016.06.0.x)

PR #40, opened 2016-10-24
MCR-1416 no-chapter-destination pdf chapter loading now is more tough

sebhofmann:


yagee-de:
please open pull-request against LTS branch as target

PR #56, opened 2016-11-11
MCR-1428 Exception while delete in SVN store

bt1cu:
Ignore exceptions while delete in SVN store and write a logging information.


yagee-de (approved):
Code formatting seems to be wrong. Approved this time. Congratulations to your first pull request!

PR #62, opened 2016-11-14
MCR-1433 transform-xml-derivates added command to transform xml docum…

sebhofmann:
…ents which are located in derivates


yagee-de:
Can you please adjust the log message for this commit? Subject is too long

yagee-de (approved):

PR #63, opened 2016-11-15
MCR-1329 AutoDeploy implement JAR autodeploy feature

adlerre:


yagee-de (changes requested):
please fix code style

yagee-de (changes requested):
why is  "mycore-base/src/main/java/org/mycore/datamodel/ifs2/MCRVersionedMetadata.java" changed?

yagee-de (approved):

PR #76, opened 2016-11-18
MCR-1444 use yarn instead of npm to build WebCLI

eckstueck:
- replaced npm with yarn as package manager


yagee-de:
please open pull request against LTS branch...

PR #78, opened 2016-11-21
MCR-1446 Use Yarn instead of npm to build MyCoRe-Viewer

sebhofmann:


yagee-de:
implemented by pr #87

PR #95, opened 2016-11-30
MCR-1460 IFS checks throw more IOExceptions

yagee-de:
which is a good thing, as we already urge the user to catch
IOException. They should not check for 'null' then...


yagee-de:
please review if this fixes [MCR-1460](https://mycore.atlassian.net/browse/MCR-1460).

kkrebs (approved):
2016-12-01T06:56:42,228 INFO  administrator MCRWebCLIContainer: Command processed (54887224 ms)

looks good! Both commands now work.

PR #97, opened 2016-12-01
MCR-1462 MCRThumbnailServlet reuse filesystem to prevent deadlock

sebhofmann:


mycore-iview2/src/main/java/org/mycore/iview2/frontend/MCRThumbnailServlet.java
```diff
@@ -193,7 +193,7 @@ private BufferedImage getThumbnail(Path iviewFile, boolean centered) throws IOEx
             Path iviewFileRoot = fs.getRootDirectories().iterator().next();
             MCRTiledPictureProps props = MCRTiledPictureProps.getInstanceFromDirectory(iviewFileRoot);
             if (props.getZoomlevel() == 0) {
-                return MCRIView2Tools.getZoomLevel(iviewFile, 0);
+                return MCRIView2Tools.getZoomLevel(fs, 0);
```
yagee-de: go with iviewFileRoot instead to match other methods.
Signature would not change but JavaDoc saying that Path has to be a directory (iviewFileRoot) if caller holds a ZipFileSystem open, if it is a file it is handled like a ZIP-File

mycore-iview2/src/main/java/org/mycore/iview2/services/MCRIView2Tools.java
```diff
@@ -206,9 +206,29 @@ public static boolean isTiled(MCRPath file) {
      * @throws JDOMException
      *             if image properties could not be parsed.
      */
-    public static BufferedImage getZoomLevel(Path iviewFile, int zoomLevel) throws IOException, JDOMException {
-        ImageReader reader = getTileImageReader();
+    public static BufferedImage getZoomLevel(Path iviewFile, int zoomLevel)
+        throws IOException, JDOMException {
```
yagee-de: First check if iviewFile is a directory, handle it that way. Else open FileSystem and call the same method again with the root directory of that filesystem...

yagee-de (changes requested):

yagee-de (approved):

PR #101, opened 2016-12-06
MCR-1465 Git Improvements for MCRCoreVersion

yagee-de:
Use git-commit-id-plugin to get some Git properties useful for MCRCoreVersion.
This commit changes output of main() and return value of getCompleteVersion()


yagee-de:
do not merge yet, will add more code...

PR #124, opened 2017-01-03
MCR-1483 Rename permalink to direct link

sebhofmann:


yagee-de:
must go onto 2016.06.0.x

sebhofmann:
In 2016.06.0.x there is no mycore-viewer/src/main/resources/components/viewer/config/messages_de.properties or any viewer code.

Need two different PR to fix this. This is the one for master.

yagee-de (approved):

PR #128, opened 2017-01-04
MCR-1496 Change deprecated properties behaviour

yagee-de:
throw Exception if a deprecated property is used


fluetze (approved):
Ok for me

PR #135, opened 2017-01-06
MCR-1498 isTiled and isCompletelyTiled should use the Database

sebhofmann:


mycore-iview2/src/main/java/org/mycore/iview2/services/MCRIView2Tools.java
```diff
@@ -23,7 +23,7 @@
 
 package org.mycore.iview2.services;
 
-import java.awt.Graphics;
+import java.awt.*;
```
yagee-de: restore explicit import

mycore-iview2/src/main/java/org/mycore/iview2/services/MCRJobState.java
```diff
@@ -23,6 +27,14 @@
      */
     ERROR('e');
 
+    public final static Set<MCRJobState> NOT_COMPLETE_STATES = Stream
+        .of(MCRJobState.ERROR, MCRJobState.PROCESSING, MCRJobState.NEW).collect(
+            Collectors.toSet());
```
yagee-de: makes the Set modifiable which is not good for a static Set

yagee-de (changes requested):

mycore-iview2/src/main/java/org/mycore/iview2/services/MCRTileJob.java
```diff
@@ -17,7 +19,13 @@
  *
  */
 @Entity
-@Table(name = "MCRTileJob", uniqueConstraints = @UniqueConstraint(columnNames = { "derivate", "path" }) )
+@Table(name = "MCRTileJob", uniqueConstraints = {
+    @UniqueConstraint(columnNames = { "derivate", "path" }),
+    @UniqueConstraint(columnNames = { "derivate", "status" })
```
yagee-de: Derivate, Status is not a "unique" constraint. You want a normal Index here.

yagee-de (changes requested):

yagee-de (approved):

PR #136, opened 2017-01-06
MCR-1500 A PIRegistrationService should check permission register_serviceid

sebhofmann:


mycore-pi/src/main/java/org/mycore/pi/MCRPIRegistrationService.java
```diff
@@ -27,6 +25,8 @@
 import org.mycore.pi.backend.MCRPI_;
 import org.mycore.pi.exceptions.MCRPersistentIdentifierException;
 
+import static org.mycore.access.MCRAccessManager.PERMISSION_WRITE;
```
yagee-de: please put static imports before regular imports (like in eclipse) to keep format the same

yagee-de (approved):

PR #139, opened 2017-01-06
MCR-1472 Travis-Firefox-Fix

yagee-de:
readded firefox-esr and limit memory for maven to 512 MB


yagee-de:
firefox-esr will never do the tests within memory limit

PR #140, opened 2017-01-09
MCR-1501 rework test case to get a positive result on all machine speeds

adlerre:


mycore-base/src/test/java/org/mycore/services/queuedjob/MCRJobQueueTest.java
```diff
@@ -47,7 +47,9 @@ public void testOffer() throws InterruptedException {
         assertTrue("job should be offered", queue.offer(job));
 
         endTransaction();
-        Thread.sleep(1000);
+        synchronized (queue) {
+            queue.wait(1000);
```
yagee-de: It is nearly the same as nobody calls `queue.notify()`. You need to wait for the job you offered earlier to finish and wait for that event a maximum time, e.g. 30 seconds. One second is to less for Travis-ci to finish reliable.

yagee-de (changes requested):

yagee-de (approved):

PR #149, opened 2017-01-13
MCR-1511 MCRDOIRegistrationService should validate generated xml

sebhofmann:


kkrebs (approved):

yagee-de (changes requested):
messages_*.properties are invalid

yagee-de (dismissed):

PR #150, opened 2017-01-16
MCR-1485 CLI command "set parent of ..." creates relatedItem entry in mods without object ID

sebhofmann:


kkrebs (approved):
please add documentation for repair stylesheet usage, perhaps into ticket comment?

yagee-de (approved):

PR #192, opened 2017-03-08
MCR-1551 Added property to select which menu should be shown on the l…

sebhofmann:
…eft side

### DIFF
--- a/mycore-iview2/src/main/java/org/mycore/iview2/frontend/configuration/MCRViewerBaseConfiguration.java
+++ b/mycore-iview2/src/main/java/org/mycore/iview2/frontend/configuration/MCRViewerBaseConfiguration.java
@@ -55,6 +55,11 @@ public abstract class MCRViewerBaseConfiguration extends MCRViewerConfiguration 
             setProperty("canvas.startup.fitWidth", true);
         }
 
+        String leftShowOnStart = MCRConfiguration.instance().getString("MCR.Viewer.leftShowOnStart", null);
+        if(leftShowOnStart!=null){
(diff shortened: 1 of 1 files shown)


kkrebs (approved):

mycore-iview2/src/main/java/org/mycore/iview2/frontend/configuration/MCRViewerBaseConfiguration.java
```diff
@@ -55,6 +55,11 @@ public MCRViewerConfiguration setup(HttpServletRequest request) {
             setProperty("canvas.startup.fitWidth", true);
         }
 
+        String leftShowOnStart = MCRConfiguration.instance().getString("MCR.Viewer.leftShowOnStart", null);
+        if(leftShowOnStart!=null){
```
yagee-de: setProperty() should handle null values to eliminate all "if (value != null)" checks

yagee-de, line 59: setProperty() should handle null values to eliminate all "if (value != null)" checks

yagee-de (dismissed):

mycore-iview2/src/main/java/org/mycore/iview2/frontend/configuration/MCRViewerBaseConfiguration.java
sebhofmann: setProperty can be used to build json configuration structure like { "test" : null } and if i put a if check in the setProperty then this is not possible anymore.

PR #227, opened 2017-04-25
MCR-1593 get date issued from rdf (pborchert)

kkrebs:


PaulBorchert (changes requested):
The changes doesn't fit to the Issue. The code transform the issue-period of an journal.

PR #230, opened 2017-04-26
MCR-1480 Missing children in parent structure

yagee-de:
added migration command 'add missing children' to fix children in parent


mycore-migration-2015/src/main/java/org/mycore/migration2015/cli/MCRMigrationCommands.java
```diff
@@ -1,48 +1,62 @@
 package org.mycore.migration2015.cli;
```
kkrebs: shouldn't this migration command be part of 2017 commands?

yagee-de: the fix goes into the LTS but we can rename mycore-migration-2015 to mycore-migration in master with another ticket

kkrebs (approved):

PR #248, opened 2017-05-16
MCR-724 switch to token based querying

Mewel:


mycore-oai/src/main/java/org/mycore/oai/MCROAIDeletedSearcher.java
```diff
@@ -0,0 +1,123 @@
...
+        DeletedRecordPolicy deletedRecordPolicy = this.identify.getDeletedRecordPolicy();
+        if (from == null || DeletedRecordPolicy.No.equals(deletedRecordPolicy)
+            || DeletedRecordPolicy.Transient.equals(deletedRecordPolicy)) {
+            return new ArrayList<>();
+        }
+        LOGGER.info("Getting identifiers of deleted items");
+        try {
+            List<String> deletedItems = MCRDeletedItemManager.getDeletedItems(from, Optional.ofNullable(until));
+            List<String> types = getConfig().getStrings(getConfigPrefix() + "DeletedRecordTypes", null);
+            if (types == null) {
```
yagee-de: shouldn't this be: `type == null || types.isEmpty()`

mycore-oai/src/main/java/org/mycore/oai/MCROAIDeletedSearcher.java
```diff
@@ -0,0 +1,123 @@
...
+            }
+            return deletedItems.stream()
+                               .map(MCRObjectID::getInstance)
+                               .filter(id -> {
+                                   return types.contains(id.getTypeId());
+                               })
+                               .map(MCRObjectID::toString)
+                               .collect(Collectors.toList());
+        } catch (Exception ex) {
+            LOGGER.warn("Could not retrieve identifiers of deleted objects", ex);
```
yagee-de: Error are not handled by logging them. Exceptions should be thrown until response code != 200 is given to the client.

yagee-de (changes requested):

PR #255, opened 2017-05-18
MCR-1609 advance computing of yearIssued

PaulBorchert:


mycore-mods/src/main/resources/xsl/mods-solr.xsl
```diff
@@ -115,15 +115,27 @@
         <xsl:value-of select="text()" />
       </field>
     </xsl:for-each>
-    <xsl:for-each select="mods:originInfo[not(@eventType) or @eventType='publication']/mods:dateIssued">
+    <xsl:for-each select="mods:originInfo[not(@eventType) or @eventType='publication']/mods:dateIssued[encoding='w3cdft']">
```
kkrebs: encoding is an attribute

kkrebs (changes requested):
see comment

kkrebs (approved):

PR #280, opened 2017-06-13
MCR-1560 allow empty categories in classification

Mewel:


kkrebs (changes requested):
can you please fix this in LTS2016.06

PR #301, opened 2017-06-19
MCR-1312 use property for mods object type

kkrebs:


rsteph-de (approved):
works like discussed earlier, good job

PR #307, opened 2017-06-19
MCR-1648 use property for DOI resolving and change URL to https://doi…

eckstueck:
….org/

- added property MCR.DOI.Resolver.MasterURL


mycore-mods/src/main/resources/xsl/modsmetadata.xsl
```diff
@@ -6,6 +6,7 @@
   exclude-result-prefixes="xalan xlink mcr mcrxsl i18n acl mods mcrmods rdf"
   version="1.0">
   <xsl:param name="MCR.Handle.Resolver.MasterURL" />
+  <xsl:param name="MCR.DOI.Resolver.MasterURL" />
```
kkrebs, line 9: property is not defined in mycore
should have a default

kkrebs (changes requested):

kkrebs (approved):

PR #373, opened 2017-08-23
MCR-1699-XEditor support for java post processors

sebhofmann:
MCR-1699-XEditor support for java post processors


fluetze (changes requested):
Die Methode getDefaultRepeaterImplementation() sollte getDefaultPostProcessorImplementation() heissen.
Das Property sollte MCR.XEditor.PostProcessor.Default heissen.

PR #434, opened 2017-11-08
MCR-1747 MCRLoginServlet should use protected functions

AMill7:
change to protected on needed places, presentLoginForm can now be
overwritten


mycore-user2/src/main/java/org/mycore/user2/login/MCRLoginServlet.java
```diff
@@ -76,9 +76,9 @@
 
     private static final String LOGIN_REDIRECT_URL_KEY = "loginRedirectURL";
 
-    static final boolean LOCAL_LOGIN_SECURE_ONLY = MCRConfiguration.instance().getBoolean(HTTPS_ONLY_PROPERTY);
+    protected static final boolean LOCAL_LOGIN_SECURE_ONLY = MCRConfiguration.instance().getBoolean(HTTPS_ONLY_PROPERTY);
 
-    private static Logger LOGGER = LogManager.getLogger();
+    protected static Logger LOGGER = LogManager.getLogger();
```
yagee-de: LOGGER should stay 'private'

yagee-de (changes requested):

PR #439, opened 2017-11-14
MCR-1745 Fix exception when expanding macro references. Upgrade to latest BibTeX parser version.

fluetze:
references. Upgrade to latest BibTeX parser version.


yagee-de (changes requested):
Ticket should be MCR-1745

PR #441, opened 2017-11-15
MCR-1752 add default stylesheet step only if output method is xml

yagee-de:
if mime-type is "application/pdf" (XSL:FO) do also not run extra layout step


Possommi:
seems to work

Possommi (approved):

chiliego (approved):

PR #446, opened 2017-11-20
MCR-1743 add MCRUserManager.getUsers(String, String)

yagee-de:
a method to get users by a given user attribute

### DIFF
--- a/mycore-user2/src/main/java/org/mycore/user2/MCRUser.java
+++ b/mycore-user2/src/main/java/org/mycore/user2/MCRUser.java
@@ -1,5 +1,5 @@
 /**
- * $Revision$ 
+ * $Revision$
  * $Date$
  *
  * This file is part of the MILESS repository software.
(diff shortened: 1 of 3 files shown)


fluetze (commented):
Why do you need to filter the results afterwards?

return propertyQuery.getResultList()
+            .stream()
+            .filter(u -> u.getAttributes().get(attrName).equals(attrValue));

yagee-de:
There is no possibility to query a specific attribute in JQL like `* WHERE name=:name and value=:value`. You can simply not access 'name' and 'value' as such in any query. JQL gives you just the possibility to look into all names with `KEY(user.attributes)` and all values by `VALUE(user.attributes)`. For a typical scenario like `orcid=0000-0001-5065-6970` there is a high possibility that `0000-0001-5065-6970` is always an ORCID but in a more general scenario for attributes:

- a=b
- c=d

We want to make sure that a query for a=d will not return that user, as the JQL query would return it.

fluetze (approved):

PR #451, opened 2017-11-24
Issues/mcr 1657 cleanup

yagee-de:


Mewel (commented):
Looks good. Runs in ZS.

just see MCRSwordMediaHandler#init line 263 does not need the configuration. And line 128 can be removed.

PR #452, opened 2017-11-28
MCR-1105 fixed "change log level ..." command

yagee-de:


mycore-base/src/main/java/org/mycore/frontend/cli/MCRLoggingCommands.java
```diff
@@ -32,28 +30,21 @@
      *            level
      */
     @MCRCommand(syntax = "change log level of {0} to {1}",
-        help = "{0} the package or class name for which to change the log level, {1} the log level to set. If the log level cannot be read it is set to DEBUG by default.",
+        help = "{0} the package or class name for which to change the log level, {1} the log level to set.",
```
kkrebs, line 33: perhaps we can add here 2 or 3 examples (ERROR, WARN, INFO ...)

kkrebs (approved):
works fine
I only had problems with findig correct log level string. In log4j2.xml we wrote "info" or "error" but I learned, that log4j log level are case-sensitive. I don't know why it works in log4j2.xml, but for this command I have to use "INFO", "WARN", "DEBUG" etc ...

PR #478, opened 2018-01-03
MCR-1762 filter for Rest-API output

OPNA2608:
Bitte schaut mal, ob das okay ist.
Danke.


mycore-restapi/src/main/java/org/mycore/restapi/v1/utils/MCRRestAPIObjectsHelper.java
```diff
@@ -143,14 +154,40 @@ public static Response showMCRObject(String pathParamId, String queryParamStyle,
             }
         }
 
+        String xslfile = MCRConfiguration.instance().getString("MCR.RestAPI.v1.Filter.XML", "");
```
yagee-de: Please use [MCRContentTransformerFactory](http://www.mycore.de/generated/mycore-2017.06.0/mycore-base/apidocs/org/mycore/common/content/transformer/MCRContentTransformerFactory.html) to do transformations for better performance and more flexibility

mycore-restapi/src/main/resources/components/restapi/config/mycore.properties
```diff
@@ -9,6 +9,9 @@ MCR.RestAPI.v1.Files.URL.path=servlets/MCRFileNodeServlet/${derid}/
 MCR.RestAPI.v1.Filter.Write.IPs.Pattern=127\\.0\\.0\\.1|::1
 MCR.RestAPI.v1.Filter.Read.IPs.Pattern=\\.*
 
+# For filtering xml metadata set this property with your filter stylesheet
+# MCR.RestAPI.v1.Filter.XML=xsl/mycoreobject-xml.xsl
```
yagee-de: `MCR.RestAPI.v1.Filter.XML=mycoreobject-xml` would use the mycoreobject-xml Content-Transformer (if defined)

yagee-de (changes requested):

mycore-restapi/src/main/java/org/mycore/restapi/v1/utils/MCRRestAPIObjectsHelper.java
```diff
@@ -146,11 +152,21 @@ public static Response showMCRObject(String pathParamId, String queryParamStyle,
         StringWriter sw = new StringWriter();
         XMLOutputter outputter = new XMLOutputter(Format.getPrettyFormat());
         try {
-            outputter.output(doc, sw);
+            if (MCRConfiguration.instance().getString("MCR.ContentTransformer.RestApiXmlFilter.Stylesheet", "").length() > 0) {
```
yagee-de: Do not hardcode any transformer configuration here. It's better to make a filter property that defines a ContentTransformer like in my original review.
Simply add something like `MCR.RestAPI.v1.Filter.XML=RestApiXmlFilter` and use only `MCR.RestAPI.v1.Filter.XML` in your code to detect IF there is a content transformer defined.
ContentTransfomer MAY use Stylesheets but can be of any kind...

mycore-restapi/src/main/java/org/mycore/restapi/v1/utils/MCRRestAPIObjectsHelper.java
```diff
@@ -146,11 +152,21 @@ public static Response showMCRObject(String pathParamId, String queryParamStyle,
...
+                MCRContentTransformer trans = MCRContentTransformerFactory.getTransformer("RestApiXmlFilter");
```
yagee-de: use the value of `MCR.RestAPI.v1.Filter.XML` here and do not hard code a specific transformer here

yagee-de (changes requested):

yagee-de (approved):

kkrebs (approved):

PR #482, opened 2018-01-12
 MCR-1785

sebhofmann:
TODO: find more improvements


MyCoRe-Travis:
Something gone wrong on 9ea5b0c1ca5f7f1de3a4ac635cfbfdf93ef51296! 
 Please Check https://github.com/MyCoRe-Travis/test_artifacts/tree/2634-pull_request-482

mycore-pi/src/main/java/org/mycore/pi/doi/MCRDOIRegistrationService.java
```diff
@@ -71,10 +70,12 @@
 
     private static final int MAX_URL_LENGTH = 255;
 
-    private static final String DATACITE_SCHEMA_PATH = "xsd/datacite/metadata.xsd";
+    private static final String DEFAULT_DATACITE_SCHEMA_PATH = "xsd/datacite/v3/metadata.xsd";
```
kkrebs, line 73: meanwhile we have version 4. Perhaps we should not put this specification in java code? (could be changed in mycore master ...)

mycore-pi/src/main/java/org/mycore/pi/MCRPIRegistrationService.java
```diff
@@ -43,7 +43,7 @@
 
     public static final String GENERATOR_CONFIG_PREFIX = "MCR.PI.Generator.";
 
-    public static final String METADATA_MANAGER_CONFIG_PREFIX = "MCR.PI.Inscriber.";
+    public static final String METADATA_MANAGER_CONFIG_PREFIX = "MCR.PI.MetadataManager.";
```
kkrebs, line 47: Is this also part of deprecated properties? We should prevent property renaming within an LTS version ...

sebhofmann: It was renamed pre LTS but i forgot to rename this constant. See also:      
`public static final String METADATA_MANAGER_DEPRECATED_CONFIG_PREFIX = "MCR.PI.Inscriber.";`

But that does not matter because everything is backward compatible (You just get a log message).

rsteph-de (commented):
Looks good for my part.
I would suggest to move MCRGsonUTCDateAdapter to mycore-base, since it is not mycore-pi specific?

kkrebs (approved):

rsteph-de (approved):

PR #484, opened 2018-01-16
MCR-1783 Remove whitespace after classname properties

bt1cu:
Add new methods getClassOf to get classes via configuration properties and integrate it in methods of getInstnceOf . Replace class calls in used classes.

### DIFF
--- a/mycore-base/src/main/java/org/mycore/common/config/MCRConfiguration.java
+++ b/mycore-base/src/main/java/org/mycore/common/config/MCRConfiguration.java
@@ -361,38 +361,64 @@ public class MCRConfiguration {
     }
 
     /**
+     * Returns a new class object specified in the configuration property with the given name.
+     * 
+     * @param name
+     *            the non-null and non-empty qualified name of the configuration property
+     * @param defaultname
+     *            the qualified default class name
+     * @return Class of the value of the configuration property
+     * @throws MCRConfigurationException
+     *             if the property is not set or the class can not be loaded or instantiated
+     */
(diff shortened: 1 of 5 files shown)


kkrebs (approved):

mycore-base/src/main/java/org/mycore/common/config/MCRConfiguration.java
```diff
@@ -361,38 +361,64 @@ public MCRProperties getDeprecatedProperties() {
...
      * @param defaultname
-     *            the qualified class name
-     * @return Instance of the value of the configuration property
+     *            the qualified default class name
+     * @return Class of the value of the configuration property
      * @throws MCRConfigurationException
      *             if the property is not set or the class can not be loaded or instantiated
      */
-    public <T> T getInstanceOf(String name, String defaultname) throws MCRConfigurationException {
+    public Class<?> getClassOf(String name, String defaultname) throws MCRConfigurationException {
```
yagee-de, line 374: the default value should be a Class object like getLong, getBoolean, etc.

yagee-de (changes requested):

mycore-base/src/main/java/org/mycore/common/config/MCRConfiguration.java
yagee-de: maybe naming should also be changed to getClass instead of getClassOf
the method with the default value may not throw a MCRConfigurationException as the default value is provided

yagee-de:
Will be moved to master

PR #485, opened 2018-01-18
MCR-1787 encode classification id

sebhofmann:


kkrebs (commented):
this should also be fixed in 2017er branch

kkrebs (approved):

PR #488, opened 2018-01-25
MCR-1790 update to SolrJ 7.2.1

yagee-de:


Mewel (approved):
Looks good, jportal runs w/o any problems.

PR #489, opened 2018-01-25
MCR-1783 added MCRConfiguration.getClass() methods

yagee-de:
used to get Class objects from property values.


Mewel (approved):
Code looks good, works in jportal too.

PR #490, opened 2018-01-25
MCR-1791 update to JQuery 3.3.1

yagee-de:


kkrebs (changes requested):
WebCLI can not be used with this update! Clicking on Commands-DropDown leads to en empty page

kkrebs (approved):

PR #502, opened 2018-02-14
MCR-1802 remove redundant SOLR code

yagee-de:
remove MCRSolrFileSizeStrategy as it makes no sense in production


Possommi (commented):
Seems to work as good as it did before ; )

Mewel (approved):
Works with jportal.

PR #504, opened 2018-02-14
resolve solr search_result_link_text field length with mycore property

rochpaul:
-> Set previous fixed length as default


rochpaul:
Resolve into lts branch instead of master!

PR #519, opened 2018-02-27
MCR-1716 improve processing gui performance

Mewel:
-enable production mode
-update gui only every 20ms if something changed


yagee-de (approved):
no lags detected in test environment anymore

PR #530, opened 2018-03-15
MCR-1820 renamed MCR.inject.module.* to MCR.Inject.Module.*

yagee-de:


chiliego:
The refactoring seems to be OK. But there may be some improvements in the logic of retrieving the instances like

```java
MCRConfiguration
   .instance()
   .getPropertiesMap("MCR.Inject.Module")
   .values()
   .stream()
   .map(MCRConfiguration2::instantiateClass)
   .map(Module.class::cast)
   .collect(Collectors.toList());
```

You can avoid the roundtrip of retrieving the value of the properties again, because it's already done by getPropertiesMap()
"Module.class::cast" can be wrapped into a method and generate a nice error message (maybe class XYZ is not a Module etc.) in case of a class cast exception.

Mewel (approved):
Works in jportal.

PR #545, opened 2018-04-03
MCR-1504 improve solr index strategy

Mewel:
-check if object/derivate is marked for deletion -> do not index object
-check if owner of path is marked for deletion -> do not index path


Possommi (commented):
this fix seems to resolve the issue

PR #563, opened 2018-05-02
MCR-1850 fix viewer mobile view

sebhofmann:


Mewel (approved):
Works with jportal and android phone.

PR #570, opened 2018-05-03
MCR-1855 replace rfc4646 by rfc5646 and include all languages

eckstueck:
changed rfc4646 to rfc5646


kkrebs (commented):
seems that this changes need some migrations to be done, can you add migration steps to https://github.com/MyCoRe-Org/mycore/blob/master/mycore-migration/README.md

yagee-de (approved):

PR #584, opened 2018-05-16
MCR-1849 set increment before log output

bt1cu:
move increment before log output

### DIFF
--- a/mycore-solr/src/main/java/org/mycore/solr/index/statistic/MCRSolrIndexStatistic.java
+++ b/mycore-solr/src/main/java/org/mycore/solr/index/statistic/MCRSolrIndexStatistic.java
@@ -45,15 +45,13 @@ public class MCRSolrIndexStatistic {
     }
 
     public long addTime(long time) {
-        accumulatedTime.addAndGet(time);
         LOGGER.debug("{}: adding {} ms", name, time);
-        return accumulatedTime.get();
+        return accumulatedTime.addAndGet(time);
     }
 
     public int addDocument(int docs) {
-        documents.addAndGet(docs);
-        LOGGER.debug("{}: adding {} documents", name, documents);
-        return documents.get();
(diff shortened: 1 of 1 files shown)


mycore-solr/src/main/java/org/mycore/solr/index/statistic/MCRSolrIndexStatistic.java
```diff
@@ -45,13 +45,15 @@ public MCRSolrIndexStatistic(String name) {
     }
 
     public long addTime(long time) {
+        accumulatedTime.addAndGet(time);
         LOGGER.debug("{}: adding {} ms", name, time);
-        return accumulatedTime.addAndGet(time);
+        return accumulatedTime.get();
```
yagee-de: should be addAndGet like before to be an atomic operation

mycore-solr/src/main/java/org/mycore/solr/index/statistic/MCRSolrIndexStatistic.java
```diff
@@ -45,13 +45,15 @@ public MCRSolrIndexStatistic(String name) {
...
     }
 
     public int addDocument(int docs) {
+        documents.addAndGet(docs);
         LOGGER.debug("{}: adding {} documents", name, documents);
-        return documents.addAndGet(docs);
+        return documents.get();
```
yagee-de: should be addAndGet like before to be an atomic operation

yagee-de (commented):
Will have to look into MCR-1849. Your changes should result in the same return value, as by definition of 'addAndGet()'

MyCoRe-Travis:
Something gone wrong on fa80a4a6265ae32598e9e90ee12e813540315492! 
 Please Check https://github.com/MyCoRe-Travis/test_artifacts/tree/3271-pull_request-584

kkrebs (approved):

PR #591, opened 2018-05-17
MCR-1864 ids are not shown per default in classification editor

kkrebs:
[Link to MCR-1864](https://mycore.atlassian.net/browse/MCR-1864).


Mewel (approved):
Looks right.

PR #595, opened 2018-05-24
MCR-1868 MCRStoreIFS2 thread safe

yagee-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-1868).


fluetze (approved):
Looks good for me

PR #603, opened 2018-06-04
MCR-1877 fixed searchfield mods.publisher

kkrebs:
[Link to MCR-1877](https://mycore.atlassian.net/browse/MCR-1877).


PaulBorchert (changes requested):
1. Perhaps it's better to get only 1 publisher b object. 2. Gets the publisher from type=series. 3. Get the publisher from grandparent.

PaulBorchert (approved):

PR #604, opened 2018-06-05
MCR-1878-Toolbar should resize if more space needed

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-1878).


rsteph-de (approved):
solves the problem

PR #619, opened 2018-06-26
MCR-1892 Remove workaround to find user by name identifier

fluetze:

### DIFF
--- a/mycore-orcid/src/main/java/org/mycore/orcid/works/MCRWorkEventHandler.java
+++ b/mycore-orcid/src/main/java/org/mycore/orcid/works/MCRWorkEventHandler.java
@@ -1,8 +1,8 @@
 package org.mycore.orcid.works;
 
 import java.util.HashSet;
-import java.util.Optional;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import org.apache.logging.log4j.LogManager;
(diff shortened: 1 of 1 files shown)


mycore-orcid/src/main/java/org/mycore/orcid/works/MCRWorkEventHandler.java
```diff
@@ -81,9 +82,10 @@ private void publishToORCID(MCRObjectID oid, MCRORCIDUser user) {
             String name = MCRORCIDUser.ATTR_ID_PREFIX + key.split(":")[0];
             String value = key.split(":")[1];
 
-            // Workaround, because MCRUserManager.getUsers(name,value) returns users with incomplete attributes
-            MCRUserManager.listUsers(null, null, null).stream()
-                .filter(u -> value.equals(u.getUserAttribute(name))).forEach(u -> users.add(u));
+            Optional<MCRUser> user = MCRUserManager.getUsers(name, value).findFirst();
```
yagee-de: isn't `users.addAll(MCRUserManager.getUsers(name, value).collect(Collectors.toSet()))` better?

works even if multiple user have the same attribute value - just in case

yagee-de (changes requested):

kkrebs (approved):

PR #623, opened 2018-06-26
MCR-1694 add command to rename derivate files with reg ex

kkrebs:
[Link to MCR-1694](https://mycore.atlassian.net/browse/MCR-1694).


Mewel (changes requested):
Can you add javadoc to the class and its methods. Its public API so it should be described.
How does the pattern look like, how the replacement, and an example would be nice too.

kkrebs:
@eckstueck please rebase branch, resolve conflicts and add requested changes

yagee-de (approved):

PR #626, opened 2018-06-26
MCR-1824 add support for X-MCR-Ignore-Message-Body header

yagee-de:
required to create directories from brwoser

[Link to jira](https://mycore.atlassian.net/browse/MCR-1824).


yagee-de:
will add a fix to get baseURL correctly soon

PR #628, opened 2018-06-26
MCR-1879 fixed adding images

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-1879).


kkrebs (approved):

kkrebs (changes requested):
can you please fix put this fix in LTS2017?

kkrebs (approved):

kkrebs:
seems to be a bug only in 2018.05 ... jira description was wrong here

PR #631, opened 2018-06-27
MCR-1896 fixed listIds* methods in MCRXMLMetadataManager

yagee-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-1896).


Mewel (approved):
looks good

PR #640, opened 2018-07-05
MCR-1899 _derivate_000 in file name causes Viewer to fail

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-1899).


rsteph-de (approved):
makes sense

PR #650, opened 2018-07-13
MCR-1910 delete tmp files after upload

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-1910).


Possommi (approved):
Files are deleted from the temporary directory once uploaded

PR #651, opened 2018-07-16
MCR-1911 close InputStream if SOLR doesn't do it

yagee-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-1911).


Mewel (approved):
Seems ok.

PR #659, opened 2018-08-01
MCR-1904 fixed SOLR expert search

yagee-de:
added core selection for predefined aliases 'main' and 'classification'

[Link to jira](https://mycore.atlassian.net/browse/MCR-1904).

### DIFF
--- a/mycore-solr/src/main/resources/META-INF/resources/search/search-expert.xed
+++ b/mycore-solr/src/main/resources/META-INF/resources/search/search-expert.xed
@@ -121,6 +121,19 @@ form .optional.expanded .fieldset{
                 </h3>
               </div>
               <div class="panel-body">
+                <xed:bind xpath="param[@name='core']" default="main">
+                  <div class="form-group">
+                    <label class="col-md-{$label-width} control-label" for="core">
+                      core
+                    </label>
+                    <div class="col-md-{$input-width}">
+                      <select title="SOLR core (alias)." class="form-control" id="core">
+                        <option>main</option>
+                        <option>classification</option>
+                      </select>
+                    </div>
+                  </div>
+                </xed:bind>
                 <mcrsolr:textfield name="qt" title="Request handler in solrconfig.xml." default="/select" />
 
                 <fieldset class="common">
--- a/mycore-solr/src/main/resources/META-INF/resources/search/search-expert.xed
+++ b/mycore-solr/src/main/resources/META-INF/resources/search/search-expert.xed
@@ -121,19 +121,7 @@ form .optional.expanded .fieldset{
                 </h3>
               </div>
               <div class="panel-body">
-                <xed:bind xpath="param[@name='core']" default="main">
-                  <div class="form-group">
-                    <label class="col-md-{$label-width} control-label" for="core">
-                      core
-                    </label>
-                    <div class="col-md-{$input-width}">
-                      <select title="SOLR core (alias)." class="form-control" id="core">
-                        <option>main</option>
-                        <option>classification</option>
-                      </select>
-                    </div>
-                  </div>
-                </xed:bind>
+                <mcrsolr:textfield name="core" title="SOLR core (alias, e.g. 'main' or 'classification')." default="main" />
                 <mcrsolr:textfield name="qt" title="Request handler in solrconfig.xml." default="/select" />
 
                 <fieldset class="common">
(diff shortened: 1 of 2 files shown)


mycore-solr/src/main/resources/xsl/solr-xeditor-templates.xsl
```diff
@@ -192,10 +194,23 @@
...
+      <xsl:with-param name="type" select="'classification'" />
+    </xsl:apply-templates>
+  </xsl:template>
+  <xsl:template match="mcrsolr:fieldsHelp" mode="with-core">
+    <xsl:param name="core" />
+    <xsl:param name="type" />
     <div class="table-responsive">
       <table class="table table-striped table-hover table-condensed">
-        <xsl:variable name="url" select="concat($MCR.Solr.ServerURL, '/admin/luke')" />
+        <xsl:variable name="url" select="concat($MCR.Solr.ServerURL, 'solr/', $core, '/admin/luke?wt=xml')" />
```
kkrebs: It is possible to specify a ServerURL for each Core, MCR.Solr.ServerURL is only the fallback if no URL for core is specified

mycore-solr/src/main/resources/META-INF/resources/search/search-expert.xed
```diff
@@ -121,6 +121,19 @@ form .optional.expanded .fieldset{
...
+                      core
+                    </label>
+                    <div class="col-md-{$input-width}">
+                      <select title="SOLR core (alias)." class="form-control" id="core">
+                        <option>main</option>
+                        <option>classification</option>
+                      </select>
+                    </div>
+                  </div>
+                </xed:bind>
```
kkrebs: I'm not sure if this is the best solution. It is user friendly, but filters flexibility. We can specify more core types (not only classification and core). If so, we have to add this here ...

kkrebs (changes requested):

mycore-solr/src/main/resources/META-INF/resources/search/search-expert.xed
yagee-de: this will not work for an undefined set of core aliases but only for a fixed set. a 'real' expert has to change the URL parameter to any other core. No field names and types will be fetched in either case.

mycore-solr/src/main/resources/xsl/solr-xeditor-templates.xsl
yagee-de: I'll take a look on this

mycore-solr/src/main/resources/META-INF/resources/search/search-expert.xed
yagee-de: another possibility would be to replace the select box with an input field that defaults to 'main'

kkrebs: This sounds good for me. In the most use cases "main" will be used I think.

MyCoRe-Travis:
Something gone wrong on 1a1673afa27505581594a48ad5d4170c752ffdcb! 
 Please Check https://github.com/MyCoRe-Travis/test_artifacts/tree/3763-pull_request-659

mycore-solr/src/main/resources/xsl/solr-xeditor-templates.xsl
```diff
@@ -2,6 +2,10 @@
 <xsl:stylesheet version="1.0" xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:xed="http://www.mycore.de/xeditor"
   xmlns:mcrsolr="http://www.mycore.de/components/solr" xmlns:mcrxml="xalan://org.mycore.common.xml.MCRXMLFunctions">
   <xsl:param name="MCR.Solr.ServerURL" />
+  <xsl:param name="MCR.Solr.Core.main.Name" />
+  <xsl:param name="MCR.Solr.Core.main.ServerURL" />
+  <xsl:param name="MCR.Solr.Core.classification.Name" />
+  <xsl:param name="MCR.Solr.Core.classification.ServerURL" />
```
kkrebs, line 8: Now searchmask is flexible but this template still only works for "classification" or "main". This code have also to be that flexible or we can switch back to drop down menu in xed.

yagee-de:
accepted for now as of Rostock meeting

PR #665, opened 2018-08-15
MCR-1924 use permissions as defined by MCRAccessManager

yagee-de:
ACL permissions used by the REST API are the same as for objects, derivates and files

[Link to jira](https://mycore.atlassian.net/browse/MCR-1924).


sebhofmann (approved):

rsteph-de:
This MUST be documented and is relevant for migration !!!

PR #670, opened 2018-08-21
MCR-1904 fix solr schema and config issues

kkrebs:
[Link to jira](https://mycore.atlassian.net/browse/MCR-1904).


yagee-de (approved):
multiValued: "false" will be addressed at a later time

PR #677, opened 2018-08-27
merge 2017.06.0.x

kkrebs:


rsteph-de (commented):
Wouldn't it be better to comply with DSGVO and stability to link copies of the images within our application instead of embedding them from external sources?

kkrebs:
Of course it is! We should make image url part of licence classification. But this needs some code refactoring and should be done in master branch. This should only be a hot fix to prevent mixed content on https pages.

PR #682, opened 2018-08-30
MCR-1937 fixed for metadata api

bt1cu:
[Link to jira](https://mycore.atlassian.net/browse/MCR-1937).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/common/MCRObjectMerger.java
+++ b/mycore-base/src/main/java/org/mycore/common/MCRObjectMerger.java
@@ -81,11 +81,15 @@ public class MCRObjectMerger {
                         }
                     }
                     if (!equal) {
-                        metaElementTarget.addMetaObject(metaInterfaceSource.clone());
-                        if (validate && !validate(this.target)) {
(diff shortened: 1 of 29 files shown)


sebhofmann (changes requested):
The MCRMetaInterface and all Implementations are violating the java clone convetions. We should fix this.

1. All clone methods should throw a CloneNotSupportedExcetion
2. If you implement a clone method, then you always should call super.clone() and the extend the inherited object

Why we have a clone method in MCRMetaInterface which has a different signatur then the original clone method in object?

bt1cu:
MCRMetaInterfache throw now the CloneNotSupportedExcetion as default of clone().

The super.clone() is not required cause we use the XML-create-parse way to get a complete clone. These methods have a super() implicit. This seems the best way to get a correct clone of all data by the same way.

sebhofmann:
JavaDoc of Object.clone says: By convention, the returned object should be obtained by calling super.clone.

sebhofmann:
Creating clones by the XML-create-parse way does not obey Java Convention and has also a big Overhead.

sebhofmann (approved):

kkrebs (approved):

PR #704, opened 2018-10-16
MCR-1960 fixed MCRSession leak in /rsc/* resources

yagee-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-1960).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/frontend/jersey/filter/MCRSessionHookFilter.java
+++ b/mycore-base/src/main/java/org/mycore/frontend/jersey/filter/MCRSessionHookFilter.java
@@ -31,6 +31,7 @@ import javax.ws.rs.container.ContainerResponseContext;
 import javax.ws.rs.container.ContainerResponseFilter;
 import javax.ws.rs.core.Context;
 
+import org.apache.commons.io.output.ProxyOutputStream;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.mycore.common.MCRSession;
@@ -47,11 +48,15 @@ public class MCRSessionHookFilter implements ContainerRequestFilter, ContainerRe
     @Context
     private HttpServletResponse httpResponse;
 
+    private final static String ATTR = MCRSessionHookFilter.class.getName() + ".session";
+
     private static final Logger LOGGER = LogManager.getLogger(MCRSessionHookFilter.class);
 
     @Override
-    public void filter(ContainerRequestContext request) throws IOException {
+    public void filter(ContainerRequestContext request) {
+        MCRSessionMgr.unlock();
         MCRSession session = MCRServlet.getSession(httpRequest);
+        request.setProperty(ATTR, session);
         MCRSessionMgr.setCurrentSession(session);
         LOGGER.info(MessageFormat.format("{0} ip={1} mcr={2} user={3}", request.getUriInfo().getPath(),
             MCRFrontendUtil.getRemoteAddr(httpRequest), session.getID(), session.getUserInformation().getUserID()));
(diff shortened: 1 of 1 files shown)


sebhofmann:
Can you explain how a session leak?         
MCRSessionMgr.releaseCurrentSession(); should handle that.
Creating a extra property with session seems not necessary, because we already have the mycore.session attribute.

yagee-de:
> Can you explain how a session leak?

Session leak may occur if we release a session like we did and a MessageBodyWriter handles the output transformation, e.g. POJO --> JSON, and requires a MCRSession.

> MCRSessionMgr.releaseCurrentSession(); should handle that.
> Creating a extra property with session seems not necessary, because we already have the mycore.session attribute.

It's an attribute for the request so that we can check at the end of the request if the current session is the same as original session. A MCRSession is only closed if it is bound to a HttpSession. We close it in the responsefilter as it would leak.

mycore-base/src/main/java/org/mycore/frontend/jersey/filter/MCRSessionHookFilter.java
```diff
@@ -47,20 +48,68 @@
...
+            LOGGER.debug("No Entity in response, closing MCRSession");
+            releaseSessionIfNeeded(requestSession);
+        }
+    }
+
+    private static void releaseSessionIfNeeded(MCRSession requestSession) {
+        if (MCRSessionMgr.hasCurrentSession()) {
+            MCRSession currentSession = MCRSessionMgr.getCurrentSession();
+            try {
+                if (currentSession.isTransactionActive()) {
```
kkrebs, line 93: shouldn't this be part of MCRSessionMgr.releaseCurrentSession?

mycore-base/src/main/java/org/mycore/frontend/jersey/filter/MCRSessionHookFilter.java
```diff
@@ -47,20 +48,68 @@
...
+                        currentSession.rollbackTransaction();
+                    } else {
+                        currentSession.commitTransaction();
+                    }
+                } else {
+                    LOGGER.debug("Active MCRSession found. Clearing up");
+                }
+            } finally {
+                MCRSessionMgr.releaseCurrentSession();
+                if (!currentSession.equals(requestSession)) {
```
kkrebs, line 105: why can a jersey session be not bound to an HTTP session? I seems that we fix here symptoms and not the root cause if it is so.

kkrebs (changes requested):
I see this as well as paschty. Also for me it seems to make the code more complicated ...

mycore-base/src/main/java/org/mycore/frontend/jersey/filter/MCRSessionHookFilter.java
yagee-de: It could be implemented that way that transaction are automatically finished when a session is released (or closed). This should get another ticket and would be part of other code changes. That would definitely make the code simply to read by hiding boilerplate code. We have to to think carefully of any side-effects this may have, but I think there are none.

But at current state this code is handling corner cases better than before and cannot be written simpler - suggestions are welcome.

mycore-base/src/main/java/org/mycore/frontend/jersey/filter/MCRSessionHookFilter.java
yagee-de: > why can a jersey session be not bound to an HTTP session? I seems that we fix here symptoms and not the root cause if it is so.

There are 3 phases and this class handles 2.

1. An MCRSession (A) is optained or created and bound to HttpSession and stored in the current request. When the HttpSession hits EOL the MCRSession is closed, too.
1. Other request or response filters take over and the resource code is executed.
1. The response is filtered and we check if session A is still used or was switched in phase 2. If that is the case the session will leak as it is not bound to the HttpSession. We close it for that reason.

mycore-base/src/main/java/org/mycore/frontend/jersey/filter/MCRSessionHookFilter.java
yagee-de: It _**could**_ be part of releaseCurrentSession() and would simplify this code a lot. This change is out of scope of this ticket and should be an extra one as this would include changes in other parts. We should carefully think about side-effects though. Currently, I think there are none. But definitely a good point we should keep in mind.

sebhofmann (approved):

kkrebs (approved):

kkrebs:
ok, I created a new ticket -> https://mycore.atlassian.net/browse/MCR-1982

PR #711, opened 2018-10-26
MCR-1967 add /api/v2/classifications/{classid}/{categid}

yagee-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-1967).


sebhofmann (changes requested):
Everything works. 
In my opinion the returned structure should be centered to arround the requested category.
The root element should be the category. The category should have children and parents.
The way it now works is more like a filter, instead of request the category. We should discuss further.

yagee-de:
It uses the same structure already known from the URIResolver where parent category are parent elements/objects. I see no problem in filtering the complete classification while maintaining the structure. This makes implementation easier as the same classes/structure can be used as for the complete classification.

kkrebs (approved):
We should discuss this on our next developer meeting. For details see:
https://cmswiki.rrz.uni-hamburg.de/hummel/MyCoRe/Umsetzung/RestApiV2

PR #715, opened 2018-11-02
MCR-1971 added distinctValues function for text nodes

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-1971).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/common/xml/MCRXMLFunctions.java
+++ b/mycore-base/src/main/java/org/mycore/common/xml/MCRXMLFunctions.java
@@ -40,13 +40,16 @@ import java.text.Normalizer;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.Date;
 import java.util.List;
 import java.util.ListIterator;
 import java.util.Locale;
 import java.util.Optional;
 import java.util.Set;
+import java.util.SortedSet;
 import java.util.TimeZone;
+import java.util.TreeSet;
 import java.util.concurrent.CancellationException;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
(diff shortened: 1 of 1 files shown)


mycore-base/src/main/java/org/mycore/common/xml/MCRXMLFunctions.java
```diff
@@ -1031,4 +1035,48 @@ public static String toNCNameSecondPart(String name) {
         return name.replaceAll("[^\\w\\-\\.]*", "");
     }
 
+    /**
+     * This only works with text nodes
```
yagee-de, line 1040: javadoc is no help. It should clearly state what is done with which input. If it's only working with text nodes, throw an Exception if a node is not of that type.

mycore-base/src/main/java/org/mycore/common/xml/MCRXMLFunctions.java
```diff
@@ -1031,4 +1035,48 @@ public static String toNCNameSecondPart(String name) {
...
+     * @param nodes
+     * @return
+     */
+    public static NodeList distinctValues(NodeList nodes) {
+        SortedSet<Node> set = new TreeSet<>(new Comparator<Node>() {
```
yagee-de: What is the resulting order? Is it unchanged?

mycore-base/src/main/java/org/mycore/common/xml/MCRXMLFunctions.java
```diff
@@ -1031,4 +1035,48 @@ public static String toNCNameSecondPart(String name) {
...
+            @Override public int compare(Node node, Node t1) {
```
yagee-de: variable naming

yagee-de (changes requested):

mycore-base/src/main/java/org/mycore/common/xml/MCRXMLFunctions.java
```diff
@@ -1031,4 +1035,48 @@ public static String toNCNameSecondPart(String name) {
...
+
+                String nodeValue = node.getNodeValue();
+                String nodeValue1 = t1.getNodeValue();
+
+                if (nodeValue != null && nodeValue1 != null) {
+                    return nodeValue.equals(nodeValue1) ? 0 : -1;
```
yagee-de: skip to many if clauses:
```suggestion
                    return (nodeValue == nodeValue1 || nodeValue.equals(nodeValue1)) ? 0 : -1;
```

mycore-base/src/main/java/org/mycore/common/xml/MCRXMLFunctions.java
sebhofmann: I don't want to throw a exception, because you can not handle them in XSL.

yagee-de (approved):

PR #725, opened 2018-11-13
MCR-1978 add a netmask test to check the correct session IP

bt1cu:
[Link to jira](https://mycore.atlassian.net/browse/MCR-1978).


yagee-de (changes requested):
Works for IPv4 and will probably fail if mycore is using IPv6. Maybe we should use a customizable prefix length for IPv4 (defaults to 32) and IPv6 (defaults to 128)?

bt1cu:
should fix in LTS 2019

PR #727, opened 2018-11-22
MCR-1981 MCRXEditorTransformerStoreUtils is not needed

eckstueck:
removed MCRXEditorTransformerStoreUtils

[Link to jira](https://mycore.atlassian.net/browse/MCR-1981).


kkrebs:
I think this should also be reviewed from fluetze

fluetze (approved):
Looks right for me

PR #738, opened 2018-12-10
MCR-1993 fix metadata format for single records

yagee-de:
also return empty record list if all entries on a list are not available
with the requested metadata format

[Link to jira](https://mycore.atlassian.net/browse/MCR-1993).


kkrebs (approved):
works fine!

PR #759, opened 2019-01-29
MCR-2015 update dependencies

yagee-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2015).


yagee-de:
tested with mir-it

PR #764, opened 2019-02-07
MCR-2017 lock MCRSessionMgr by default

yagee-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2017).


sebhofmann (approved):
Could not find any Session Exception in MIR.

PR #769, opened 2019-02-14
MCR-2020 main file is not set after mods file import

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2020).

### DIFF
--- a/mycore-mods/src/main/java/org/mycore/mods/MCRMODSCommands.java
+++ b/mycore-mods/src/main/java/org/mycore/mods/MCRMODSCommands.java
@@ -20,10 +20,13 @@ package org.mycore.mods;
 
 import java.io.File;
 import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.text.MessageFormat;
 import java.util.Arrays;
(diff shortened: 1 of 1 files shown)


mycore-mods/src/main/java/org/mycore/mods/MCRMODSCommands.java
```diff
@@ -178,9 +182,17 @@ private static MCRDerivate createDerivate(MCRObjectID documentID, File fileDir)
         linkId.setReference(documentID, null, null);
         derivate.getDerivate().setLinkMeta(linkId);
 
+        final Predicate<Path> isDirectory = Files::isDirectory;
+        final Path rootPath = fileDir.toPath();
+        final Optional<String> firstRegularFile = Files.list(rootPath).filter(isDirectory.negate())
```
yagee-de: why not use [`Files::isRegularFile`](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#isRegularFile-java.nio.file.Path-java.nio.file.LinkOption...-l)?

yagee-de (changes requested):

PR #781, opened 2019-03-01
MCR-2019 Viewer endButton and startButton causes endless scroll

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2019).


rochpaul (approved):
tested with opera 58.0.3135.79; Edge, Internet Explorer 11, Firefox (Everything seems to be okay, there is no automatic scroll anymore)

PR #788, opened 2019-03-06
MCR-2039 sometimes Viewer does not load properly in embedded mode

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2039).


kkrebs (approved):
tested in production

PR #789, opened 2019-03-12
MCR-2040 better Italian translation

bt1cu:


kkrebs (approved):
looks fine (with greetings from our Italian colleague)

PR #796, opened 2019-04-10
MCR-2043 override resources in webapp resolver

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2043).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/common/config/MCRConfigurationDir.java
+++ b/mycore-base/src/main/java/org/mycore/common/config/MCRConfigurationDir.java
@@ -22,21 +22,19 @@ import java.io.File;
 import java.io.IOException;
 import java.net.MalformedURLException;
 import java.net.URL;
-import java.nio.file.Files;
 import java.nio.file.Path;
-import java.nio.file.Paths;
 import java.security.CodeSource;
 import java.security.ProtectionDomain;
 import java.util.Enumeration;
 import java.util.List;
 import java.util.Locale;
 import java.util.Optional;
-import java.util.stream.Stream;
 
 import javax.servlet.ServletContext;
 
 import org.apache.logging.log4j.LogManager;
 import org.mycore.common.MCRClassTools;
+import org.mycore.common.MCRDeveloperTools;
 
 /**
  * This helper class determines in which directory to look for addition configuration files.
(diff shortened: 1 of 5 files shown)


mycore-base/src/main/java/org/mycore/common/config/MCRConfigurationDir.java
```diff
@@ -214,20 +212,11 @@ public static URL getConfigResource(String relativePath) {
...
-            final String[] uris = override.split(",");
-            final Optional<Path> resource = Stream.of(uris).map(Paths::get)
-                .filter(Files::exists)
-                .map(p -> p.resolve(relativePath))
-                .filter(Files::exists)
-                .peek(p -> LogManager.getLogger().info("Found overridden File in path: " + p.toAbsolutePath()))
-                .findFirst();
-
-            if (resource.isPresent()) {
+        if (MCRDeveloperTools.overrideActive()) {
```
chiliego, line 215: When MCRDeveloperTools.overrideActive() is true it will be called inside MCRDeveloperTools.getOverriddenFilePath a second time, maybe there is a better way.

mycore-base/src/main/java/org/mycore/common/config/MCRConfigurationDir.java
```diff
@@ -214,20 +212,11 @@ public static URL getConfigResource(String relativePath) {
...
+            final Optional<Path> overriddenFilePath = MCRDeveloperTools.getOverriddenFilePath(relativePath, false);
+            if (overriddenFilePath.isPresent()) {
                 try {
-                    return resource.get().toUri().toURL();
+                    return overriddenFilePath.get().toUri().toURL();
                 } catch (MalformedURLException e) {
                     // Ignore
                 }
```
chiliego, line 222: I would using the full potential of Optionals in this way

```

public static URL getConfigResource(String relativePath, ClassLoader classLoader) {
        return MCRDeveloperTools .getOverriddenFilePath(relativePath, false)
                .or(() -> getResolvedFile(relativePath))
                .map(p -> pathToURL(p))
                .orElseGet(() -> getClassPathResource(relativePath, classLoader == null ?
                        MCRClassTools.getClassLoader() : classLoader));
}

private static Optional<Path> getResolvedFile(String relativePath){
     return Optional.ofNullable(getConfigFile("resources/" + relativePath))
                .filter(File::exists)
                .map(File::toPath);
}

private static URL pathToURL(Path path){
     try {
         return path.toUri().toURL();
     } catch (MalformedURLException e) {
         LogManager.getLogger(MCRConfigurationDir.class)
                 .warn("Exception while returning URL for file: {}", path.toString(), e);
     }
     return null;
}

```

chiliego (approved):

PR #800, opened 2019-04-23
MCR-2051 fixed i18n method in mets-editor

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2051).


Possommi (approved):
works like a charm

PR #838, opened 2019-08-15
MCR-1792 recreate tei xsd files

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-1792).


bt1cu (approved):

yagee-de:
xml*.xsd and xlink*.xsd are already part of mycore-base...

sebhofmann:
> xml*.xsd and xlink*.xsd are already part of mycore-base...

The resolution of http://www.w3.org/XML/2008/06/xlink.xsd -> xlink-2008-06.xsd does not work if the catalog and xsd are in mycore-base

yagee-de:
> > xml*.xsd and xlink*.xsd are already part of mycore-base...
> 
> The resolution of http://www.w3.org/XML/2008/06/xlink.xsd -> xlink-2008-06.xsd does not work if the catalog and xsd are in mycore-base

It does work, as you can see in mycore-mods. But we can fix this later...

PR #840, opened 2019-08-16
MCR-2088 Fixed NPE in case WEB-INF/classes/xsl/ is not used

fluetze:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2088).


kkrebs (approved):
successfully tested with MIR

PR #842, opened 2019-08-16
MCR-2041 replaces bs3 validation css classes

TiloNN:
Replaced outdated bs3 validatation css classes by new mcr classes.
The dawn of a new age: MyCoRe has its own validation classes.
The liberation from bootstrap?!

[Link to jira](https://mycore.atlassian.net/browse/MCR-2041).


kkrebs (changes requested):
if we specify not bootstrap classes here, we have to add a css declaration in this mycore components also

kkrebs (approved):

kkrebs:
I added .mcr-invalid and .mcr-valid css classes as workaround to acl editor css. I also added more information to our documentation. We have to refacture our html here and use html5 features for client side validation.

PR #845, opened 2019-08-21
MCR-2091 epub viewer does not work in ie

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2091).


kkrebs (approved):
tested with ie11

PR #847, opened 2019-08-28
MCR-2043 override static xml files

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2043).


fluetze (commented):
Looks good for me, but:
Please check co-operation with MCRStaticXEditorFileServlet, which extends this servlet.
MCRStaticXEditorFileServlet overrides method
  protected MCRContent getResourceContent(..)
Probably the change there is not necessary any more, there are two places where the override is done.

fluetze (approved):

PR #852, opened 2019-09-11
MCR-2101 added public List<MCRJob> getJobs(Map<String, String> params…

Possommi:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2101).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/services/queuedjob/MCRJobQueue.java
+++ b/mycore-base/src/main/java/org/mycore/services/queuedjob/MCRJobQueue.java
@@ -366,7 +366,7 @@ public class MCRJobQueue extends AbstractQueue<MCRJob> implements Closeable {
 
         EntityManager em = MCREntityManagerProvider.getCurrentEntityManager();
 
-        StringBuilder qStr = new StringBuilder("FROM MCRJob job JOIN FETCH job.parameters WHERE action = '"
+        StringBuilder qStr = new StringBuilder("FROM MCRJob job WHERE action = '"
                 + action.getName() + "' ");
         for (String paramKey : params.keySet()) {
             qStr.append(" AND job.parameters['")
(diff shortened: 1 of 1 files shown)


mycore-base/src/main/java/org/mycore/services/queuedjob/MCRJobQueue.java
```diff
@@ -349,6 +349,44 @@ private MCRJob getJob(Class<? extends MCRJobAction> action, Map<String, String>
...
+        EntityManager em = MCREntityManagerProvider.getCurrentEntityManager();
+
+        StringBuilder qStr = new StringBuilder("FROM MCRJob job WHERE action = '"
+                + action.getName() + "' ");
+        for (String paramKey : params.keySet()) {
+            qStr.append(" AND job.parameters['")
+                    .append(paramKey)
+                    .append("'] = '")
+                    .append(params.get(paramKey))
+                    .append('\'');
```
adlerre: should be double quotes

adlerre (changes requested):

mycore-base/src/main/java/org/mycore/services/queuedjob/MCRJobQueue.java
Possommi: but that is not going to work, the suggested changes are leading to a runtime failure while the current state is working like a charme

sebhofmann: The current state is vulnerable to sql injection. 
Also the code ist 99% duplicate. 
The method getJob is private so you can change it to
private <T> T getJob(Class<? extends MCRJobAction> action, Map<String, String> params, Function<TypedQuery<MCRJob>, T> queryTransfomer)  and reuse it.

Please also use the query api to prevent SQL Injection.
For examples see:
org.mycore.pi.MCRPIManager

PR #859, opened 2019-09-18
MCR-2099 - fix name of derivate search field maindoc

kkrebs:
it cause problems with usage of mycoreobject-dynamicfields.xsl and new
derobject xml

[Link to jira](https://mycore.atlassian.net/browse/MCR-2099).


rsteph-de (approved):
tested and works as designed

PR #873, opened 2019-10-23
MCR-2110 fixed removeEmptyLeaves in classbrowser when parameter was p…

yagee-de:
…resent but empty

[Link to jira](https://mycore.atlassian.net/browse/MCR-2110).


yagee-de:
mir-it tests run locally without failures

PR #902, opened 2020-01-10
MCR-2080 make user attributes repeatable

yagee-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2080).

### DIFF
--- a/mycore-user2/src/main/java/org/mycore/user2/MCRUser.java
+++ b/mycore-user2/src/main/java/org/mycore/user2/MCRUser.java
@@ -23,10 +23,12 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Date;
-import java.util.HashMap;
 import java.util.HashSet;
-import java.util.Map;
-import java.util.Map.Entry;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
+import java.util.SortedSet;
+import java.util.TreeSet;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -45,7 +47,6 @@ import javax.persistence.Id;
 import javax.persistence.Index;
 import javax.persistence.JoinColumn;
 import javax.persistence.ManyToOne;
-import javax.persistence.MapKeyColumn;
 import javax.persistence.NamedQueries;
 import javax.persistence.NamedQuery;
 import javax.persistence.Table;
@@ -61,9 +62,9 @@ import javax.xml.bind.annotation.XmlType;
 
 import org.hibernate.annotations.Cache;
 import org.hibernate.annotations.CacheConcurrencyStrategy;
+import org.hibernate.annotations.SortNatural;
 import org.mycore.common.MCRException;
 import org.mycore.common.MCRUserInformation;
-import org.mycore.user2.annotation.MCRUserAttribute;
 import org.mycore.user2.annotation.MCRUserAttributeJavaConverter;
 import org.mycore.user2.utils.MCRRolesConverter;
 import org.mycore.user2.utils.MCRUserNameConverter;
@@ -82,12 +83,12 @@ import org.mycore.user2.utils.MCRUserNameConverter;
 @Access(AccessType.PROPERTY)
 @Table(name = "MCRUser", uniqueConstraints = @UniqueConstraint(columnNames = { "userName", "realmID" }))
 @NamedQueries(@NamedQuery(name = "MCRUser.byPropertyValue",
-    query = "SELECT u FROM MCRUser u JOIN FETCH u.attributes ua WHERE KEY(ua) = :name  AND :value IN (VALUE(ua))"))
+    query = "SELECT u FROM MCRUser u JOIN FETCH u.attributes ua WHERE ua.name = :name  AND ua.value = :value"))
 // TODO use @Cacheable instead
 @Cache(usage = CacheConcurrencyStrategy.NONSTRICT_READ_WRITE)
 @XmlRootElement(name = "user")
 @XmlAccessorType(XmlAccessType.NONE)
-@XmlType(propOrder = { "ownerId", "realName", "eMail", "lastLogin", "validUntil", "roles", "attributesMap",
+@XmlType(propOrder = { "ownerId", "realName", "eMail", "lastLogin", "validUntil", "roles", "attributes",
     "password" })
 public class MCRUser implements MCRUserInformation, Cloneable, Serializable {
     private static final long serialVersionUID = 3378645055646901800L;
@@ -103,7 +104,7 @@ public class MCRUser implements MCRUserInformation, Cloneable, Serializable {
     private boolean disabled;
 
     /** The login user name */
-    @MCRUserAttribute
+    @org.mycore.user2.annotation.MCRUserAttribute
     @MCRUserAttributeJavaConverter(MCRUserNameConverter.class)
     @XmlAttribute(name = "name")
     private String userName;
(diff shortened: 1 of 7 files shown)


mycore-user2/src/main/java/org/mycore/user2/MCRUser.java
```diff
@@ -119,12 +120,12 @@
...
 
     /** The name of the person that this login user represents */
-    @MCRUserAttribute
+    @org.mycore.user2.annotation.MCRUserAttribute
     @XmlElement
     private String realName;
 
     /** The E-Mail address of the person that this login user represents */
-    @MCRUserAttribute
+    @org.mycore.user2.annotation.MCRUserAttribute
```
adlerre: Codacy Issue found: [Unnecessary use of fully qualified name 'org.mycore.user2.annotation.MCRUserAttribute' due to existing same package import 'org.mycore.user2.*'](https://app.codacy.com/manual/MyCoRe/mycore/pullRequest?prid=4797507)

adlerre:
Codacy Here is an overview of what got changed by this pull request:


```diff

Issues
======
+ Solved 1
- Added 2
           

Complexity increasing per file
==============================
- mycore-user2/src/main/java/org/mycore/user2/MCRUserAttribute.java  3
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/MyCoRe/mycore/pullRequest?prid=4797507&bid=15968144)

mycore-user2/src/main/java/org/mycore/user2/MCRUser.java
```diff
@@ -664,7 +668,7 @@ public void setValidUntil(Date validUntil) {
         return Arrays.stream(getRoles()).map(MCRRole::getName).collect(Collectors.toSet());
     }
 
-    @MCRUserAttribute(name = "roles", separator = ";")
+    @org.mycore.user2.annotation.MCRUserAttribute(name = "roles", separator = ";")
```
adlerre: Codacy Issue found: [Unnecessary use of fully qualified name 'org.mycore.user2.annotation.MCRUserAttribute' due to existing same package import 'org.mycore.user2.*'](https://app.codacy.com/manual/MyCoRe/mycore/pullRequest?prid=4797507)

mycore-user2/src/main/java/org/mycore/user2/MCRUser.java
```diff
@@ -710,27 +714,16 @@ private void setRoles(MCRRole[] roles) {
...
+    public void setUserAttribute(String name, String value) {
+        Optional<MCRUserAttribute> anyMatch = getAttributes().stream()
+            .filter(a -> a.getName().equals(Objects.requireNonNull(name)))
+            .findAny();
+        if (anyMatch.isPresent()) {
+            MCRUserAttribute attr = anyMatch.get();
+            attr.setValue(value);
+            getAttributes().removeIf(a -> a.getName().equals(name) && a != attr);
+        } else {
+            getAttributes().add(new MCRUserAttribute(name, value));
```
rsteph-de, line 726: If the order of attributes is not relevant (as it was in the previous map), it could just be:
 getAttributes().removeIf(a -> a.getName().equals(name));
 getAttributes().add(new MCRUserAttribute(name, value));

fluetze (commented):
is it right that
 public String getUserAttribute(String attribute)
will throw an exception if the attribute is not unique?
We want some attributes to be repeatable, esp. those für identifiers.

Also, MCROrcidUser.getUserIdentiferKeys should be changed to handle all identifiers of the user, not just one per identifier type, and the method must not expect that there is just only one per type

mycore-user2/src/main/java/org/mycore/user2/MCRUser.java
yagee-de: saves a database instruction in case an attribute is present (UPDATE vs. DELETE & INSERT) at the cost of more local instructions

yagee-de (commented):
> is it right that
> public String getUserAttribute(String attribute)
> will throw an exception if the attribute is not unique?
> We want some attributes to be repeatable, esp. those für identifiers.

the getUserAttribute(String) is there if the caller expect just one value - and for compatiblilty. If there are multiple values an exception is thrown. If the caller expect multiple values it should call getAttributes() instead:

```
user.getAttributes().stream()
    .filter(a -> a.getName().equals(attribute))
    .collect(Collectors.toSet());
``` 

> Also, MCROrcidUser.getUserIdentiferKeys should be changed to handle all identifiers of the user, not just one per identifier type, and the method must not expect that there is just only one per type

It should already do that, as far as I know.

fluetze:
> > is it right that
> > public String getUserAttribute(String attribute)
> > will throw an exception if the attribute is not unique?
> > We want some attributes to be repeatable, esp. those für identifiers.
> 
> the getUserAttribute(String) is there if the caller expect just one value - and for compatiblilty. If there are multiple values an exception is thrown. If the caller expect multiple values it should call getAttributes() instead:

For compatibility, I would expect the first value, or at least any value.
Say a user record currently has one value, current code works.
When a second value is added to the record, current code won't work any more.

> > Also, MCROrcidUser.getUserIdentiferKeys should be changed to handle all identifiers of the user, not just one per identifier type, and the method must not expect that there is just only one per type
> 
> It should already do that, as far as I know.

It calls user.getUserAttribute(...) and that will throw an Exception, see above

yagee-de:
> > > is it right that
> > > public String getUserAttribute(String attribute)
> > > will throw an exception if the attribute is not unique?
> > > We want some attributes to be repeatable, esp. those für identifiers.
> > 
> > 
> > the getUserAttribute(String) is there if the caller expect just one value - and for compatiblilty. If there are multiple values an exception is thrown. If the caller expect multiple values it should call getAttributes() instead:
> 
> For compatibility, I would expect the first value, or at least any value.
> Say a user record currently has one value, current code works.
> When a second value is added to the record, current code won't work any more.

As the returned value would be random, this should not happen. It acts the same as singleResult() in JPA and throws a RuntimeException if more than value is present. This is at least deterministic and normally should not affect old code beside the ORCID usage.

> 
> > > Also, MCROrcidUser.getUserIdentiferKeys should be changed to handle all identifiers of the user, not just one per identifier type, and the method must not expect that there is just only one per type
> > 
> > 
> > It should already do that, as far as I know.
> 
> It calls user.getUserAttribute(...) and that will throw an Exception, see above

NOW I see. Fixed it in 8fdcdcc!

adlerre:
Codacy Here is an overview of what got changed by this pull request:


```diff

Issues
======
+ Solved 1
- Added 2
           

Complexity increasing per file
==============================
- mycore-user2/src/main/java/org/mycore/user2/MCRUserAttribute.java  3
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/MyCoRe/mycore/pullRequest?prid=4797507&bid=15968144)

mycore-user2/src/main/java/org/mycore/user2/MCRUser.java
adlerre, line 128: Codacy Issue found: [Unnecessary use of fully qualified name 'org.mycore.user2.annotation.MCRUserAttribute' due to existing same package import 'org.mycore.user2.*'](https://app.codacy.com/manual/MyCoRe/mycore/pullRequest?prid=4797507)

mycore-user2/src/main/java/org/mycore/user2/MCRUser.java
adlerre, line 671: Codacy Issue found: [Unnecessary use of fully qualified name 'org.mycore.user2.annotation.MCRUserAttribute' due to existing same package import 'org.mycore.user2.*'](https://app.codacy.com/manual/MyCoRe/mycore/pullRequest?prid=4797507)

fluetze (approved):

kkrebs (approved):

PR #904, opened 2020-01-23
MCR-2143 migrate expert search mask to bootstrap 4

kkrebs:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2143).


sebhofmann (approved):
Works with MIR. We should consider adding col-md-6 to both boxes and mb-1 to the repeated input fields.

PR #946, opened 2020-03-31
MCR-2184 MCRPdfThumbnailGenerator can use the initial view of the pdf

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2184).


mycore-media/src/main/java/org/mycore/media/services/MCRPdfThumbnailGenerator.java
```diff
@@ -43,4 +49,30 @@ public boolean matchesFileType(String mimeType, MCRPath path) {
             return Optional.of(pdfRender);
         }
     }
+
+    private PDPage resolveFirstPage(PDDocument pdf) throws IOException {
```
yagee-de: better name: resolveInitialPage
first should be used for listable items and should return the head of the list. In this case the returned page can actually be the 7th

yagee-de (changes requested):
Good feature!

kkrebs (approved):

PR #976, opened 2020-05-26
MCR-2171 fix ThumbnailGenerator property

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2171)


rsteph-de (approved):
fixes the problem

PR #981, opened 2020-05-29
MCR-2219 add LIDO namespace

bt1cu:

### DIFF
--- a/mycore-base/src/main/java/org/mycore/common/MCRConstants.java
+++ b/mycore-base/src/main/java/org/mycore/common/MCRConstants.java
@@ -78,6 +78,11 @@ public final class MCRConstants {
 
     public static final Namespace DV_NAMESPACE = Namespace.getNamespace("dv", DV_URL);
 
+    /** The URL of the LIDO */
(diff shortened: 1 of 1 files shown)


mycore-base/src/main/java/org/mycore/common/MCRConstants.java
```diff
@@ -78,6 +78,11 @@
 
     public static final Namespace DV_NAMESPACE = Namespace.getNamespace("dv", DV_URL);
 
+    /** The URL of the LIDO */
+    private static final String LIDO_URL = "http://www.lido-schema.org";
+
+    public static final Namespace LIDO_NAMESPACE = Namespace.getNamespace("lido", LIDO_URL);
+
```
rsteph-de, line 85: Why not shorter:
public static final Namespace LIDO_NAMESPACE = Namespace.getNamespace("lido", ""http://www.lido-schema.org");

Shouldn't we inline the other private static final String *_URL constants as well?

kkrebs (approved):

PR #986, opened 2020-06-03
MCR-2185 change btn-default to btn-primary

kkrebs:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2185).


TiloNN (approved):
After hours of intense code review, 
I think we should give those complex changes a try.

PR #989, opened 2020-06-05
MCR-2220 add an URIResolver for i18n translation

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2220).


bt1cu (approved):
I think it should be okay.

PR #994, opened 2020-06-09
MCR-2226 iiif use colon in image identifier

rsteph-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2226).

### DIFF
--- a/mycore-iiif/src/main/java/org/mycore/iiif/image/MCRIIIFImageUtil.java
+++ b/mycore-iiif/src/main/java/org/mycore/iiif/image/MCRIIIFImageUtil.java
@@ -20,10 +20,10 @@ package org.mycore.iiif.image;
 
 import static org.mycore.iiif.image.resources.MCRIIIFImageResource.IIIF_IMAGE_API_2_LEVEL2;
 
-import java.net.URI;
-import java.net.URISyntaxException;
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
 
-import org.mycore.common.MCRException;
+import org.mycore.common.config.MCRConfiguration2;
 import org.mycore.frontend.MCRFrontendUtil;
 import org.mycore.iiif.image.impl.MCRIIIFImageImpl;
 import org.mycore.iiif.image.model.MCRIIIFImageProfile;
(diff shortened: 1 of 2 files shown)


MyCoRe-Travis:
Something gone wrong on 24541acb2b4dca30a77d10722edb0efbcb048740! 
 Please Check https://github.com/MyCoRe-Travis/test_artifacts/tree/5677-pull_request-994

mycore-iiif/src/main/java/org/mycore/iiif/image/MCRIIIFImageUtil.java
```diff
@@ -39,14 +39,7 @@ public static void completeProfile(MCRIIIFImageImpl impl, MCRIIIFImageProfile pr
...
     public static String encodeImageIdentifier(String imageIdentifier) {
-
-        try {
-            final URI uri = new URI(null, null, imageIdentifier, null, null);
-            return uri.toASCIIString().replace("/", "%2F");
-        } catch (URISyntaxException e) {
-            //99.99% this won't happen
-            throw new MCRException("Could not encode image identifier: " + imageIdentifier, e);
-        }
+        return URLEncoder.encode(imageIdentifier, StandardCharsets.UTF_8);
```
sebhofmann, line 42: This will not work because / needs to be encoded. Its also better to use URI encoding.

sebhofmann (changes requested):
Identifier with : is less readable when encoded.

mycore-iiif/src/main/java/org/mycore/iiif/image/MCRIIIFImageUtil.java
rsteph-de: It does work! - "/" is encoded by URLEncoder.encode().
https://docs.oracle.com/javase/7/docs/api/java/net/URLEncoder.html

URI.toASCIIString() did not work for me, because it interprets the image identifier as URI.
The part before a colon (":") in the identifier gets interpreted as scheme part (like http:, ...) and does not necessarily match the rules (e.g.: less than 8 characters, ...).

Since we are encoding the identifier to be used in a URL, I think URLEncoder.encode() is a good choice.

kkrebs:
PR #996 is a merge request against 2019.06 and contains new junit test. We should decide which implementation we would use and for which branch. The junit test should be included in any case.

mycore-iiif/src/main/java/org/mycore/iiif/image/MCRIIIFImageUtil.java
sebhofmann: Spaces are encoded wrong

rsteph-de:
won't merge - after discussion relevant changes will be handled in separate issues

PR #1001, opened 2020-06-14
MCR-2235 add searchfield derivateType

rsteph-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2235).

This implementation assumes, that there is only ONE derivate_type for each derivate.
If we agree on this - please merge.
If multiple derivate_types shall be allowed, the code needs to be adjusted:
The field needs to be set to multivalued and xpath modified.

### DIFF
--- a/mycore-solr/src/main/resources/components/solr/config/solr/main/solr-schema.json
+++ b/mycore-solr/src/main/resources/components/solr/config/solr/main/solr-schema.json
@@ -368,6 +368,12 @@
   },
   {
     "add-field": {
+      "name": "derivateType",
+      "type": "string"
+    }
+  },
+  {
+    "add-field": {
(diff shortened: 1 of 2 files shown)


mycore-solr/src/main/resources/components/solr/config/solr/main/solr-schema.json
```diff
@@ -366,6 +366,12 @@
       "type": "pint"
     }
   },
+  {
+    "add-field": {
+      "name": "derivateType",
+      "type": "string"
+    }
```
yagee-de: should be multivalued

yagee-de (changes requested):

yagee-de (approved):

PR #1016, opened 2020-06-22
MCR-2243 add xsl:function to convert mods to mycoreclass

yagee-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2243).

### DIFF
--- a/mycore-mods/pom.xml
+++ b/mycore-mods/pom.xml
@@ -58,6 +58,11 @@
       <artifactId>junit</artifactId>
     </dependency>
     <dependency>
+      <groupId>net.sf.saxon</groupId>
+      <artifactId>Saxon-HE</artifactId>
+      <optional>true</optional>
+    </dependency>
+    <dependency>
       <groupId>org.apache.logging.log4j</groupId>
       <artifactId>log4j-api</artifactId>
     </dependency>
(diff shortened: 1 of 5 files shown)


MyCoRe-Travis:
Something gone wrong on a48001b970335ce6d621e6a09ff0e08a06b4b3ac! 
 Please Check https://github.com/MyCoRe-Travis/test_artifacts/tree/5747-pull_request-1016

MyCoRe-Travis:
Something gone wrong on cbb9bdcf319280dd707d715015bd6bf57449fdea! 
 Please Check https://github.com/MyCoRe-Travis/test_artifacts/tree/5749-pull_request-1016

fluetze (commented):
Wouldn't it be better to choose another separator, not slash?
mycoremods:/{sourcePath}/{subPath}
for example, with authorityURI and valueURI as parameters, both separated by a "/" ?

yagee-de:
> Wouldn't it be better to choose another separator, not slash?
> mycoremods:/{sourcePath}/{subPath}
> for example, with authorityURI and valueURI as parameters, both separated by a "/" ?

I think that '/' as a seperator shows the hierarchical nature better: ´, e.g.`modsclass:/authority/marcrelator/aut`
Another issue with 'query parameters' would be the usage of an URLEncoder that is, as far as I know, not available in XSL (yet) to encode the values. In the current state I could simply use the build-in `fn:encode-for-uri()`.

sebhofmann (approved):

PR #1042, opened 2020-07-17
MCR-2260 fix problem with cleaning SOLR configuration

bt1cu:
This change add a method to clean the SOLR configuration defined by
proerties before cleaning the schema.
Renaming the internal methods from clear to clean.

[Link to jira](https://mycore.atlassian.net/browse/MCR-2260).

### DIFF
--- a/mycore-solr/src/main/java/org/mycore/solr/commands/MCRSolrCommands.java
+++ b/mycore-solr/src/main/java/org/mycore/solr/commands/MCRSolrCommands.java
@@ -131,8 +131,8 @@ public class MCRSolrCommands extends MCRAbstractCommands {
             + "by using the Solr schema api for core with the id {0}",
         order = 70)
     public static void reloadSolrConfiguration(String configType, String coreID) {
-        MCRSolrConfigReloader.cleanConfig(configType, coreID);
-        MCRSolrSchemaReloader.cleanSchema(configType, coreID);
+        MCRSolrConfigReloader.reset(configType, coreID);
+        MCRSolrSchemaReloader.reset(configType, coreID);
         MCRSolrSchemaReloader.processSchemaFiles(configType, coreID);
         MCRSolrConfigReloader.processConfigFiles(configType, coreID);
     }
--- a/mycore-solr/src/main/java/org/mycore/solr/commands/MCRSolrCommands.java
+++ b/mycore-solr/src/main/java/org/mycore/solr/commands/MCRSolrCommands.java
@@ -131,7 +131,8 @@ public class MCRSolrCommands extends MCRAbstractCommands {
             + "by using the Solr schema api for core with the id {0}",
         order = 70)
     public static void reloadSolrConfiguration(String configType, String coreID) {
-        MCRSolrSchemaReloader.clearSchema(configType, coreID);
+        MCRSolrConfigReloader.cleanConfig(configType, coreID);
+        MCRSolrSchemaReloader.cleanSchema(configType, coreID);
         MCRSolrSchemaReloader.processSchemaFiles(configType, coreID);
         MCRSolrConfigReloader.processConfigFiles(configType, coreID);
     }
--- a/mycore-solr/src/main/java/org/mycore/solr/schema/MCRSolrConfigReloader.java
+++ b/mycore-solr/src/main/java/org/mycore/solr/schema/MCRSolrConfigReloader.java
@@ -111,15 +111,18 @@ public class MCRSolrConfigReloader {
                 String ignoreListConfiguration = MCRConfiguration2
                     .getString("MCR.Solr.ObserverConfigTypes." + observedType + ".toClean").orElse("");
                 List<String> cleanList = Arrays.asList(ignoreListConfiguration.split("\\s*,\\s*"));
-                if (cleanList.size() == 0) {
+                if (cleanList.isEmpty()) {
                     continue;
                 }
                 JsonObject observedConfig = configPart.getAsJsonObject(observedType);
+                if (observedConfig == null) {
+                    continue;
+                }
                 Set<Map.Entry<String, JsonElement>> entries = observedConfig.entrySet();
                 for (Map.Entry<String, JsonElement> entry : entries) {
                     if (cleanList.contains(entry.getKey())) {
                         final JsonObject cleanJsonObject = new JsonObject();
-                        final String commandName = "delete-" + observedType.toLowerCase(Locale.ROOT);
+                        final String commandName = "delete-" + observedType.toLowerCase();
                         if (isKnownSolrConfigCommmand(commandName)) {
                             cleanJsonObject.addProperty(commandName, entry.getKey());
                             executeSolrCommand(coreURL, cleanJsonObject.toString());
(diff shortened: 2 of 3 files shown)


MyCoRe-Travis:
Something gone wrong on 9be07e22c2968a96d34b10b5b792dbe015c6f4cb! 
 Please Check https://github.com/MyCoRe-Travis/test_artifacts/tree/5874-pull_request-1042

rsteph-de (approved):
looks good for me, works as discussed

mycore-solr/src/main/java/org/mycore/solr/commands/MCRSolrCommands.java
```diff
@@ -131,7 +131,8 @@ public static void switchSolrCore(String coreID1, String coreID2) {
             + "by using the Solr schema api for core with the id {0}",
         order = 70)
     public static void reloadSolrConfiguration(String configType, String coreID) {
-        MCRSolrSchemaReloader.clearSchema(configType, coreID);
+        MCRSolrConfigReloader.cleanConfig(configType, coreID);
+        MCRSolrSchemaReloader.cleanSchema(configType, coreID);
```
yagee-de: Shouldn't this stay clear?
Like List.clear() or clearing a cache? You don't wipe it clean, but remove something. You are clearing it up.

yagee-de (changes requested):

bt1cu:
Clear means in my opinion that the object will be remove but clean means that some things in the object will be remove. So the methods clean some defined elements but NOT ALL entries. If you clean your flat it is not empty after cleaning. The overlay file is not removed and the managed schema is not empty.

I can't understand, that this important bug fix is blocked by a developer just because of this naming discussion.

mycore-solr/src/main/java/org/mycore/solr/schema/MCRSolrConfigReloader.java
```diff
@@ -89,19 +90,64 @@
...
+            String coreURL = MCRSolrClientFactory.get(coreID)
+                .orElseThrow(() -> MCRSolrUtils.getCoreConfigMissingException(coreID)).getV1CoreURL();
+            List<String> observedTypes = getObserverConfigTypes();
+            JsonObject currentSolrConfig = retrieveCurrentSolrConfig(coreURL);
+            JsonObject configPart = currentSolrConfig.getAsJsonObject("config");
+
+            for (String observedType : observedTypes) {
+                String ignoreListConfiguration = MCRConfiguration2
+                    .getString("MCR.Solr.ObserverConfigTypes." + observedType + ".toClean").orElse("");
+                List<String> cleanList = Arrays.asList(ignoreListConfiguration.split("\\s*,\\s*"));
```
yagee-de: Property Naming: Use uppercase after '.'
Property list values should be separated by ','
```
                List<String> cleanList = MCRConfiguration2
                    .getString("MCR.Solr.ObserverConfigTypes." + observedType + ".ToClean")
                    .map(MCRConfiguration2::splitValue)
                    .orElseGet(Stream::empty)
                    .collect(Collectors.toList());

```

mycore-solr/src/main/java/org/mycore/solr/schema/MCRSolrConfigReloader.java
```diff
@@ -89,19 +90,64 @@
...
+     * MCR.Solr.ObserverConfigTypes defined SOLR configuration parts for all in the
+     * MCR.Solr.ObserverConfigTypes.{observedType}.toClean property defined entries. For each entry the 
+     * method will process a SOLR delete command via API.
+     *
+     * @param configType the name of the configuration directory containing the SOLR core configuration
+     * @param coreID the ID of the core, which the configuration should be applied to
+     */
+    public static void cleanConfig(String configType, String coreID) {
+        LOGGER.info("Clean config definitions for core " + coreID + " using configuration " + configType);
+        try (CloseableHttpClient httpClient = HttpClients.createDefault()) {
```
yagee-de: httpClient seems not to be used

mycore-solr/src/main/java/org/mycore/solr/schema/MCRSolrConfigReloader.java
```diff
@@ -89,19 +90,64 @@
...
+                    continue;
+                }
+                Set<Map.Entry<String, JsonElement>> entries = observedConfig.entrySet();
+                for (Map.Entry<String, JsonElement> entry : entries) {
+                    if (cleanList.contains(entry.getKey())) {
+                        final JsonObject cleanJsonObject = new JsonObject();
+                        final String commandName = "delete-" + observedType.toLowerCase(Locale.ROOT);
+                        if (isKnownSolrConfigCommmand(commandName)) {
+                            cleanJsonObject.addProperty(commandName, entry.getKey());
+                            executeSolrCommand(coreURL, cleanJsonObject.toString());
```
yagee-de: the only method to throw IOException, please adjust scope of try-catch around this or escalate around for statement

yagee-de (changes requested):

rsteph-de:
We discussed this issue in our last phone conference.
We came to the conclusion that names of public methods should only be changed if absolutely necessary.
After another discussion about the meaning of "clear" and "clean" in this context, we came up with the suggestion to rename 

```
        MCRSolrConfigReloader.cleanConfig(configType, coreID);
        MCRSolrSchemaReloader.cleanSchema(configType, coreID);
```
to
```
        MCRSolrConfigReloader.reset(configType, coreID);
        MCRSolrSchemaReloader.reset(configType, coreID);
```
The terms config/schema are already part of the class name and don't have to be repeated in method names.

For us **clear** means, to remove ALL elements, which does not reflect, that we keep some "default" elements,
so **clean** might better fit to the situation. To avoid the confusion, we came up with
**reset**, which means that we set back the config/schema into the inital state with some predefinied fields.

mycore-solr/src/main/java/org/mycore/solr/schema/MCRSolrConfigReloader.java
rsteph-de: I would like to see this as a utitlity method in MCRConfiguration2. getStringList(String propName).
Even if streams are elegant and this code is part of the Javadoc, 
it would simplify the code in modules / application where a list of strings is needed.

bt1cu:
Wenn derartige Kleinigkeiten dazu führen, dass problematische Bug-Fixes über lange Zeit nicht in den Code kommen: macht wass ihr wollt, aber macht!!!! Offenbar ist das wichtiger, als das Problem zu beheben ...

MyCoRe-Travis:
Something gone wrong on 92a50b40d32465f7d9caab2796c52aba890c400d! 
 Please Check https://github.com/MyCoRe-Travis/test_artifacts/tree/5948-pull_request-1042

MyCoRe-Travis:
Something gone wrong on 92a50b40d32465f7d9caab2796c52aba890c400d! 
 Please Check https://github.com/MyCoRe-Travis/test_artifacts/tree/5948-pull_request-1042

MyCoRe-Travis:
Something gone wrong on 92a50b40d32465f7d9caab2796c52aba890c400d! 
 Please Check https://github.com/MyCoRe-Travis/test_artifacts/tree/5948-pull_request-1042

bt1cu:
It seems that the last commit work in my environment.

PR #1044, opened 2020-07-23
MCR-2264 add 'set classification of derivate {0} to {1}' command

yagee-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2264).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/frontend/cli/MCRDerivateCommands.java
+++ b/mycore-base/src/main/java/org/mycore/frontend/cli/MCRDerivateCommands.java
@@ -954,13 +954,14 @@ public class MCRDerivateCommands extends MCRAbstractCommands {
 
     @MCRCommand(syntax = "set classification of derivate {0} to {1}",
         help = "Sets the classification of derivate {0} to the categories {1} (comma separated) "
-            + "of classification 'derivate_types', removing any previous definition.")
+            + "of classification 'derivate_types' or any fully qualified category, removing any previous definition.")
     public static void setClassificationOfDerivate(String derivateIDStr, String categoriesCommaList)
         throws MCRAccessException {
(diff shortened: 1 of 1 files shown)


sebhofmann:
maybe it would be better if you could pass a comma seperated list of classifications.

yagee-de:
> maybe it would be better if you could pass a comma seperated list of classifications.

Do we have a use-case with a derivate having more than one classification category?

rsteph-de:
> Do we have a use-case with a derivate having more than one classification category?

Currently I haven't, but since we are using "standard MyCoRe classification behaviour" internally, we should be prepared for that.

yagee-de:
> > Do we have a use-case with a derivate having more than one classification category?
> 
> Currently I haven't, but since we are using "standard MyCoRe classification behaviour" internally, we should be prepared for that.

OK, solr field definition is also multivalued already. Implemented requested changes.

MyCoRe-Travis:
Something gone wrong on 518354a9b798617bd860a9c4f626ba841a7555cf! 
 Please Check https://github.com/MyCoRe-Travis/test_artifacts/tree/5892-pull_request-1044

sebhofmann:
Its only possible to set derivate_types not all other classifications.

rsteph-de:
and advanced command call could look like this:
`set classification of derivate mir_derivate_0123 to derivate_types:fulltext,derivate_types:cover,hello:world`

yagee-de:
> and advanced command call could look like this:
> `set classification of derivate mir_derivate_0123 to derivate_types:fulltext,derivate_types:cover,hello:world`

or simplier
* single default classification: `set classification of derivate mir_derivate_12345678 to content`
* multiple combined: `set classification of derivate mir_derivate_12345678 to content,hello:world`

sebhofmann (approved):

PR #1045, opened 2020-07-29
MCR-2265 add EventHandler for derivate classification

yagee-de:
not enabled by default

[Link to jira](https://mycore.atlassian.net/browse/MCR-2265).


sebhofmann (changes requested):
It would be better if the Property allows multiple Classifications. It should not only allow derivate_types, otherwise the name should be changed from DefaultCategory to DefaultDerivateType.

sebhofmann (approved):

PR #1057, opened 2020-09-02
MCR-2273 Add the URI resolver to check permission

bt1cu:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2273).


yagee-de (approved):
For me, I am happy with the solution for XSL3 we have a good replacement for the Xalan Extension we used before.

PR #1119, opened 2020-11-20
MCR-2316 Value substitution in i18n URIResolver

OPNA2608:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2316).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/common/xml/MCRURIResolver.java
+++ b/mycore-base/src/main/java/org/mycore/common/xml/MCRURIResolver.java
@@ -1603,7 +1603,7 @@ public final class MCRURIResolver implements URIResolver {
          * Resolves the I18N String value for the given property.<br><br>
          * <br>
          * Syntax: <code>i18n:{i18n-code},{i18n-prefix}*,{i18n-prefix}*...</code> or <br>
-         *         <code>i18n:{i18n-code}</code>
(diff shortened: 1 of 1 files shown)


yagee-de:
How do you handle values that contain special characters: ",", "[", "]"?

bt1cu:
The syntax should be i18n:{property}:{param1} ... more Parameter with :{paramn}

bt1cu (approved):

mycore-base/src/main/java/org/mycore/common/xml/MCRURIResolver.java
```diff
@@ -1631,7 +1638,15 @@ public Source resolve(String href, String base) {
 
             final Element i18nElement = new Element("i18n");
             if (!target.contains("*") && !target.contains(",")) {
-                i18nElement.addContent(MCRTranslation.translate(target));
+                String translation;
+                if (target.contains(":")) {
+                    translation = new MessageFormat(
```
sebhofmann: Formating should be handled by MCRTranslation:
```
final int i = target.indexOf(":");
MCRTranslation.translate(target.substring(0, i), (Object[]) target.substring(i+1).split(":"));
```
(not tested)

OPNA2608: Appears to work, thanks.

sebhofmann (approved):

PR #1136, opened 2021-01-21
MCR-2332 commands to load update classification from uri

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2332).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/frontend/cli/MCRClassification2Commands.java
+++ b/mycore-base/src/main/java/org/mycore/frontend/cli/MCRClassification2Commands.java
@@ -34,6 +34,7 @@ import java.util.stream.Collectors;
 
 import javax.xml.transform.Transformer;
 import javax.xml.transform.TransformerConfigurationException;
+import javax.xml.transform.TransformerException;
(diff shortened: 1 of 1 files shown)


MyCoRe-Travis:
Something gone wrong on d31d0ce0be5b3746e1474f74970b174952bd063d! 
 Please Check https://github.com/MyCoRe-Travis/test_artifacts/tree/6350-pull_request-1136

fluetze (changes requested):
Why not use MCRSourceContent.getInstance( String uri ).asXML();
Why not completely replace  existing "from url" method?

sebhofmann:
> Why not use MCRSourceContent.getInstance( String uri ).asXML();

I will change that.

> Why not completely replace existing "from url" method?

Because i want this change in LTS and its would be a to big semantically change in my opinion.

rsteph-de:
Nice Feature, it allows us finally to load classifications from classpath.

I would suggest to delete the "from url" commands only in master (to keep list of commands clean).

fluetze (approved):

PR #1148, opened 2021-02-09
MCR-2329 missing systemid entry for xml-2009.xsd for mycore-pi is missing

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2329).


fluetze (approved):
Tested in Duisburg-Essen, works

PR #1155, opened 2021-02-10
MCR-2343 UserID as reference for guest

golsch:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2343).


yagee-de:
MCR-2343 is not resolved and is marked as fixed for 2020.06.1 but this PR for merge into master.

PR #1171, opened 2021-02-17
MCR-2337 Allow custom user attributes to be stored in JWT token

golsch:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2337).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/frontend/jersey/MCRJWTUtil.java
+++ b/mycore-base/src/main/java/org/mycore/frontend/jersey/MCRJWTUtil.java
@@ -56,11 +56,7 @@ public class MCRJWTUtil implements MCRStartupHandler.AutoExecutable {
 
     private static Algorithm SHARED_SECRET;
 
-    public static JWTCreator.Builder getJWTBuilder(MCRUserInformation userInformation) {
-        return getJWTBuilder(userInformation, null);
-    }
-
-    public static JWTCreator.Builder getJWTBuilder(MCRUserInformation userInformation, String[] userAttributes) {
+    public static JWTCreator.Builder getJWTBuilder(MCRUserInformation userInformation, String... userAttributes) {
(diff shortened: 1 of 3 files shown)


mycore-base/src/main/java/org/mycore/frontend/jersey/MCRJWTUtil.java
```diff
@@ -55,18 +57,31 @@
     private static Algorithm SHARED_SECRET;
 
     public static JWTCreator.Builder getJWTBuilder(MCRUserInformation userInformation) {
+        return getJWTBuilder(userInformation, null);
+    }
+
+    public static JWTCreator.Builder getJWTBuilder(MCRUserInformation userInformation, String[] userAttributes) {
```
yagee-de: better use vararg `String... userAttributes` here, without any userattributes one can simply call getJWTBuilder(userInformation) and omit null as a second parameter

mycore-restapi/src/main/java/org/mycore/restapi/MCRSessionFilter.java
```diff
@@ -320,6 +320,9 @@ public String getUserAttribute(String attribute) {
             if (MCRUserInformation.ATT_EMAIL.equals(attribute)) {
                 return jwt.getClaim("email").asString();
             }
+            if (!jwt.getClaim(MCRJWTUtil.JWT_USER_ATTRIBUTE_PREFIX + attribute).isNull()) {
```
yagee-de: check is not needed as getClaim().asString() already returns null

yagee-de (changes requested):

mycore-base/src/main/java/org/mycore/frontend/jersey/MCRJWTUtil.java
golsch: nice, I didn't know that, then I can save the adapter.

yagee-de (approved):

PR #1174, opened 2021-02-18
MCR-2339 login ArrayIndexOutOfBoundException if you enter a Username with @ as last char

OPNA2608:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2339).

### DIFF
--- a/mycore-user2/src/main/java/org/mycore/user2/MCRUserManager.java
+++ b/mycore-user2/src/main/java/org/mycore/user2/MCRUserManager.java
@@ -89,7 +89,7 @@ public class MCRUserManager {
             return getUser(userName, MCRRealmFactory.getLocalRealm());
         } else {
             String[] parts = userName.split("@");
-            return getUser(parts[0], parts[1]);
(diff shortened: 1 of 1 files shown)


yagee-de (changes requested):
I do not think we should allow an empty realmId here
if `parts.length != 2` throw IllegalArgumentException here

sebhofmann:
> I do not think we should allow an empty realmId here
> if `parts.length != 2` throw IllegalArgumentException here

Also second "@" is possible.

OPNA2608:
> if `parts.length != 2` throw IllegalArgumentException here

I don't think replacing one MyCoRe kernel exception screen with a slightly different one is the best UX solution here.

Bildschirmfoto von 2021-02-18 12-17-25

---

What if I do this instead:
```java
if (parts.length == 2) {
    return getUser(parts[0], parts[1]);
} else {
    return null;
}
```

Since `checkPassword` already tests for `getUser(userName) == null, that should work without pushing an empty realmId through the realm checking code.

yagee-de (approved):

PR #1178, opened 2021-02-18
MCR-2140 Make XPaths unique in editors

sebhofmann:
* Fixed buildXPath() to always use pos predicates for elements
* Updated tests
* Rename tests to be more specific
* Checkboxes may be multiple, xpath [1] still means "all"
* Fixed select multiple handling and added test
* Fixed resubmission check for select multiple values
* Fixed unused imports

[Link to jira](https://mycore.atlassian.net/browse/MCR-2140).


kkrebs (approved):
successfully testet in UBO

fluetze (approved):

PR #1179, opened 2021-02-19
MCR-1867 On small screens one cannot access all commands

eckstueck:
fixed dropdown

[Link to jira](https://mycore.atlassian.net/browse/MCR-1867).


Possommi (approved):
works like a charm

PR #1216, opened 2021-03-28
MCR-2364 remove JPA dependency from MCRSession

yagee-de:
added MCRPersistenceTransaction as an interface for this indirection
while keeping support for "onCommit" jobs

[Link to jira](https://mycore.atlassian.net/browse/MCR-2364).


MyCoRe-Travis:
Something gone wrong on eb7d63f387030740ea2329e46167027b411bcdde! 
 Please Check https://github.com/MyCoRe-Travis/test_artifacts/tree/6872-pull_request-1216

sebhofmann:
MCRSession does enhance the JPA API, so we can listen for Transaction Events (commit). The Javadoc of MCRSession says:
`Instances of this class collect information kept during a session like the currently active user, the preferred language etc.`

The class does things its not suppose to do, according to the Javadoc.

We should either:

1. Keep the dependency, because MCRSession handles with JPA, so it should have JPA as a Dependency.
2. Remove the Code from MCRSession and put it somewhere else.

This PR does:
3. Remove the dependency and create a "abstraction" of EntityTransaction, but there is still a runtime dependency to JPA or another Implementation. It does this by creating a almost identical Copy of EntityTransaction (which is a JPA Class), called MCRSessionTransaction. The class still does things which it is not supposed to do and still have a (runtime)dependency to JPA. 

The best Way in my opinion would be 2. Remove the Code ....

1. org.mycore.backend.jpa.MCREntityManagerProvider.getCurrentEntityManager() should return a MCRWrappedEntityManager
2. MCRWrappedEntityManager.getTransaction() should return a MCRWrappedTransaction
3. MCRWrappedTransaction knows when a Transaction is created/destroyed. It could store that Transaction to the MCRSession 
4. Create MCRWrappedTransactionManager which gets informed by MCRWrappedTransaction if a Transaction is created/destroyed and could do the Listener Work which is now in MCRSession.

Pro
- No extra methods to start Transactions, just JPA API
- MCRSession would do what its Javadoc says
- There would be no runtime dependency from MCRSession to MCRJPA

Contra
- Its maybe more Code
- requires to change the API

yagee-de:
Thanks for looking at my code. I thought of it long time on how to provide the needed change without big changes through the code. This is quite a central class as you know. You brought up some valid point but I try to discuss with you why I think this PR is sufficient and a better option - for now.

> MCRSession does enhance the JPA API, so we can listen for Transaction Events (commit). The Javadoc of MCRSession says:
> `Instances of this class collect information kept during a session like the currently active user, the preferred language etc.`
> 
> The class does things its not suppose to do, according to the Javadoc.
> 
> We should either:
> 
>     1. Keep the dependency, because MCRSession handles with JPA, so it should have JPA as a Dependency.

No, the ticket is to remove the dependency to JPA from MCRSession. MCRSession should not depend on larger 3rd party dependencies as it is a central class. As such we need to provide a reverse dependency through this indirection provided by this PR.

> 
>     2. Remove the Code from MCRSession and put it somewhere else.

The actual transaction handling is now part of the JPA package. On the other side there is code left in MCRSession that is handling the "onCommit" case that is no directly linked to JPA but to the abstract concept that data stored during a transaction is available to other linked contexts, e.g. other threads spawned by event handler. Due to I-solation we have to wait for a transaction to complete.

> This PR does:
> 3. Remove the dependency and create a "abstraction" of EntityTransaction, but there is still a runtime dependency to JPA or another Implementation. It does this by creating a almost identical Copy of EntityTransaction (which is a JPA Class), called MCRSessionTransaction. The class still does things which it is not supposed to do and still have a (runtime)dependency to JPA.

MCRSessionTransaction is an interface and has no compile or runtime dependencies to JPA. Its only implementation MCREntityTransaction which is part of the jpa package has the dependency to JPA. MCRSession will run fine without MCREntityTransaction. 

> The best Way in my opinion would be 2. Remove the Code ....
> 
>     1. org.mycore.backend.jpa.MCREntityManagerProvider.getCurrentEntityManager() should return a MCRWrappedEntityManager
> 
>     2. MCRWrappedEntityManager.getTransaction() should return a MCRWrappedTransaction
> 
>     3. MCRWrappedTransaction knows when a Transaction is created/destroyed. It could store that Transaction to the MCRSession
> 
>     4. Create MCRWrappedTransactionManager which gets informed by MCRWrappedTransaction if a Transaction is created/destroyed and could do the Listener Work which is now in MCRSession.
> 
> 
> Pro
> 
>     * No extra methods to start Transactions, just JPA API
> 
>     * MCRSession would do what its Javadoc says
> 
>     * There would be no runtime dependency from MCRSession to MCRJPA
> 
> 
> Contra
> 
>     * Its maybe more Code
> 
>     * requires to change the API

Contra
Base classes which now handle transactions through MCRSession have to have direct dependency to JPA then. We are not talking about classes that do EntityManager calls anyway but classes that provide the context. And those classes run normally fine without JPA.

And just to state this clear again. This PR does not remove only a compile dependency to JPA in exchange of introducing a runtime dependency. There is NO dependency TO JPA. In fact it should run fine without JPA or it should be handled like a bug.

This PR by no means targets to be the optimal shot but it gets the requirement done and as far as I know will provide the smallest change necessary. This will hopefully not introduce any major bugs.

sebhofmann:
> Base classes which now handle transactions through MCRSession have to have direct dependency to JPA then. We are not talking about classes that do EntityManager calls anyway but classes that provide the context. And those classes run normally fine without JPA.

If a base class starts or ends a database transaction or wants to know when a database transaction is commited, then it should have a dependency to JPA.

mycore-base/src/main/java/org/mycore/common/MCRSession.java
```diff
@@ -153,14 +158,16 @@ public void close() {
...
+
     /**
      * The constructor of a MCRSession. As default the user ID is set to the value of the property variable named
      * 'MCR.Users.Guestuser.UserName'.
      */
     MCRSession() {
         userInformation = guestUserInformation;
         setCurrentLanguage(MCRConfiguration2.getString("MCR.Metadata.DefaultLang").orElse(MCRConstants.DEFAULT_LANG));
-        dataBaseAccess = MCRHIBConnection.isEnabled();
+        dataBaseAccess = MCRConfiguration2.getBoolean("MCR.Persistence.Database.Enable").orElse(true)
```
rsteph-de, line 169: There should be a utility method / class for this essential check and not just only a "low-level" property lookup.

The same check is repeated later in the code.

mycore-base/src/main/java/org/mycore/backend/jpa/MCREntityTransaction.java
```diff
@@ -0,0 +1,59 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with MyCoRe.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.mycore.backend.jpa;
+
+import org.mycore.common.MCRSessionTransaction;
+
+public class MCREntityTransaction implements MCRSessionTransaction {
```
rsteph-de: I am "out" - the naming is confusing for me:
MCREntityTransaction / MCRSessionTransaction 

Do we really implement our own/new Transaction or are we just managing an existing JPA-Transaction here?

mycore-base/src/main/java/org/mycore/datamodel/classifications2/impl/MCRCategLinkServiceImpl.java
```diff
@@ -63,9 +62,7 @@
  */
 public class MCRCategLinkServiceImpl implements MCRCategLinkService {
 
-    private static MCRHIBConnection HIB_CONNECTION_INSTANCE;
```
rsteph-de: Wouldn't it be better to do this MCRHIBConnection removal in another (separate, smaller) PULL-request?

rsteph-de (commented):
my first thoughts on this pull-request ...

mycore-base/src/main/java/org/mycore/datamodel/classifications2/impl/MCRCategLinkServiceImpl.java
yagee-de: We could make a separate "remove old hibernate code" ticket

mycore-base/src/main/java/org/mycore/backend/jpa/MCREntityTransaction.java
yagee-de: MCRSessionTransaction could also be named MCRDBTransaction or MCRDatabaseTransaction, while MCREntityTransaction fits well I think.

MyCoRe-Travis:
Something gone wrong on 3cafb5d174e05efa60fe39489e803dddc35b28e2! 
 Please Check https://github.com/MyCoRe-Travis/test_artifacts/tree/6991-pull_request-1216

mycore-base/src/main/java/org/mycore/common/MCRSession.java
yagee-de: It's just reading a property with a default value and you need an extra method for it?

rsteph-de (approved):
looks good for me (after discussion)

PR #1228, opened 2021-04-12
MCR-2359 (3) MCRConfigurableInstance bugfix generator init

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2359).


mycore-pi/src/main/java/org/mycore/pi/MCRGenericPIGenerator.java
```diff
@@ -143,7 +144,11 @@
 
     public MCRGenericPIGenerator() {
```
rsteph-de, line 145: Do you really need the "empty" default constructor?

rsteph-de (commented):
see comments

mycore-pi/src/main/java/org/mycore/pi/MCRGenericPIGenerator.java
sebhofmann: Yes because there is a non public constructor for testing.

rsteph-de (approved):
looks good for me

PR #1267, opened 2021-06-25
MCR-2429 Remove Transaction methods from MCRSession

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2429).


yagee-de:
I would prefer a step before this PR that removes the binding of EntityManager and MCRSession and removes the need for this ugly session handling that would be necessary after the PR.

sebhofmann:
> I would prefer a step before this PR that removes the binding of EntityManager and MCRSession and removes the need for this ugly session handling that would be necessary after the PR.

This is the next step, according to the Plan:
https://mycore.atlassian.net/browse/MCR-2364?focusedCommentId=48550

rsteph-de (approved):
looks good for me 
as the next step to cleanup session and transaction handling.

PR #1276, opened 2021-07-07
Upload test results on CI failure

OPNA2608:
To upload all the logs & screenshots as an artifact if a build fails.

Failed CI example with this change applied: https://github.com/OPNA2608/mycore/actions/runs/1004600649

grafik
grafik


rsteph-de (approved):
That's what we need.

Lets merge into branches and adjust later if necessary

kkrebs (approved):

PR #1289, opened 2021-07-26
MCR-2443 Add basic OCFL-based metadata store

OPNA2608:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2443).

I wasn't able to get this finished in time, so I'll have to push it as a draft. Cramming the original reworked store structure + the new OCFL store into the now-changed metadatamanager structure was too overwhelming for me to get done with the amount of time that was left. I'm sorry about that.

The idea here is to make a new Store - Metadata - MetadataVersion structure that's reusable and storage-agnostic (not relying on directly touchable or even locally existing structure) to make adding further stores in the future more seamless than this. For this, only new stores that extend MCRNewMetadataStore need to be written while MCRNewMetadata and MCRNewMetadataVersion only delegate operations to their linked store, unlike the currently in MyCoRe classes while need to be reimplemented for each new store and directly interact with the data.


rsteph-de (approved):
i.O.

PR #1294, opened 2021-08-05
MCR-2446 add support for XInclude

yagee-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2446).


kkrebs (changes requested):
I would close this PR, because this implementation is part of https://github.com/MyCoRe-Org/mycore/pull/1295.

PR #1295, opened 2021-08-05
MCR-2446 xinclude support

rsteph-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2446).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/common/xml/MCREntityResolver.java
+++ b/mycore-base/src/main/java/org/mycore/common/xml/MCREntityResolver.java
@@ -37,6 +37,8 @@ import javax.xml.catalog.CatalogException;
 import javax.xml.catalog.CatalogFeatures;
 import javax.xml.catalog.CatalogManager;
 import javax.xml.catalog.CatalogResolver;
+import javax.xml.transform.Source;
+import javax.xml.transform.TransformerException;
 
 import org.apache.commons.io.IOUtils;
 import org.apache.logging.log4j.LogManager;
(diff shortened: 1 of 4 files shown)


mycore-base/src/main/java/org/mycore/common/xml/MCREntityResolver.java
```diff
@@ -104,15 +108,29 @@ private static boolean isAbsoluteURL(String url) {
...
+            MCRURIResolver uriResolver = MCRURIResolver.instance();
+            String scheme = uriResolver.getScheme(systemId, null);
+            try {
+                if (scheme != null && uriResolver.getResolver(scheme) != null) {
+                    Source s = MCRURIResolver.instance().resolve(systemId, null, false);
+                    return new MCRSourceContent(s).getInputSource();
+                }
+            } catch (MCRUsageException | TransformerException e) {
+                return null;
+            }
```
yagee-de: Simply returning null without any notice? Could cause hard to find errors. We should at least **warn** in the logs.

yagee-de (changes requested):

mycore-base/src/main/java/org/mycore/common/xml/MCREntityResolver.java
rsteph-de: O.K. I will add a log message, but the old implementation did also return silent null

fluetze (commented):
Are there any security considerations, e.g. including a secret/hidden xml resource from the filesystem into a public webpage?

kkrebs:
we should look into that more closely, so I convert this pull request to draft

kkrebs:
This solution is to generic, we need a seperate ticket

PR #1306, opened 2021-08-16
MCR-2452 Add libraries neccesary to render JPEG 2000 images in PDF fi…

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2452).

### DIFF
--- a/mycore-iview2/pom.xml
+++ b/mycore-iview2/pom.xml
@@ -15,6 +15,16 @@
   </properties>
   <dependencies>
     <dependency>
+      <!-- needed for PDF documents with JPEG 2000 images-->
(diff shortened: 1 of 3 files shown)


yagee-de (changes requested):
Please use another library as this library may not be used with GPLv3 Code:

> The jj2000 package in this module is licensed under the JJ2000 license and is therefore not compatible with the GPL 3 license.

>Those intending to use this software module in hardware or software products are advised that their use may infringe existing patents. 

Link https://github.com/jai-imageio/jai-imageio-jpeg2000/blob/master/README.md

kkrebs:
until license questions are clarified this pr should only be a draft

PR #1312, opened 2021-08-18
MCR-2431 improve xml access system

kkrebs:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2431).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/access/MCRAccessManager.java
+++ b/mycore-base/src/main/java/org/mycore/access/MCRAccessManager.java
@@ -65,11 +65,20 @@ public class MCRAccessManager {
         MCRShutdownHandler.getInstance().addCloseable(EXECUTOR_SERVICE::shutdownNow);
     }
 
+    @SuppressWarnings("unchecked")
     public static <T extends MCRAccessInterface> T getAccessImpl() {
-        return (T) MCRConfiguration2.getSingleInstanceOf("MCR.Access.Class", MCRAccessBaseImpl.class).get();
+        return (T) MCRConfiguration2.<MCRAccessInterface>getInstanceOf("MCR.Access.Class")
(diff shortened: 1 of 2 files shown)


mycore-base/src/main/java/org/mycore/access/MCRAccessManager.java
```diff
@@ -65,11 +65,20 @@
...
-        return (T) MCRConfiguration2.getSingleInstanceOf("MCR.Access.Class", MCRAccessBaseImpl.class).get();
+        return (T) MCRConfiguration2.<MCRAccessInterface>getInstanceOf("MCR.Access.Class")
+            .orElseGet(MCRAccessBaseImpl::new);
     }
 
     private static MCRAccessCheckStrategy getAccessStrategy() {
+        // if acccessStrategy equals accessImp we reuse the accessImpl, 
+        // to make sure, that only one singleton gets created
+        // (used to instantiate fact-based access system) 
+        if (MCRConfiguration2.getString("MCR.Access.Strategy.Class").orElse("UNDEFINED")
```
sebhofmann: This doesnt work if the Strategy is defined in MCR.Access.Strategy (which is the preferred way).

sebhofmann: Also "UNDEFINED" makes no sense since you can check if the optional is null.

sebhofmann (approved):

sebhofmann (changes requested):
check comments

mycore-base/src/main/java/org/mycore/access/MCRAccessManager.java
rsteph-de: reimplemented property comparison

sebhofmann (approved):

PR #1320, opened 2021-08-23
MCR-2464 Add URI resolver and XSLT functions that allow dynamic lookup of properties

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2464).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/common/xml/MCRPropertiesResolver.java
+++ b/mycore-base/src/main/java/org/mycore/common/xml/MCRPropertiesResolver.java
@@ -1,0 +1,81 @@
+package org.mycore.common.xml;
+
+import org.jdom2.DocType;
+import org.jdom2.Document;
+import org.jdom2.Element;
+import org.jdom2.transform.JDOMSource;
+import org.mycore.common.config.MCRConfiguration2;
+
+import javax.xml.transform.Source;
+import javax.xml.transform.URIResolver;
+
+/**
+ * Resolves the Property values for the given key or key prefix.<br><br>
+ * <br>
+ * Syntax: <code>property:{key-prefix}*</code> or <br>
+ *         <code>property:{key}</code>
+ * <br>
+ * Result for a key prefix: <code> <br>
+ *     &lt;!DOCTYPE properties SYSTEM "http://java.sun.com/dtd/properties.dtd"&gt; <br>
+ *     &lt;properties&gt; <br>
+ *   &lt;entry key=&quot;key1&quot;&gt;value1&lt;/property&gt; <br>
+ *   &lt;entry key=&quot;key2&quot;&gt;value2&lt;/property&gt; <br>
+ *   &lt;entry key=&quot;key3&quot;&gt;value3&lt;/property&gt; <br>
+ * &lt;/properties&gt; <br>
+ * </code>
+ * If no entries with the given key prefix exist, a properties element without entry elements is returned.
+ * <br/>
+ * Result for a key: <code> <br>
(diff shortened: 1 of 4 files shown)


fluetze (changes requested):
Please fix the JavaDocs, the sentence
"If no entries with the given key prefix exists, no a properties elements with no properties is returned."

toKrause:
> Please fix the JavaDocs, the sentence
> "If no entries with the given key prefix exists, no a properties elements with no properties is returned."

Fixed. That seemes to have been two sentences mixed together.

toKrause:
> I would prefer Java standard property format:
> http://tutorials.jenkov.com/java-collections/properties.html#xml-property-file-format
> 
> We already use it for message properties in REST-API:
> https://www.mycore.de/mir/api/v1/messages?filter=metadata;button&format=xml

Fair, but shouldn't the same apply for the already existing `i18n` URI resolver (after which the proposed URI resolver is modeled)? After all, i18n-messages are typically provided as properties, as your 2nd link shows.

toKrause:
@rsteph-de Should the generated XML contain the `<!DOCTYPE properties SYSTEM "http://java.sun.com/dtd/properties.dtd">` line? As far as i can see, this would be the only URI resolver that returns a document with a doc type. Will this cause problems somewhere else?

toKrause:
@rsteph-de Actually, since `MCRURIResolver#resolve(String)` will return the root element of any document provided by a concrete `MCRURIResolver`s `resolve(String, String)` method (and therefore discard the document node and, if set, the document type node), it would be futile to add a document type.

By the same token, it will be impossible to consistently provide a compliant `properties` document trour URI resolvers. How to proceed?

fluetze (dismissed):

rsteph-de:
> Fair, but shouldn't the same apply for the already existing `i18n` URI resolver (after which the proposed URI resolver is modeled)? After all, i18n-messages are typically provided as properties, as your 2nd link shows.

Good Point! We should disscuss this in our next TelKo!

rsteph-de:
> Should the generated XML contain the `<!DOCTYPE properties SYSTEM "http://java.sun.com/dtd/properties.dtd">` line?

I guess, since our XML validatidation is mainly based on XMLSchema this entry would be ignored.
If you look into https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Properties.html#loadFromXML(java.io.InputStream)
it states, that de DTD should be present. But at the moment I do not see a scenario where we use the XSLT-Output as direct input to the Java Properties class.

I think, if the DTD causes problems, we could omit the line and comply with the XML structure, without making it explicite.
A short note in the Javadocs, that the generated XML complies with the Java Properties XML syntax might be useful.

kkrebs:
please rebase to 2021.06.x and change target branch of pull request to 2021.06.x

toKrause:
```
First, rewinding head to replay your work on top of it...
Applying: [maven-release-plugin] prepare for next development iteration
Applying: [maven-release-plugin] prepare release v2021.08
Applying: [maven-release-plugin] prepare for next development iteration
Applying: MCR-2464 Add URI resolver and XSLT functions that allow dynamic lookup of properties
```

Rebasing from `main` to `2021.06.x` also included the three `[maven-release-plugin] ...` commits that weren't created by me but were already part of `main` but not of `2021.06.x`, .... what to do?

toKrause:
To resolve the problem, I created a new branch an cherry-picked my commits before re-pushing. Don't know if this is the proper way.

kkrebs (approved):

PR #1329, opened 2021-08-27
MCR-2469 cryptpackage and cryptresolver

PaulBorchert:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2469).


toKrause:
```
    @MCRProperty(name = "KeyFile", required = true)
    public void setRulesURI(String path) {
        keyFile = path;
    }
```
Seems oddly named :)

kkrebs:
please rebase to branch 2021.06.x

PaulBorchert:
> 
> 
> ```
>     @MCRProperty(name = "KeyFile", required = true)
>     public void setRulesURI(String path) {
>         keyFile = path;
>     }
> ```
> 
> Seems oddly named :)

The Attribut is named like the property. Ex.:
MCR.Crypt.Cipher.abstract.KeyFile=/mcr/test/secret/exampleKeyFile.secret

Possible solution: rename the property to KeyFilePath?

toKrause:
I was talking about the `setRulesURI` method name.

mycore-base/src/main/java/org/mycore/crypt/MCRAESCipher.java
```diff
@@ -0,0 +1,165 @@
...
+            Files.writeString(FileSystems.getDefault().getPath(keyFile), cryptKey);
+        } catch (NoSuchAlgorithmException e) {
+            throw new MCRException("Error while generating keyfile.", e);
+        } catch (IOException e) {
+            throw new MCRException("Error while write key to file.", e);
+        }
+    }
+    
+    private String generateKey () throws NoSuchAlgorithmException {
+        SecretKey tmpSecretKey = KeyGenerator.getInstance("AES").generateKey();
```
sebhofmann, line 134: The Exception should be handled in this function. In this case it is actually better to throw a runtime Exception like  MCRCryptException, because you can not actually handle this in MCRAESCipher.

mycore-base/src/main/java/org/mycore/crypt/MCRAESCipher.java
```diff
@@ -0,0 +1,165 @@
...
+    }
+    
+    public void generateKeyFile () throws FileAlreadyExistsException {
+        try {
+            LOGGER.info("generate Key File");
+            String cryptKey = generateKey ();
+            Files.writeString(FileSystems.getDefault().getPath(keyFile), cryptKey, StandardOpenOption.CREATE_NEW);
+        } catch (NoSuchAlgorithmException e) {
+            throw new MCRException("Error while generating keyfile.", e);
+        } catch (FileAlreadyExistsException e) {
```
sebhofmann, line 111: I would not catch the FileAlreadyExistsException, because we do not really get more information, by creating our own FileAlreadyExistsException.

mycore-base/src/main/java/org/mycore/crypt/MCRAESCipher.java
```diff
@@ -0,0 +1,165 @@
...
+                    + " not found. Generate new one with cli command or copy file to path."
+                    , e ) ;
+        } catch (IOException e) {
+            throw new MCRException ("Can't read keyFile " + keyFile + ".",e) ; 
+        } catch (NoSuchAlgorithmException | NoSuchPaddingException e) {
+            throw new MCRCryptCipherConfigurationException (
+                    "The algorithm AES/ECB/PKCS5PADDING ist not provided by this javaversion."
+                    + "Update Java or configure an other chipher in mycore.properties.", e);
+        } catch (InvalidKeyException e) {
+            throw new MCRCryptInvalidKeyException("Please ensure that keyfile ist correct or generate new one.",e);
```
sebhofmann: No reason to have a own Runtime MCRCryptInvalidKeyException.

mycore-base/src/main/java/org/mycore/crypt/MCRCipher.java
```diff
@@ -0,0 +1,103 @@
...
+    abstract public void generateKeyFile() throws FileAlreadyExistsException;
+    
+    /**
+     * Generate the secret key an write it to the keyfile. Overwrites
+     * exsisting keyfile.   
+     */
+    abstract public void overwriteKeyFile() ;
+    
+    
+    public String encrypt(String text) throws MCRCryptKeyNoPermission {
```
sebhofmann: The MCRCryptKeyNoPermission is not thrown anywhere, it should be removed or you throw it in the case you have no permission.

mycore-base/src/main/java/org/mycore/crypt/MCRAESCipher.java
```diff
@@ -0,0 +1,165 @@
...
+            throw new MCRException("Error while write key to file.", e);
+        }
+    }
+    
+    private String generateKey () throws NoSuchAlgorithmException {
+        SecretKey tmpSecretKey = KeyGenerator.getInstance("AES").generateKey();
+        return java.util.Base64.getEncoder().encodeToString(tmpSecretKey.getEncoded());
+    }
+    
+    protected String pencrypt(String text) throws MCRCryptCipherConfigurationException {
```
sebhofmann: A better name for pencrypt would be encryptImpl.

mycore-base/src/main/java/org/mycore/crypt/MCRAESCipher.java
```diff
@@ -0,0 +1,165 @@
...
+            encodedKey = Files.readString(FileSystems.getDefault().getPath(keyFile));
+            byte[] decodedKey = java.util.Base64.getDecoder().decode(encodedKey);
+            LOGGER.info("Set secret key");
+            secretKey = new SecretKeySpec(decodedKey, 0, decodedKey.length, "AES");
+            encryptCipher = Cipher.getInstance("AES/ECB/PKCS5PADDING");
+            encryptCipher.init(Cipher.ENCRYPT_MODE, secretKey);
+            decryptCipher = Cipher.getInstance("AES/ECB/PKCS5PADDING");
+            decryptCipher.init(Cipher.DECRYPT_MODE, secretKey);
+        } catch (NoSuchFileException e) {
+            throw new MCRCryptKeyFileNotFoundException (
```
sebhofmann, line 80: Better throw NoSuchFileException here, because its a CheckedException or make MCRCryptKeyFileNotFoundException a Checked Exception

mycore-base/src/main/java/org/mycore/crypt/MCRCipherManager.java
```diff
@@ -0,0 +1,83 @@
...
+    /**
+     * Create a single instance of a MCRCipher and initialize it. If the cipher can't initialized an exception
+     * will thrown. Common issue is an missing key. In this case the methods throws 
+     * a MCRCryptKeyFileNotFoundException. 
+     * 
+     * @param id ID of cipher as configured
+     * @return instance of cipher
+     */
+    
+    public static MCRCipher getCipher(String id) throws MCRCryptKeyFileNotFoundException {
```
sebhofmann, line 47: Better throw NoSuchFileException here and explain it in the Javadocs.

mycore-base/src/main/java/org/mycore/crypt/MCRCipherManager.java
```diff
@@ -0,0 +1,83 @@
...
+        LOGGER.debug("getCipher for id {} .", id );
+        String propertiy = "MCR.Crypt.Cipher." + id + ".class";
+        MCRCipher cipher = MCRConfiguration2.<MCRCipher>getSingleInstanceOf(propertiy)
+                .orElseThrow( 
+                    () -> new MCRCryptCipherConfigurationException (
```
sebhofmann, line 52: Use the MCRConfigurationException or let MCRCryptCipherConfigurationExcetion inherit from it.

mycore-base/src/main/java/org/mycore/crypt/MCRCryptCommands.java
```diff
@@ -0,0 +1,160 @@
...
+     *            String id of cipher configured in properties
+     */
+    @MCRCommand(syntax = "decrypt {0} with cipher {1}",
+        help = "The command encrypt the value with cipher.",
+        order = 40)
+    public static void decrypt(String value, String cipherid) {
+        try {
+            MCRCipher cipher = MCRCipherManager.getCipher(cipherid);
+            LOGGER.info("Decrypted Value: " + cipher.decrypt(value));
+        } catch (MCRCryptKeyNoPermission e) {
```
sebhofmann: I think it is better to just throw the Exception, so the command will end with a error, if it can not decrypt the key.

sebhofmann (changes requested):

mycore-base/src/main/java/org/mycore/crypt/MCRAESCipher.java
PaulBorchert: The methods is renamed

mycore-base/src/main/java/org/mycore/crypt/MCRAESCipher.java
PaulBorchert: MCRCryptKeyFileNotFoundException now inherits from NoSuchFileException

mycore-base/src/main/java/org/mycore/crypt/MCRAESCipher.java
PaulBorchert: Yes right

mycore-base/src/main/java/org/mycore/crypt/MCRAESCipher.java
PaulBorchert: I want enrich the exception with reason why it is ab bad idea to overwrite a existing key with new one.

mycore-base/src/main/java/org/mycore/crypt/MCRAESCipher.java
PaulBorchert: If it's handled in this method, the system can't react to an missed Algorithm. Possible scenario:

There are 2 Ciphers implemented:
Chipher A : Very secure but rare. (Perhaps uses hardware for coding)
Chipher B : Not so secure but on all systems available.

The key is configured in the way, that it use Cipher A and if it not available Cipher B as fallback.
In the case of an enviroment unable to generate a KeyFile with Chipher A, it catches the Exception and use the fallback cipher.

mycore-base/src/main/java/org/mycore/crypt/MCRCipher.java
PaulBorchert: The Exception will be renamed and used.

mycore-base/src/main/java/org/mycore/crypt/MCRCipherManager.java
PaulBorchert: MCRCryptKeyFileNotFoundException now inherits from NoSuchFileException

mycore-base/src/main/java/org/mycore/crypt/MCRAESCipher.java
sebhofmann: The method is private and you only use it in:
- generateKeyFile where you just rethrow it as a RuntimeException
- overwriteKeyFile where you just rethrow it as a RuntimeException

I think you can just throw the configuration exception here.

mycore-base/src/main/java/org/mycore/crypt/MCRCryptCommands.java
PaulBorchert: will changed

sebhofmann (approved):

PR #1339, opened 2021-09-15
MCR-2472 Now alto highlighting will check for empty string and don't try split…

chiliego:
Now alto highlighting will check for empty string and don't try splitting them

[Link to jira](https://mycore.atlassian.net/browse/MCR-2472).

### DIFF
--- a/mycore-mets/src/main/java/org/mycore/mets/resource/MCRAltoHighlightResource.java
+++ b/mycore-mets/src/main/java/org/mycore/mets/resource/MCRAltoHighlightResource.java
@@ -170,7 +170,17 @@ public class MCRAltoHighlightResource {
             JsonArray positions = new JsonArray();
             int wordCount = StringUtils.countMatches(altoContent, "<em>");
             for (int i = wordIndex; i < wordIndex + wordCount; i++) {
-                positions.add(buildPositionObject(altoWords.get(i)));
+                String altoWord = altoWords.get(i);
+
+                if(!altoWord.equals("")){
+                    JsonObject positionObject = buildPositionObject(altoWord);
+
+                    if(!positionObject.entrySet().isEmpty()){
+                        positions.add(positionObject);
+                    }
+                } else if(wordCount < altoWords.size()){
+                    wordCount++;
+                }
             }
             wordIndex += wordCount;
             hit.add("positions", positions);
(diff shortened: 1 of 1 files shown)


mycore-mets/src/main/java/org/mycore/mets/resource/MCRAltoHighlightResource.java
```diff
@@ -170,7 +170,17 @@ private JsonArray buildQueryResponse(Map<String, Map<String, List<String>>> high
...
-                positions.add(buildPositionObject(altoWords.get(i)));
+                String altoWord = altoWords.get(i);
+
+                if(!altoWord.equals("")){
+                    JsonObject positionObject = buildPositionObject(altoWord);
+
+                    if(!positionObject.entrySet().isEmpty()){
+                        positions.add(positionObject);
+                    }
+                } else if(wordCount < altoWords.size()){
```
rsteph-de, line 181: The if-part looks O.K. for me - I used a similiar solution to temporarily fix the problem  in my own application
Can you please explain, what you are doing in the else-Part?

mycore-mets/src/main/java/org/mycore/mets/resource/MCRAltoHighlightResource.java
```diff
@@ -187,11 +197,13 @@ private JsonObject buildPositionObject(String altoWord) {
         JsonObject positionObject = new JsonObject();
         String plainWord = altoWord.replaceAll("<em>|</em>", "");
         String[] split = plainWord.split("\\|");
-        positionObject.addProperty("content", split[0]);
-        positionObject.addProperty("xpos", Integer.parseInt(split[1]));
-        positionObject.addProperty("vpos", Integer.parseInt(split[2]));
-        positionObject.addProperty("width", Integer.parseInt(split[3]));
-        positionObject.addProperty("height", Integer.parseInt(split[4]));
+        if(split.length == 5){
```
rsteph-de, line 200: Is this "if" still necessary if we filter empty word before?
But it "does not hurt", if we keep it this way.

rsteph-de (commented):
some questions ...

mycore-mets/src/main/java/org/mycore/mets/resource/MCRAltoHighlightResource.java
chiliego: As you can see in the Jira ticket, the fix will address the issue with Solr 8.8 and higher. For responses from Solr with `"hl.method":"unified"` you will get `"alto_words" : ["", "<em>some alto</em"]`. In this case the wordCount is not correct and you have to increase it to get the alto. Maybe there will be a better solution, this is what came first to my mind to fix the issue. I don't know if it will cause errors for some other unkown cases.

mycore-mets/src/main/java/org/mycore/mets/resource/MCRAltoHighlightResource.java
chiliego: I put the if for some robustness, for reasons where the incomming string can be corrupt. Throwing an exeption when string is not empty and length != 5 is another option, maybe the better option.

rsteph-de (approved):
looks good for me, tested in my own branch

PR #1342, opened 2021-09-20
MCR-2476 lint line break issue with access key manager

golsch:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2476).


rsteph-de (approved):
compiles under Windows now!

PR #1349, opened 2021-09-27
MCR-2483 option to list user with user attribute name search pattern and limit/offset

golsch:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2483).

### DIFF
--- a/mycore-user2/src/main/java/org/mycore/user2/MCRUserManager.java
+++ b/mycore-user2/src/main/java/org/mycore/user2/MCRUserManager.java
@@ -38,6 +38,7 @@ import javax.persistence.criteria.Predicate;
 import javax.persistence.criteria.Root;
 import javax.persistence.metamodel.SingularAttribute;
 
+import org.apache.commons.lang.StringUtils;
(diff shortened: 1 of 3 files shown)


mycore-user2/src/main/java/org/mycore/user2/MCRUserManager.java
```diff
@@ -335,8 +336,18 @@ public static void deleteUser(MCRUser user) {
             .getResultList();
     }
 
+    private static String buildSearchPattern(String searchPattern) {
+        searchPattern = searchPattern.replace('*', '%');
+        searchPattern = searchPattern.replace('?', '_');
+        return searchPattern.toLowerCase(MCRSessionMgr.getCurrentSession().getLocale());
+    }
+
+    private static boolean isValidSearchPattern(String searchPattern) {
```
rsteph-de, line 346: Why do we not use  java.commons.lang.StringUtils.isNotEmpty() instead?

rsteph-de (approved):
looks good for me now

PR #1363, opened 2021-11-15
MCR-2507 Add touch commands

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2507).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/frontend/cli/MCRDeveloperCommands.java
+++ b/mycore-base/src/main/java/org/mycore/frontend/cli/MCRDeveloperCommands.java
@@ -156,7 +156,7 @@ public class MCRDeveloperCommands {
 
     @MCRCommand(
             syntax = "touch object {0}",
-            help = "Touch object with id {0}",
+            help = "Load and update object with id {0} without making any modifications",
             order = 80
     )
     public static void touchObject(String id) {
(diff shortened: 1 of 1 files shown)


mycore-base/src/main/java/org/mycore/frontend/cli/MCRDeveloperCommands.java
```diff
@@ -149,4 +153,37 @@ public static void showResource(String uri) {
...
+    @MCRCommand(
+            syntax = "touch object {0}",
+            help = "Touch object with id {0}",
+            order = 80
+    )
+    public static void touchObject(String id) {
+        try {
+            MCRObjectID objectId = MCRObjectID.getInstance(id);
+            MCRObject object = MCRMetadataManager.retrieveMCRObject(objectId);
+            MCRMetadataManager.update(object);
```
kkrebs, line 166: the repair command calls here:
MCRMetadataManager.fireRepairEvent(obj);

So in difference to our repair commands not only all event handlers are fired again. The additional code handling the structure section is also running in this case.

kkrebs (changes requested):
We should think more deeply about this pull request. Do we need this command additional to our existing repair commands? What is the added value?

toKrause:
Could we leave the discussion at a single location (JIRA issue or here)? Otherwise it gets really confusing.

Yes, the proposed commands do more than firing an "repair event"; exactly my point. The whole point of this _developer command_ is to _perform an update_, with everything that is part of that process. This is helpful during debugging or to check the impacts of modifications to the software. If it would only fire an event, something important might be left out. 

Also, again, an "update event" is presumably or at least potentially handled differently then a "repair event", so i still don't understand what the repair events and the repair commands have to do with the proposed commands.

sebhofmann (approved):

yagee-de (approved):

PR #1364, opened 2021-11-15
MCR-2506 Add command to select objects by XPath

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2506).


kkrebs (approved):

kkrebs:
perhaps we should add a "use with care" hint?! We have applications with more than a million objects. Perhaps in such cases listIDs() leads to problems?!

PR #1365, opened 2021-11-15
MCR-2508 Extend existing commands to allow importing of unknown objects and derivates

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2508).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/datamodel/metadata/MCRMetadataManager.java
+++ b/mycore-base/src/main/java/org/mycore/datamodel/metadata/MCRMetadataManager.java
@@ -180,7 +180,7 @@ public final class MCRMetadataManager {
 
         // assign new id if necessary
         if (derivateId.getNumberAsInteger() == 0) {
-            derivateId = MCRObjectID.getNextFreeId(derivateId.getBase(), derivateId.getTypeId());
+            derivateId = MCRObjectID.getNextFreeId(derivateId.getBase());
             mcrDerivate.setId(derivateId);
             LOGGER.info("Assigned new derivate id {}", derivateId);
         }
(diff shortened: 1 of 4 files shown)


kkrebs:
We now have several places in our code to deal with zero IDs. Perhaps this should be part of the MCRMetadataManager.create() method?! So we have to add the same code every time before we call create():
- REST-API
- MCRCreateObjectServlet
- and now also via CLI

toKrause:
This certainly sounds like a good idea. Could this be al follow-up or should I update the Jira issue and modify this PR?

kkrebs:
update the Jira issue and modify this PR would be great. It should be no problem to leave the zero id code in REST-API and MCRCreateObjectServlet for now. We could clean the code as part of an follow up ticket. But we should not add more ...

toKrause:
I've moved the code and made some minor cleanup. A followup issue should be created when this PR is merged.

toKrause:
I've noticed that almost all calls to static methods of `MCRMetadataManager` inside of `MCRMetadataManager` are unnecessarily qualified (i.e. `MCRMetadataManager.create(mcrDerivate);`). Is that a local style that should be adhered to?

mycore-base/src/main/java/org/mycore/datamodel/metadata/MCRMetadataManager.java
```diff
@@ -296,16 +306,26 @@ private static void importDerivate(String derivateID, Path sourceDir) throws MCR
...
         // exist the object?
-        if (MCRMetadataManager.exists(mcrObject.getId())) {
-            throw new MCRPersistenceException("The object " + mcrObject.getId() + " allready exists, nothing done.");
+        if (MCRMetadataManager.exists(objectId)) {
+            throw new MCRPersistenceException("The object " + objectId + " allready exists, nothing done.");
+        }
+
+        // assign new id if necessary
+        if (objectId.getNumberAsInteger() == 0) {
+            objectId = MCRObjectID.getNextFreeId(objectId.getBase(), objectId.getTypeId());
```
sebhofmann: I think this has to be .getNextFreeId(objectId.getProjectId(), ...

Also the parameter name of getNextFreeId is wrong. It should be projectId. (has nothing to do with this pr, but is probably the cause of the issue)

sebhofmann (changes requested):

toKrause:
Sorry, changed the wrong PR to draft.

sebhofmann (approved):

PR #1368, opened 2021-11-16
MCR-2509 Some minor improvements of the OAI module

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2509).

### DIFF
--- a/mycore-oai/src/main/resources/META-INF/resources/oai/oai2.xsl
+++ b/mycore-oai/src/main/resources/META-INF/resources/oai/oai2.xsl
@@ -597,6 +597,10 @@ p.intro {
       <xsl:apply-templates select="*" />
     </table>
   </div>
+  <h3>XML-Source</h3>
+  <div class="xmlSource">
+    <xsl:apply-templates select="." mode='xmlMarkup' />
+  </div>
 </xsl:template>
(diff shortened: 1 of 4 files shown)


mycore-oai/src/main/resources/META-INF/resources/oai/oai2.xsl
```diff
@@ -512,7 +512,7 @@ p.intro {
   <xsl:apply-templates select="oai:setSpec" />
   </table>
   <xsl:if test="@status='deleted'">
-    <p>This record has been deleted.</p>
```
rsteph-de, line 515: oai2.xsl is the nearly unchanged file as it was provided by eprints.
I am not sure, if we should start modifying texts in it.

toKrause: There have already been multiple modifications to the file, so this shouldn't be any additional issue. But it would answering potential questions preemptively.

Has there ever been an effort to keep the version of this file used by MyCoRe in sync with the version used by EPrints and to backport changes made by the EPrints community?

rsteph-de (approved):
After todays discussion I agree to these changes!

PR #1372, opened 2021-11-17
MCR-2514 Unexpected exceptions cause unneccessary null-Document results

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2514).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/common/content/MCRSourceContent.java
+++ b/mycore-base/src/main/java/org/mycore/common/content/MCRSourceContent.java
@@ -18,12 +18,12 @@
 
 package org.mycore.common.content;
 
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.StringReader;
-import java.io.StringWriter;
-import java.net.MalformedURLException;
-import java.net.URL;
+import org.jdom2.Document;
+import org.jdom2.Element;
+import org.jdom2.transform.JDOMSource;
+import org.mycore.common.MCRException;
+import org.mycore.common.xml.MCRURIResolver;
+import org.w3c.dom.Node;
 
 import javax.xml.bind.util.JAXBSource;
 import javax.xml.transform.Source;
@@ -34,15 +34,11 @@ import javax.xml.transform.dom.DOMSource;
 import javax.xml.transform.sax.SAXSource;
 import javax.xml.transform.stream.StreamResult;
 import javax.xml.transform.stream.StreamSource;
-
-import org.jdom2.Document;
-import org.jdom2.Element;
-import org.jdom2.JDOMException;
-import org.jdom2.input.SAXBuilder;
-import org.jdom2.transform.JDOMSource;
-import org.mycore.common.MCRException;
-import org.mycore.common.xml.MCRURIResolver;
-import org.w3c.dom.Node;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.MalformedURLException;
+import java.net.URL;
 
 /**
  * @author Thomas Scheffler (yagee)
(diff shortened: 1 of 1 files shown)


fluetze (changes requested):
This will change the API of MCRContent. There still should be a default asXML() method.
Thomas will have a look at the root cause.

mycore-base/src/main/java/org/mycore/common/content/MCRContent.java
```diff
@@ -260,11 +260,20 @@ protected MCRDataURLEncoding getDataURLEncoding() throws IOException {
 
     /**
      * Parses content, assuming it is XML, and returns the parsed document.
-     * 
+     *
```
toKrause, line 263: I fail to see the mentioned API change. 

Actually, there is another API change, because some child classes had deviating lists of thrown exception classes. That's why there are changes to, say, `MCRMailer`.

mycore-base/src/main/java/org/mycore/common/content/MCRContent.java
```diff
@@ -260,11 +260,20 @@ protected MCRDataURLEncoding getDataURLEncoding() throws IOException {
...
+     * @return the XML document parsed from content
+     */
+    public final Document asXML() throws JDOMException, IOException, SAXException {
```
toKrause: I fail to see the mentioned API change.

Actually, there is an API change, because some child classes had deviating lists of thrown exception classes. That's why there are changes to, say, `MCRMailer`. This also effects classes in MIR and is one of the reasons why this PR is only a draft.

sebhofmann: Why not make asXML() not final and just overwrite it in the Classes which have less Exceptions?

toKrause: Because  `asXML()` is now only a convenience method that provides a default parameter for the actual method `asXML(boolean)`.

As a rule, I make all convenience methods final. This removes a common and subtle pitfall.

Making convenience methods final enforces that subclasses that wish to overwrite the functionality have to override the actual method. Otherwise it would be easy to accidentally overwrite the convenience method and create inconsistent behavior.

toKrause: Another question is whether clients of `asXML(boolean)` and `asXML(boolean)` actually care about the exact exception class or if it would suffice for them to know that exceptions are to be expected (i.e. replace with `throws Exception`).

mycore-base/src/main/java/org/mycore/common/content/MCRSourceContent.java
```diff
@@ -76,6 +84,17 @@ public MCRSourceContent(Source source) {
...
+            try {
+                JAXBSource src = (JAXBSource) source;
+                StringWriter writer = new StringWriter();
+                TransformerFactory transformerFactory = TransformerFactory.newInstance();
+                Transformer transformer = transformerFactory.newTransformer();
+                transformer.transform(src, new StreamResult(writer));
+                baseContent = new MCRJDOMContent(new SAXBuilder().build(new StringReader(writer.toString())));
+            } catch (TransformerException | JDOMException | IOException e) {
+                throw new MCRException("Error while resolving JAXBSource", e);
+            }
```
yagee-de: Should be lazier, as we do not know what code runs after it (Does it need to be parsed?):

```java
TransformerFactory transformerFactory = TransformerFactory.newDefaultInstance();
try {
    Transformer transformer = transformerFactory.newTransformer();
    ByteArrayOutputStream bout = new ByteArrayOutputStream();
    transformer.transform(source, new StreamResult(bout));
    baseContent = new MCRByteContent(bout.toByteArray());
} catch (TransformerException e) {
    throw new MCRException("Error while resolving JAXBSource", e);
}

```

yagee-de (changes requested):

sebhofmann (approved):

kkrebs (approved):

PR #1373, opened 2021-11-22
MCR-2510 - Revision Attribute in Root

Hammer1279:
- object head gets checked for deleted status
- new attribute gets added to the root element of a object, that includes the revision number

### DIFF
--- a/mycore-base/src/main/java/org/mycore/datamodel/common/MCRDefaultXMLMetadataManager.java
+++ b/mycore-base/src/main/java/org/mycore/datamodel/common/MCRDefaultXMLMetadataManager.java
@@ -41,12 +41,15 @@ import java.util.stream.Stream;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.jdom2.Document;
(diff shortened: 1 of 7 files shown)


sebhofmann:
The second part (a01dfff) is nice.

The revision will be only added if you request a specific Revision, that could also be done in MCRMetadataManager and it would be more generic. 

Did you check if the revision attribute will be stored if you request a Revision and save it? The MCRMetadataManager could filter that.

Hammer1279:
@sebhofmann The revision is only added there to be passed along, using the existing framework of SVN and will not be saved, as when editing a file, the editor will default to the newest version.

What is also checked is restoring, if a prior version is restored, the rev attribute is not getting added.

As for https://github.com/MyCoRe-Org/mycore/pull/1373/commits/47d26c1f4da8f15b849570946d1266fc26e1cf14, as discussed for the reasons of restoring deleted versions, the head change got reverted for now to allow viewing of deleted files using the revision attribute.

sebhofmann (approved):

sebhofmann (approved):

PR #1379, opened 2021-11-26
MCR-2523 allow to filter for works created by this application

chemicus:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2523).


fluetze (commented):
Could you please comment, here or in the Jira ticket, why there is a need for this? Use case? Thx!

chemicus:
In the current implementation findWork(MCRObject obj) returns the first work found in the ORCID profile matching the specified mods:identifiers. This is not necessarily a work that has been created by the respective MyCoRe-application, which we might want to update. Therefore, with the current implementation it is not possible to reliably detect whether or not we have already synced a publication with ORCID or not.
With the addition of findOwnWork(...) we can restrict the found works to those being created by us. We can therefore decide whether we want/need to add/update the work on the remote site.
All in all this commit does not introduce any changes to the existing interface, it simply adds functionality which might be useful (and is required for my implementation).

fluetze (approved):

PR #1382, opened 2021-11-26
MCR-2526 do not rely on object id in matching

chemicus:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2526).


fluetze (commented):
Same question as with MCR-2527:
Could you please explain (here - or in the Jira ticket) why works from other sources should match? I assumed that our application must not update or delete works from other sources.
What happens if we do so? Do we have the permission?

chemicus:
See my comment in MCR-2527. If MCR-2527 is accepted, MCR-2526 can be deleted since it does not contain any changes not already present in MCR-2527.

kkrebs (approved):

PR #1383, opened 2021-11-26
MCR-2522 allow to limit user selection to particular name identifier type

chemicus:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2522).

### DIFF
--- a/mycore-orcid/src/main/java/org/mycore/orcid/user/MCRORCIDUser.java
+++ b/mycore-orcid/src/main/java/org/mycore/orcid/user/MCRORCIDUser.java
@@ -138,14 +138,12 @@ public class MCRORCIDUser {
     public static Set<String> getNameIdentifierKeys(MCRMODSWrapper wrapper) {
         Set<String> identifierKeys = new HashSet<>();
 
-        List<Element> nameIdentifiers = null;
-        if(MATCH_ONLY_NAME_IDENTIFIER.isEmpty()) {
-            nameIdentifiers = wrapper.getElements("mods:name/mods:nameIdentifier");
-        } else {
-            nameIdentifiers = wrapper
-		    .getElements("mods:name/mods:nameIdentifier[@type = '" + MATCH_ONLY_NAME_IDENTIFIER + "']");
-        }
-        for (Element nameIdentifier : nameIdentifiers) {
+        String xPath = "mods:name/mods:nameIdentifier";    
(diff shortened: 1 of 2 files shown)


fluetze (commented):
Could you please explain, here or in the Jira ticket, why one might wish to limit a match to only certain name identifier types? If so, wouldn't it be better to be able to specify more than one type, and not just one or all?

chemicus:
In the current implementation all nameIdentifiers are used, regardless of being verified or not. The ORCID guidelines require, however, that users need to approve that a work that has been assigned to them, really "belongs" to them. Therefore we should only sync publications where the mods:nameIdentifier[@type='orcid'] has been set, thus the need to limit the type(s) of identifier(s).
I do not see a necessity to limit to more than one identifier at the moment, although one could think about that of course. A general solution would be more complicated, however, since the boolean operator (and, or, ...) between different identifiers would need to be specified as well.
I would therefore suggest to accept the current modification for now, since this is already an improvement. If we see a need for a more general solution, I can certainly add that later.

fluetze (commented):
Refactoring ok?

chemicus:
Look ok to me!

fluetze (dismissed):

mycore-orcid/src/main/java/org/mycore/orcid/user/MCRORCIDUser.java
```diff
@@ -134,7 +138,12 @@ public boolean isMyPublication(MCRObjectID oid) {
     public static Set<String> getNameIdentifierKeys(MCRMODSWrapper wrapper) {
         Set<String> identifierKeys = new HashSet<>();
 
-        List<Element> nameIdentifiers = wrapper.getElements("mods:name/mods:nameIdentifier");
+        String xPath = "mods:name/mods:nameIdentifier";
+        if (! MATCH_ONLY_NAME_IDENTIFIER.isEmpty()) {
+            xPath += "[@type = '" + MATCH_ONLY_NAME_IDENTIFIER + "']";
+        }
+
+        List<Element> nameIdentifiers = wrapper.getElements(xPath));
```
rsteph-de: Syntax error - because of doubled ")" at the end !!!

rsteph-de (dismissed):
Build failed!

fluetze (dismissed):

kkrebs (approved):

PR #1384, opened 2021-11-26
MCR-2527 do not rely on object id being stored in work

chemicus:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2527).

### DIFF
--- a/mycore-orcid/src/main/java/org/mycore/orcid/resources/MCRORCIDResource.java
+++ b/mycore-orcid/src/main/java/org/mycore/orcid/resources/MCRORCIDResource.java
@@ -130,7 +130,7 @@ public class MCRORCIDResource {
             } else if (!status.isInORCIDProfile()) {
                 works.addWorkFrom(oid);
             } else if (status.isInORCIDProfile()) {
-                works.findWork(oid).get().update();
+                works.findWork(oid).get().update(oid);
             }
 
             return publicationStatus(oid, user);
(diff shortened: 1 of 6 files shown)


fluetze:
Could you please explain (here - or in the Jira ticket) why there might be a work without mcrobjectid in the works section - that we wish to update or delete.
I assumed that our application must not update or delete works from other sources.
What happens if we do so? Do we have the permission?
For example, I think a work in orcid profile shouldn't be deleted when it is deleted in the MyCoRe repository - unless this repository is the only source of the work.

chemicus:
The current implementation relies on the presence of the mcrobjectid stored as common:external-id[@common:external-id-type='source-work-id'] on the remote site. While the current implementation makes sure that this id is always present, this solution works only as long as the local mcrobjectid remains unchanged. This approach might even lead to wrong local/remote matching when the local mcrobjectid changes and the old mcrobjectid refers to a different object afterwards. This might happen when objects are being merged and new objects are being created with ids that have been assigned previously (this does happen!). Furthermore we might want to keep the mcrobjectid private to our application and thus do not export it ORCID (which is not mandatory).
All in all there are good reasons to not export the mcrobjectid to ORCID at all and as shown in this commit, there is not even a necessity since all information for syncing is present locally. The only tiny drawback is the loss of the info message in the delete(MCRWork work) function, which is not used anywhere anyway.

fluetze (dismissed):

kkrebs (approved):

PR #1387, opened 2021-12-03
MCR-2528 / MCR-2533 - OCFL Layout and Migration

Hammer1279:
### Description
This adds a new Layout Styled after the Native XML Metadata Manager and
Migration from OCFL to XML/SVN* and OCFL to OCFL between different layouts.
*Versions cannot be migrated back into SVN

### New CLI Commands:
- migrate metadata from {0} to {1}
  - migrate from StoreType {0} to {1}
  - Options here are: "xml", "svn", "ocfl"
- migrate metadata from {0} to {1} with {2}
  - same as above but with added RepositoryKey {2}
- export repository {0}
  - export the entire repository {0} to `ocfl-export`
- export object {0} in repository {1}
  - export a single object with ID {0} from repository {1}
- import repository {0}
  - import the entire repository from `ocfl-export`
- restore repository {0}
  - restore OCFL repository {0} from backup
- purge repository backup
  - delete backup directory for freeing space

_All commands that require a repository are also available for the default (active) repository by typing it without the repository part._

### Configuration Changes
Configuration no longer requires Repository to be set, but it can still be overwritten.
`MCR.Metadata.Manager.Repository`
Repository Layout can now be specified to be either "hash" (default) or "mcr"
`MCR.Metadata.Manager.Repository.Layout`

#### New Optional Configuration
`MCR.OCFL.Util.ExportDir` - Directory for Repository Exports
`MCR.OCFL.Util.BackupDir` - Directory for (Migration-) Backups

`MCR.OCFL.Repository.{Repository-Name}` - OCFL Repository Provider to use
`MCR.OCFL.Repository.{Repository-Name}.RepositoryRoot` - Root Directory for Repo
`MCR.OCFL.Repository.{Repository-Name}.WorkDir` - Temp Directory for Repo
`MCR.OCFL.Repository.{Repository-Name}.Layout` - Layout to use for Repo `hash | mcr`

> The Repository `Adapt` is reserved and should not get changed to avoid issues

### Dependency Changes
Bumped OCFL Libary to 1.4.1 (latest) and updated code accordingly to prevent possible problems with the custom Layout

[Link to jira - MCRLayout](https://mycore.atlassian.net/browse/MCR-2528).
[Link to jira - CLI Migration](https://mycore.atlassian.net/browse/MCR-2533).


Hammer1279:
Java-doc build didn't error locally, but I've acknowledged that and going to fix those issues.

mycore-ocfl/src/main/java/org/mycore/ocfl/MCROCFLAdaptionRepositoryProvider.java
```diff
@@ -0,0 +1,283 @@
...
+import edu.wisc.library.ocfl.core.extension.OcflExtensionConfig;
+import edu.wisc.library.ocfl.core.extension.OcflExtensionRegistry;
+import edu.wisc.library.ocfl.core.extension.storage.layout.config.HashedNTupleIdEncapsulationLayoutConfig;
+
+/**
+ * The Adaptation Repository allows conversation between different OCFL Layout Types
+ * @author Tobias Lenhardt [Hammer1279]
+ * @version 1.0
+ */
+public class MCROCFLAdaptionRepositoryProvider extends MCROCFLRepositoryProvider {
```
sebhofmann: A MCROCFLRepositoryProvider should only provide a OcflRepository .
All other methods like export or import should reside in another class.

mycore-ocfl/src/main/java/org/mycore/ocfl/MCROCFLAdaptionRepositoryProvider.java
```diff
@@ -0,0 +1,283 @@
...
+            } catch (IOException e) {
+                throw new MCRException("Error while cleaning directory: ", e);
+            }
+        }
+    });
+
+    public MCROCFLAdaptionRepositoryProvider(String repositoryKey) {
+        manager = new MCROCFLXMLMetadataManager();
+        manager.setRepositoryKey(repositoryKey);
+        setRepositoryRoot(MCRConfiguration2.getStringOrThrow("MCR.OCFL.Repository.Adapt.RepositoryRoot"));
```
sebhofmann: Properties for a MCROCFLRepositoryProvider should be loaded with the [@MCRProperty() annotation](https://www.mycore.de/documentation/basics/basics_configurable_instance/).

mycore-ocfl/src/main/java/org/mycore/ocfl/MCROCFLAdaptionRepositoryProvider.java
```diff
@@ -0,0 +1,283 @@
...
+        this.mainRoot = Paths.get(MCRConfiguration2.getStringOrThrow("MCR.OCFL.Repository.Main.RepositoryRoot"));
+        setExportDir(MCRConfiguration2.getStringOrThrow("MCR.OCFL.Repository.Adapt.ExportDir"));
+        this.backupDir = Paths.get(MCRConfiguration2.getStringOrThrow("MCR.OCFL.Repository.Adapt.BackupDir"));
+        setWorkDir(MCRConfiguration2.getStringOrThrow("MCR.OCFL.Repository.Adapt.WorkDir"));
+        this.layout = MCRConfiguration2.getStringOrThrow("MCR.OCFL.Repository.Adapt.Layout");
+    }
+
+    @Override
```
sebhofmann, line 95: Most of the Code already exists in MCRSimpleOcflRepositoryProvider, why not just extend that class.

mycore-ocfl/src/main/java/org/mycore/ocfl/MCROCFLBaseClass.java
```diff
@@ -0,0 +1,260 @@
...
+import edu.wisc.library.ocfl.api.OcflRepository;
+
+/**
+ * <strong><u>The Base Class</u></strong> for migrating and
+ * converting OCFL Objects into others and some other functions
+ * that are documented per method
+ * @author Tobias Lenhardt [Hammer1279]
+ * @author Sebastian Hofmann
+ */
+public class MCROCFLBaseClass {
```
sebhofmann: Better name would be MCROCFLMigrationUtil or MCROCFLMigrationHelper.

sebhofmann (commented):
I would prefer splitting these tickets in to two different PR. One for the new Extension and one for migration.

kkrebs:
here is the new ticket for commands:
https://mycore.atlassian.net/browse/MCR-2533

bt1cu:
I see not, that we should split the pull request cause one part is the layout structure and the other is the way to it by CLI migration commands. Both are a unit.

yagee-de:
Keep in mind: Using a non-standard repository layout will loose compatibility with other OCFL tools or at least result in HEAVY performance penalties. Take [pwinckles/rocfl](https://github.com/pwinckles/rocfl) as an example. Make the user to choose also adds complexity and requires more testing.

Hammer1279:
I have to partially disagree with that, indeed using non-standard OCFL Layouts will make it loose compatibility with other OCFL tools that don't support custom layouts, but how exactly is using a own layout resulting in HEAVY performance penalties?

By default the HashedNTurpleIDEncapsulation layout is used and trying to navigate through it can be quite complicated on its own without tools, on top of that it creates a lot of folders for one or two files, where the new layout will be nearly exactly how MyCoRe behaved, plus the files for the OCFL repository itself.
It also takes into account for settings such as the SlotLayout and the Pattern of MCRID's.

 For the complexity, by default, no change of any configuration except for enabling / disabling the OCFL MetadataManager is required, defaulting to the before discussed HashedNTurpleIDEncapsulation layout, the layout is fully handled by the OCFL Library and for MyCoRe itself, it doesn't matter what layout has been used.
 With the options it can now easily be adjusted to the own preference and if you don't like it anymore, it is very easy to switch back to the other layout (reason behind that is as above mentioned: the object structure is the same, only the root structure changes)

If someone doesn't like the new layout, it can simply be ignored. This is not a replacement, only a addition to the existing layout.

Looking back at the [Jira Ticket](https://mycore.atlassian.net/browse/MCR-2528) at the [GWDG Pad mentioned](https://pad.gwdg.de/56Tt5gtXSty-qEPdmPTCaQ#Konzept-f%C3%BCr-eigenes-Storage-Layout-Robert), a own layout was already suggested, which was considered too complex, as far as i have read from it (correct me if I'm wrong), due to the need of having multiple layouts, one per Slot-Layout. This layout here has one layout that dynamically reads and applies the Slot-Layout from the config.

A Sample of the structure within the MCRLayout can be seen here: [MCRLayout.md](https://github.com/MyCoRe-Org/mycore/blob/issues/MCR-2528-OCFL_layout_and_migration/mycore-ocfl/src/main/resources/ocfl-specs/MCRLayout.md#structure-sample)
_Keep in mind that the above mentioned Spec-Sheet isn't finalized and will change up to the point of release of these changes._

yagee-de:
> I have to partially disagree with that, indeed using non-standard OCFL Layouts will make it loose compatibility with other OCFL tools that don't support custom layouts, but how exactly is using a own layout resulting in HEAVY performance penalties?

To pick a specific object if the layout is known, the runtime is O(1), when you don't know the layout it is O(n) which is significantly slower. If a tool uses an index (needs a rebuild every time the repository changes) the lookup time can be reduced to O(log(n)) while building the index costs O(n\*log(n)).
If we have a list (of size l) from objects we want to pick (and do something with it), it cost O(l) for a supported layout while it costs O(l\*n) for the unsupported layout (O(l\*log(n)) with index). That's what is usually meant by HEAVY performance hit.

>By default the HashedNTurpleIDEncapsulation layout is used and trying to navigate through it can be quite complicated on its own without tools, on top of that it creates a lot of folders for one or two files, where the new layout will be nearly exactly how MyCoRe behaved, plus the files for the OCFL repository itself.

The "a lot of folder" I can not see. They look to me like slot directories.

>It also takes into account for settings such as the SlotLayout and the Pattern of MCRID's.

Not all OCFL objects in the OCFL repository are going to be MyCoRe objects or derivates!

bt1cu:
I can't understand your comment without practical test. It is only a theory. I think it is a good way to have different storage layouts. So every costumer can decide, which of them he will be use. By the way we will test the performance with around 100.000 datasets (oriental manuscripts). I can't see arguments against this solution. That not all MyCoRe objects are in the current solution is not an argument against this pull request. The pull request is a next step in the roadmap to use OCFL in MyCoRe production systems. It is your choice which layout you use - hash or mcr.

Hammer1279:
See #1431

PR #1395, opened 2021-12-13
MCR-2530 register handle with ePIC api

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2530).


rsteph-de (approved):
looks good as initial commit

PR #1396, opened 2021-12-14
MCR-2538 add option to disable acl checks in MCRCipher

golsch:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2538).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/crypt/MCRAESCipher.java
+++ b/mycore-base/src/main/java/org/mycore/crypt/MCRAESCipher.java
@@ -18,20 +18,16 @@
 
 package org.mycore.crypt;
 
-import org.apache.logging.log4j.LogManager;
-import org.apache.logging.log4j.Logger;
-import org.mycore.common.config.annotation.MCRProperty;
-import org.mycore.common.MCRException;
-
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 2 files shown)


mycore-base/src/main/java/org/mycore/crypt/MCRAESCipher.java
```diff
@@ -55,6 +56,15 @@
     public void setKeyFile(String path) {
         keyFile = path;
     }
+
+    @MCRProperty(name = "EnableACL", required = false)
+    public void setAclEnabled(final String enabled) {
+        if ("true".equalsIgnoreCase(enabled) || "false".equalsIgnoreCase(enabled)) {
+            setAclEnabled(Boolean.valueOf(enabled));
+        } else {
```
sebhofmann, line 64: "false" is not allowed?

mycore-base/src/main/java/org/mycore/crypt/MCRAESCipher.java
```diff
@@ -55,6 +56,15 @@
     public void setKeyFile(String path) {
         keyFile = path;
     }
+
+    @MCRProperty(name = "EnableACL", required = false)
```
sebhofmann: This should be a generic MCRCipher property.

sebhofmann (changes requested):

sebhofmann (approved):

PR #1405, opened 2021-12-15
MCR-2541 moving around namespaces should be done by mycore

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2541).


fluetze (changes requested):
I'd prefer a rename for the property: ...MoveNamespaceDeclarationsToRoot or MoveNamespaceDeclarationsUp. Also, property naming should be same as method naming.

fluetze (approved):

PR #1406, opened 2021-12-15
MCR-2542 Add filter for rest api to use access keys

golsch:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2542).


rsteph-de (approved):
looks good for me

PR #1419, opened 2022-01-10
MCR-2553 Viewer can't display files with single quotes in file path

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2553).

### DIFF
--- a/mycore-viewer/src/main/resources/xsl/IViewConfig-js.xsl
+++ b/mycore-viewer/src/main/resources/xsl/IViewConfig-js.xsl
@@ -107,7 +107,7 @@
   viewerLoader.addRequiredCss(configuration.resources.css);
   viewerLoader.addRequiredScripts(configuration.resources.script, configuration.properties.mobile);
   viewerLoader.addConstructorExecution(function(){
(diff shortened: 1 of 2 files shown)


sebhofmann:
Better use the method [jQuery.escapeSelector](https://api.jquery.com/jQuery.escapeSelector/) for this.
There are also other places where the name needs to be escaped. I can do that later.

toKrause:
> Better use the method [jQuery.escapeSelector](https://api.jquery.com/jQuery.escapeSelector/) for this.

Yes, I know, couldn't get it to work in the TypeScript file though. Presumably some import is needed. Feel free to edit.

sebhofmann:
Is the filepath even assigned to the thumbnail-id? The ID is read from mets (its a ncname) or its a page number from PDF and both should not contain any Symbols which need to be escaped.

toKrause:
I don't know what exactly is happening. I didn't try to understand the whole code. I can only say that the viewer doesn't work, unless both modifications are made.

toKrause:
Closed in favor of https://github.com/MyCoRe-Org/mycore/pull/1445

PR #1423, opened 2022-01-17
MCR-2503 improve MCRAccessCacheManager to invalidate all caches for an object ID

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2503).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/access/MCRAccessCacheEventHandler.java
+++ b/mycore-base/src/main/java/org/mycore/access/MCRAccessCacheEventHandler.java
@@ -66,11 +66,12 @@ public class MCRAccessCacheEventHandler extends MCREventHandlerBase {
     private void clearPermissionCache(MCRObject obj) {
         LOGGER.info("Invalidate permission cache for obj {}", obj.getId());
         MCRAccessManager.invalidAllPermissionCachesById(obj.getId().toString());
-        final String[] ids = Stream.of(obj.getStructure().getChildren().stream(), obj.getStructure().getDerivates().stream())
-                .flatMap(s -> s)
-                .map(MCRMetaLinkID::getXLinkHrefID)
-                .map(MCRObjectID::toString)
-                .toArray(String[]::new);
+        final String[] ids = Stream
+            .of(obj.getStructure().getChildren().stream(), obj.getStructure().getDerivates().stream())
+            .flatMap(s -> s)
+            .map(MCRMetaLinkID::getXLinkHrefID)
+            .map(MCRObjectID::toString)
+            .toArray(String[]::new);
 
         MCRAccessManager.invalidAllPermissionCachesById(ids);
     }
--- a/mycore-base/src/main/java/org/mycore/access/MCRAccessCacheEventHandler.java
+++ b/mycore-base/src/main/java/org/mycore/access/MCRAccessCacheEventHandler.java
@@ -1,0 +1,73 @@
+/*
+ * This file is part of ***  M y C o R e  ***
+ * See http://www.mycore.de/ for details.
+ *
+ * MyCoRe is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * MyCoRe is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with MyCoRe.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.mycore.access;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.mycore.common.events.MCREvent;
+import org.mycore.common.events.MCREventHandlerBase;
+import org.mycore.datamodel.metadata.MCRDerivate;
+import org.mycore.datamodel.metadata.MCRMetaLinkID;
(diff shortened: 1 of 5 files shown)


mycore-base/src/main/java/org/mycore/access/MCRAccessCacheManager.java
```diff
@@ -96,4 +103,26 @@ public void removePermission(String id, String permission) {
         permissionCache.remove(handle);
     }
 
+    public void removePermission(String id) {
```
yagee-de: Should be vararg to support multiple IDs

mycore-base/src/main/java/org/mycore/access/MCRAccessCacheManager.java
```diff
@@ -96,4 +103,26 @@ public void removePermission(String id, String permission) {
...
+
+    private void removePermissionFromCache(String id, MCRCache<MCRPermissionHandle, Boolean> permissionCache) {
+        final List<MCRPermissionHandle> handlesToRemove = permissionCache.keys()
+            .stream()
+            .filter(hdl -> hdl.getId().equals(id))
+            .collect(Collectors.toList());
+        handlesToRemove.forEach(permissionCache::remove);
+    }
+
+    public void removePermissionFromAllCachesById(String id) {
```
yagee-de: Should be vararg to support multiple IDs

mycore-base/src/main/java/org/mycore/access/MCRAccessManager.java
```diff
@@ -306,6 +306,10 @@ public static void invalidPermissionCache(String id, String permission) {
         ACCESS_CACHE.removePermission(id, permission);
     }
 
+    public static void invalidPermissionCacheByID(String id){
```
yagee-de: Should be vararg to support multiple IDs

mycore-base/src/main/java/org/mycore/access/MCRAccessCacheManager.java
```diff
@@ -96,4 +103,26 @@ public void removePermission(String id, String permission) {
         permissionCache.remove(handle);
     }
 
+    public void removePermission(String id) {
+        MCRCache<MCRPermissionHandle, Boolean> permissionCache = accessCache.get();
+        removePermissionFromCache(id, permissionCache);
+    }
+
+    private void removePermissionFromCache(String id, MCRCache<MCRPermissionHandle, Boolean> permissionCache) {
```
yagee-de: id should be varg

yagee-de (commented):
a batch invalidate per MCRSession could be nicer.

mycore-base/src/main/java/org/mycore/access/MCRAccessCacheManager.java
```diff
@@ -96,4 +105,28 @@ public void removePermission(String id, String permission) {
...
+    }
+
+    private void removePermissionFromCache(MCRCache<MCRPermissionHandle, Boolean> permissionCache, String... ids) {
+        Set<String> idSet = Stream.of(ids).collect(Collectors.toSet());
+        final List<MCRPermissionHandle> handlesToRemove = permissionCache.keys()
+            .stream()
+            .filter(hdl-> hdl.getId()!=null)
+            .filter(hdl -> idSet.contains(hdl.getId()))
+            .collect(Collectors.toList());
+        handlesToRemove.forEach(permissionCache::remove);
```
rsteph-de, line 119: Why do you not call forEach on the stream directly? There is no need to save the intermediate result as list.
```
 permissionCache.keys()
            .stream()
            .filter(hdl-> hdl.getId()!=null)
            .filter(hdl -> idSet.contains(hdl.getId()))
            .forEach(permissionCache::remove);
```

mycore-base/src/main/java/org/mycore/access/MCRAccessCacheEventHandler.java
```diff
@@ -0,0 +1,87 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.mycore.common.events.MCREvent;
+import org.mycore.common.events.MCREventHandlerBase;
+import org.mycore.datamodel.common.MCRLinkTableManager;
+import org.mycore.datamodel.metadata.MCRDerivate;
+import org.mycore.datamodel.metadata.MCRObject;
+
+/**
+ * Updates the access cache when a object is updated or deleted
```
rsteph-de: "Clears the access cache when an object is updated or deleted.
It invalidates all cache entries for the object, it's derivates and all it's decendants
in all active MCRSessions."

mycore-base/src/main/java/org/mycore/access/MCRAccessCacheManager.java
```diff
@@ -96,4 +105,28 @@ public void removePermission(String id, String permission) {
         permissionCache.remove(handle);
     }
 
+    public void removePermission(String... ids) {
+        MCRCache<MCRPermissionHandle, Boolean> permissionCache = accessCache.get();
+        removePermissionFromCache(permissionCache, ids);
+    }
+
+    private void removePermissionFromCache(MCRCache<MCRPermissionHandle, Boolean> permissionCache, String... ids) {
```
rsteph-de: use `Set<String>` instead of `String...` for ids, 
this would save you from having to create the set every time the method is called again.

rsteph-de (changes requested):
some javadoc and performance issues

mycore-base/src/main/java/org/mycore/access/MCRAccessCacheManager.java
sebhofmann: This would cause a java.util.ConcurrentModificationException, because we are iterating over the same collection where the items are removed.

rsteph-de: O.K., your're right

rsteph-de (approved):
looks good for me

PR #1433, opened 2022-01-20
MCR-2562 MyCoRe Storage Layout Derivate Implementation in the OCFLXMLMetadataManager

Hammer1279:
Make the OCFLXMLMetadataManager use the ID Prefix on Derivates `mcrderivate` instead of always using `mcrobject`.

~Also adds new throws declaration for `NotFoundException`~ *moved to #1434*

# This brings BREAKING changes to old OCFL Repositories
### That is because derivates are now stored as `mcrderivate:project_type_id` instead of `mcrobject:project_type_id`

[Link to jira](https://mycore.atlassian.net/browse/MCR-2562).

### DIFF
--- a/mycore-ocfl/src/main/java/org/mycore/ocfl/MCROCFLXMLMetadataManager.java
+++ b/mycore-ocfl/src/main/java/org/mycore/ocfl/MCROCFLXMLMetadataManager.java
@@ -25,6 +25,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.Date;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.TimeUnit;
@@ -77,6 +78,8 @@ public class MCROCFLXMLMetadataManager implements MCRXMLMetadataManagerAdapter {
         Map.entry(MESSAGE_DELETED, MCROCFLMetadataVersion.DELETED)));
 
     public static final String MCR_OBJECT_ID_PREFIX = "mcrobject:";
+
+    public static final String MCR_DERIVATE_ID_PREFIX = "mcrderivate:";
 
     private String repositoryKey = "Default";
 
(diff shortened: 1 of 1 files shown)


mycore-ocfl/src/main/java/org/mycore/ocfl/MCROCFLXMLMetadataManager.java
```diff
@@ -264,10 +268,19 @@ private VersionInfo buildVersionInfo(String message, Date versionDate, String us
         }).collect(Collectors.toList());
     }
 
+    private boolean isMetadata(String id) {
+        return id.startsWith(MCR_OBJECT_ID_PREFIX) || id.startsWith(MCR_DERIVATE_ID_PREFIX);
+    }
+
+    private String substrPrefix(String id) {
```
sebhofmann: better rename this method to removePrefix.

mycore-ocfl/src/main/java/org/mycore/ocfl/MCROCFLXMLMetadataManager.java
```diff
@@ -175,7 +178,8 @@ private String getObjName(MCRObjectID mcrid) {
     }
 
     private String getObjName(String mcrid) {
-        return MCR_OBJECT_ID_PREFIX + mcrid;
+        String objectType = MCRObjectID.getIDParts(mcrid.trim())[1].toLowerCase(Locale.ROOT).intern();
```
sebhofmann: Could use MCRObjectID.getInstance(mcrid).getTypeId() here. It uses caching and the ID is probably already cached, so no string operation needed.

mycore-ocfl/src/main/java/org/mycore/ocfl/MCROCFLXMLMetadataManager.java
```diff
@@ -248,7 +252,7 @@ private VersionInfo buildVersionInfo(String message, Date versionDate, String us
     }
 
     @Override
-    public List<MCROCFLMetadataVersion> listRevisions(MCRObjectID id) throws IOException {
+    public List<MCROCFLMetadataVersion> listRevisions(MCRObjectID id) throws IOException, NotFoundException {
```
sebhofmann: For consistency it would might be better to catch the NotFoundException like in the other methods:
        } catch (NotFoundException e) {
            throw new MCRUsageException("Object '" + objName + "' could not be found", e);
        }

sebhofmann (changes requested):

sebhofmann (approved):

PR #1454, opened 2022-02-07
MCR-2575 MCRCronjobManager blocks application shutdown

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2575).


kkrebs (approved):

yagee-de (changes requested):
you probably should not call shutdownNow() directly as it also kills any running or waiting task.
make use of prepareClose() and call shutdown() there together with awaitTermination() and a reasonable time (e.g. 60 seconds) in close() you can than keep calling shutdownNow() and again awaitTermination() with a reasonalbe time to wait for the shutdown.

yagee-de (approved):

PR #1458, opened 2022-02-14
MCR-2580 Add OCFL Specs mycore-storage-layout

Hammer1279:
Add the mycore-storage-layout to the ocfl-specs

[Link to jira](https://mycore.atlassian.net/browse/MCR-2580).


rsteph-de (approved):
good as inital draft / may be modified after our DevCon

PR #1467, opened 2022-02-17
MCR-1207 Add MCRCachingResolver

fluetze:
[Link to jira](https://mycore.atlassian.net/browse/MCR-1207).


yagee-de:
I'm okay with that. Somebody else should review our code.

sebhofmann:
I like the extra Class, it makes it possible to remove the URI Resolver or replace it with an improved variant in the application and makes the code more modular.

yagee-de:
> I like the extra Class, it makes it possible to remove the URI Resolver or replace it with an improved variant in the application and makes the code more modular.

All other URIResolver are kept internal and this URIResolver is nothing special. I don't see the point of removing URIResolver either. An improved variant should go to MyCoRe.
When we modularize our code we should definitively try to separate all URIResolver classes while keeping them private. For now it's better to handle all URIResolver the same way if they are directly handled by MCRURIResolver.

yagee-de (approved):

sebhofmann (approved):

PR #1470, opened 2022-02-21
MCR-2309 minor presentation bugs in csl styles

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2309).


mycore-mods/src/main/java/org/mycore/mods/csl/MCRModsItemDataProvider.java
```diff
@@ -339,47 +353,68 @@ protected void processPublicationData(CSLItemDataBuilder idb) {
...
-                case "pmcid":
-                    idb.PMCID(identifier);
-                    break;
-            }
+            applyIdentifier(idb, identifierElement, false);
         });
 
     }
 
+    private void applyIdentifier(CSLItemDataBuilder idb, Element identifierElement, boolean parent) {
```
kkrebs, line 374: We should also handle the case "URN" and "URI"

kkrebs (changes requested):

kkrebs (approved):

PR #1490, opened 2022-03-25
MCR-2607 MetsMods2IIIFConverter has issues with filenames and empty m…

eckstueck:
…ets:divs

fixed MetsMods2IIIFConverter can not handle Umlaute and Spaces in filenames

fixed MetsMods2IIIFConverter can not handle empty mets:divs that are not linked to images

[Link to jira](https://mycore.atlassian.net/browse/MCR-2607).

### DIFF
--- a/mycore-mets/src/main/java/org/mycore/mets/iiif/MCRMetsMods2IIIFConverter.java
+++ b/mycore-mets/src/main/java/org/mycore/mets/iiif/MCRMetsMods2IIIFConverter.java
@@ -18,7 +18,6 @@
 
 package org.mycore.mets.iiif;
 
-import java.io.UnsupportedEncodingException;
 import java.net.URLDecoder;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
--- a/mycore-mets/src/main/java/org/mycore/mets/iiif/MCRMetsMods2IIIFConverter.java
+++ b/mycore-mets/src/main/java/org/mycore/mets/iiif/MCRMetsMods2IIIFConverter.java
@@ -163,7 +163,7 @@ public class MCRMetsMods2IIIFConverter {
             String identifier = this.physicalIdentifierMap.get(physicalSubDiv);
             try {
                 MCRIIIFImageInformation information = imageImpl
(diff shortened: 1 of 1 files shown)


mycore-mets/src/main/java/org/mycore/mets/iiif/MCRMetsMods2IIIFConverter.java
```diff
@@ -159,7 +162,8 @@ public MCRIIIFManifest convert() {
 
             String identifier = this.physicalIdentifierMap.get(physicalSubDiv);
             try {
-                MCRIIIFImageInformation information = imageImpl.getInformation(identifier);
+                MCRIIIFImageInformation information = imageImpl
+                    .getInformation(URLDecoder.decode(identifier, StandardCharsets.UTF_8.toString()));
```
sebhofmann: Do not use .toString() here, to prevent  UnsupportedEncodingException to be thrown.

rsteph-de: UTF_8 is garanteed by all Java implementations
(https://docs.oracle.com/javase/7/docs/api/java/nio/charset/StandardCharsets.html)
But the code looks still better without .toString()

sebhofmann: > UTF_8 is garanteed by all Java implementations

Yes but you have to use the [right method](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/net/URLDecoder.html#decode(java.lang.String,java.nio.charset.Charset) ), to prevent the Exception.

sebhofmann (approved):

mycore-mets/src/main/java/org/mycore/mets/iiif/MCRMetsMods2IIIFConverter.java
eckstueck: Thanks for the hint, fixed now.

yagee-de (approved):

PR #1495, opened 2022-03-30
Fix mycore-parent-pom version

sebhofmann:


yagee-de (commented):
Fixed dependencies in https://github.com/MyCoRe-Org/mycore/commit/73f6a513344cb1bfa5dcb3662e47f9c1c40dc5d1

PR #1546, opened 2022-05-10
MCR-2604 OCFL Classification Storage

Hammer1279:
This adds a Classification Storage in a OCFL Repository, while the Database handles access to the files, the OCFL Store works like a backup and stores changes versioned. 

In additon, to work, this also brings an DAO Implementation with added Events and a ClassificationManager Interface.

[Link to jira](https://mycore.atlassian.net/browse/MCR-2604).


Hammer1279:
Docs can be found here: https://github.com/MyCoRe-Org/mycore-website/blob/ocfl-class/mycore.org/content/de/documentation/basics/basics_mcr_ocfl.md

mycore-ocfl/src/main/java/org/mycore/ocfl/MCROCFLPersistenceTransaction.java
```diff
@@ -0,0 +1,165 @@
...
+        managerOpt.get().rollbackSession(currentSession);
+        rollbackOnly = false;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public boolean getRollbackOnly() {
+        if (verboseLogging) {
```
yagee-de: LOGGER.debug already checks if Specific logger is on debug level. No need to specify any property or adding custom checks

mycore-base/src/main/java/org/mycore/datamodel/classifications2/impl/MCREventedCategoryDAOImpl.java
```diff
@@ -0,0 +1,127 @@
...
+        manager.handleEvent(evt);
+        queueForCommit(evt);
+        return rv;
+    }
+
+    @SuppressWarnings("unchecked")
+    protected void queueForCommit(MCREvent evt) {
+        String classQueue = "classQueue";
+        MCRSession currentSession = MCRSessionMgr.getCurrentSession();
+        ((ArrayList<MCREvent>) currentSession.get(classQueue)).add(evt);
```
yagee-de: String "classQueue" is used to often in Code. This should be centralized somewhere.
Also a BIG no go is to use session as a storage for transactional data without any transactional reference. There could be two different transactions running concurrently and one is rolled back. In this case all the data of two transactions are lost.

mycore-ocfl/src/main/java/org/mycore/ocfl/MCROCFLXMLClassificationManager.java
```diff
@@ -0,0 +1,332 @@
...
+        fileUpdate(newParent.getId(), newParent, newParentXml, eventData);
+        fileUpdate(oldParent.getId(), oldParent, oldParentXml, eventData);
+    }
+
+    public void commitChanges(MCREvent evt) {
+        Map<String, Object> data = MCROCFLEventHandler.getEventData(evt);
+        MCRCategoryID mcrid = (MCRCategoryID) data.get("mid");
+        MCRCategory mcrCg = (MCRCategory) data.get("ctg");
+        MCRContent rtXml = (MCRContent) data.get("rtx"); // class
+        MCREvent event = (MCREvent) evt.get("event");
```
yagee-de: Do not use magic Strings here. Make them "public static final" somewhere and document them. What is that "event" within an event anyway?

mycore-ocfl/src/main/resources/components/ocfl/config/mycore.properties
```diff
@@ -16,14 +16,73 @@
...
 #
 
-MCR.CLI.Classes.Internal=%MCR.CLI.Classes.Internal%,org.mycore.ocfl.commands.MCROCFLCommands
+#      __  ___        ______        ____           ____   ______ ______ __ 
+#     /  |/  /__  __ / ____/____   / __ \ ___     / __ \ / ____// ____// / 
+#    / /|_/ // / / // /    / __ \ / /_/ // _ \   / / / // /    / /_   / /  
+#   / /  / // /_/ // /___ / /_/ // _, _//  __/  / /_/ // /___ / __/  / /___
+#  /_/  /_/ \__, / \____/ \____//_/ |_| \___/   \____/ \____//_/    /_____/
+#          /____/                                                          
+
```
yagee-de: please: no!

yagee-de (changes requested):

yagee-de:
Properties of a transaction are:

- Atomicity
- Consistency
- Isolation
- Durability

Atomicity is hard to achieve on a filesystem storage, we could rollback to a previous version. So not that big of a deal.
Consistency seems not to be harmed by this PR.
Isolation is currently weakened as any change is visible to any other user in a mutable directory and transaction commits data from other transactions within the same session. Also different transaction could write to the same resource.
Durability is due to ocfl.

yagee-de:
You could keep your data in a transaction object and have that transaction being committed later. If you NEED to use a mutable HEAD of an OCFL object, make sure to acquire a lock to operate on that. That lock should belong to the transaction and if the transaction is committed than "commitStagedChanges" on that object. If a rollback occurs, free the lock after you "purgeStagedChanges".

yagee-de:
Thanks for your documentation and your work. I really appreciate your push towards a OCFL storage option.

mycore-ocfl/src/main/java/org/mycore/ocfl/commands/MCROCFLDevCommands.java
```diff
@@ -0,0 +1,124 @@
...
+        String repositoryKey = MCRConfiguration2.getStringOrThrow("MCR.Classification.Manager.Repository");
+        String ocflRoot = MCRConfiguration2
+            .getStringOrThrow("MCR.OCFL.Repository." + repositoryKey + ".RepositoryRoot");
+        Path classDir = Path.of(ocflRoot, "mcrclass");
+        try (Stream<Path> walker = Files.walk(classDir)) {
+            walker.sorted(Comparator.reverseOrder()).map(Path::toFile).forEach(File::delete);
+        } catch (Exception e) {
+            // throw new IOException(e);
+            // expected Error incase the directory doesn't exist yet
+        }
```
yagee-de: Does not use OCFL-API and requires a specific non-default storage layout

mycore-ocfl/src/main/java/org/mycore/ocfl/MCROCFLXMLClassificationManager.java
```diff
@@ -0,0 +1,213 @@
...
+     * @param mcrid The ID to the Classification
+     * @return The Path to the File.
+     */
+    protected String buildFilePath(MCRCategoryID mcrid) {
+        if (mcrid.isRootID()) {
+            return rootFolder + mcrid.toString() + ".xml";
+        } else {
+            return rootFolder + mcrid.getRootID() + '/' + mcrid.toString() + ".xml";
+        }
+    }
```
yagee-de: Where are all these files come from? Where is this specified? How to we ensure that if a classification is deleted all categories are deleted, too?

mycore-ocfl/src/main/java/org/mycore/ocfl/MCROCFLXMLClassificationManager.java
```diff
@@ -0,0 +1,213 @@
...
+        } catch (IOException e) {
+            throw new MCRPersistenceException("Failed to update object '" + objName + "'", e);
+        } catch (IllegalArgumentException e) {
+            // Issue seems to be with the OCFL Library trying to find a file it deleted itself
+            // LOGGER.error("Something has gone Wrong!", e);
+        }
+
+    }
+
+    public void fileDelete(MCRCategoryID mcrid) {
```
yagee-de: method naming: We have "create", "update" and "fileDelete". Should be renamed to "delete"

yagee-de (changes requested):

yagee-de:
We should have a discussion about messages beeing parsed to determine what action caused a revision. This does not look good to me.

mycore-ocfl/src/main/java/org/mycore/ocfl/MCROCFLPersistenceTransaction.java
```diff
@@ -87,12 +82,12 @@ public void commit() {
             throw new IllegalStateException("TRANSACTION NOT ACTIVE OR MARKED FOR ROLLBACK");
         }
         final Map<MCRCategoryID, MCRCategory> mapOfChanges = CATEGORY_WORKSPACE.get();
-        final MCROCFLXMLClassificationManager ocflClassficationManager = managerOpt.get();
+        final MCROCFLXMLClassificationManager ocflClassficationManager = manager;
```
Hammer1279: Why not just change all occurances of `ocflClassificationManager` to `manager`?

yagee-de (approved):

PR #1578, opened 2022-06-03
MCR-2305 Enrichment Resolver does not query newly found identifiers

chemicus:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2305).


fluetze:
This is still draft. I refactored some functionality and introduced configuration properties:
default: MCR.MODS.EnrichmentResolver.DefaultStopOnFirstResult=true
per DS: MCR.MODS.EnrichmentResolver.DataSource.[ID].StopOnFirstResult=true|false

kkrebs (approved):

PR #1579, opened 2022-06-03
MCR-2663 Repeatedly query all known sources for metadata during enric…

chemicus:
…hment

[Link to jira](https://mycore.atlassian.net/browse/MCR-2663).


kkrebs (changes requested):
This should be configurable per source. In some cases you don't want to ask the same source again (e.g. WebOfScience or Scopus where you have to pay per request and in most cases no new information is available).

chemicus:
I've already answered this raised issue 3 weeks ago! Please see my comment in [Jira](https://mycore.atlassian.net/browse/MCR-2663?focusedCommentId=48835))

fluetze:
This will be fixed together with MCR-2305

PR #1583, opened 2022-06-03
MCR-2667 Add MCRAffiliationMerger

chemicus:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2667).


chemicus:
Will make merger more general

PR #1590, opened 2022-06-16
MCR-2673 Add command to export object using a content transformer

toKrause:
[depends on MCR-2672](https://mycore.atlassian.net/browse/MCR-2672)!

[Link to jira](https://mycore.atlassian.net/browse/MCR-2673).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/frontend/cli/MCRObjectCommands.java
+++ b/mycore-base/src/main/java/org/mycore/frontend/cli/MCRObjectCommands.java
@@ -21,6 +21,7 @@ package org.mycore.frontend.cli;
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
+import java.io.OutputStream;
 import java.text.MessageFormat;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
@@ -34,6 +35,7 @@ import java.util.Map.Entry;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.BiConsumer;
 import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
@@ -51,6 +53,7 @@ import javax.xml.transform.stream.StreamSource;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.apache.xml.utils.WrappedRuntimeException;
 import org.jdom2.Document;
 import org.jdom2.JDOMException;
 import org.jdom2.filter.Filters;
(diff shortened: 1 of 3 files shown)


toKrause:
Rebased and types fixed.

mycore-base/src/main/java/org/mycore/frontend/cli/MCRObjectCommands.java
```diff
@@ -702,36 +771,34 @@ private static Transformer getTransformer(String style) {
...
      *            the MCRObjectID
      * @return true if the store was okay (see description), else return false
-     * @throws TransformerException
      * @throws IOException
      * @throws MCRException
-     * @throws SAXParseException
      */
-    private static boolean exportMCRObject(File dir, Transformer trans, String nid)
-        throws TransformerException, IOException, MCRException, SAXParseException {
+    private static boolean exportMCRObject(File dir, BiConsumer<MCRContent, OutputStream> trans, String nid) throws
```
sebhofmann: All implementations throw some kind of error, maybe it would be better to write a own specific interface for that, or we could implement or reuse a ThrowingBiConsumer. We could use https://github.com/pivovarit/throwing-function or could implement a own Interface like https://github.com/mycore-org/mycore/blob/6795c535d118e7b541a241497b7514404449dd2c/mycore-base/src/main/java/org/mycore/common/function/MCRThrowFunction.java .

sebhofmann (changes requested):

mycore-base/src/main/java/org/mycore/frontend/cli/MCRObjectCommands.java
toKrause: I've considered that myself and even implemented a `MCRThrowBiConsumer<T,U,E>` that used the same design principles as the other `MCRThrow`-classes.

The problem is line 628 where the stylesheet implementation throws multiple distinct exceptions:
```
} catch (TransformerException | SAXParseException e) {
```

The two exception types have no common parent other than `Exception`, so we would either have to settle for
`MCRThrowBiConsumer<MCRContent, OutputStream, Exception>`, by which we would gain nothing, or we would have to introduce `MCRBiThrowBiConsumer<T,U,E1,E2>`, which is clearly ridicules.

Therefore, I decided to go back to the a plain old Java `BiConsumer` and introduce the anonymous wrapper classes that handle the exception mapping.

toKrause: But I'm open for other suggestions,

rsteph-de: I would vote against using the pivovarit functions to avoid another dependency to another project and implement the interfaces ourselves if necessary.

toKrause: I agree. we should stick to `MCRThrow`-... instead of including another dependency.

Additionaly, [`ThrowingBiConsumer`](https://github.com/pivovarit/throwing-function/blob/master/src/main/java/com/pivovarit/function/ThrowingBiConsumer.java) also supports only one exception type.

sebhofmann (approved):

PR #1591, opened 2022-06-16
MCR-2674 Allow for diffeering creator permissions in two staged publication processes

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2674).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/access/strategies/MCRCreatorRuleStrategy.java
+++ b/mycore-base/src/main/java/org/mycore/access/strategies/MCRCreatorRuleStrategy.java
@@ -36,14 +36,19 @@ import org.mycore.datamodel.metadata.MCRObjectID;
 
 /**
  *
- * First a check is done if user is in role "submitter", the given object is in
- * status "submitted" and current user is creator. If not
- * it will be tried to check the permission against the rule ID
+ * First, a check is done if the user is in the submitter role, if the user is the creator of the object and if the
+ * creator ist permitted to perform the requested action in the object's current state (two groups of states, submitted,
+ * and review, are checked). If not it will be tried to check the permission against the rule ID
  * <code>default_&lt;ObjectType&gt;</code> if it exists. If not the last
  * fallback is done against <code>default</code>.
  *
- * Specify classification and category for status "submitted":
- * MCR.Access.Strategy.SubmittedCategory=state:submitted
+ * Specify classification and category for submitted and review states:
+ * MCR.Access.Strategy.SubmittedCategories=state:submitted
+ * MCR.Access.Strategy.ReviewCategories=state:review
+ *
+ * Specify permissions for submitted and review states:
+ * MCR.Access.Strategy.CreatorSubmittedPermissions=writedb,deletedb
+ * MCR.Access.Strategy.CreatorReviewPermissions=read
  *
  * You can also specify a comma separated list of categories like: <code>state:submitted,state:new</code>
  *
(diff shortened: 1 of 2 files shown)


mycore-base/src/main/java/org/mycore/access/strategies/MCRCreatorRuleStrategy.java
```diff
@@ -86,23 +105,33 @@ public boolean checkPermission(String id, String permission) {
...
-        String[] submittedCategoriesSplitted = SUBMITTED_CATEGORY.split(",");
-        for (String submittedCategoryID : submittedCategoriesSplitted) {
-            String categoryId = submittedCategoryID.trim();
-            MCRCategoryID submittedCategory = MCRCategoryID.fromString(categoryId);
-            if (LINK_SERVICE.isInCategory(reference, submittedCategory)) {
-                isSubmitted = true;
+        for (String categoryId : categoryIds) {
+            MCRCategoryID category = MCRCategoryID.fromString(categoryId);
+            if (LINK_SERVICE.isInCategory(reference, category)) {
+                result = true;
```
sebhofmann: This can be replaced with return true, to skip the check of other categories.

mycore-base/src/main/java/org/mycore/access/strategies/MCRCreatorRuleStrategy.java
```diff
@@ -118,24 +147,24 @@ public boolean hasRuleMapping(String id, String permission) {
...
-                    && objectStatusIsSubmitted(mcrObjectId)) {
-                    if (isCurrentUserCreator(mcrObjectId, currentUser)) {
-                        return true;
-                    }
+        MCRObjectID mcrObjectId = null;
+        try {
+            mcrObjectId = MCRObjectID.getInstance(id);
+            MCRUserInformation currentUser = MCRSessionMgr.getCurrentSession().getUserInformation();
+            if (userHasCreatorRole(currentUser) && userIsCreatorOf(currentUser, mcrObjectId)) {
+                if (objectStatusIsSubmitted(mcrObjectId) && CREATOR_SUBMITTED_PERMISSIONS.contains(permission)) {
```
sebhofmann, line 152: Should be changed to:
```
if (objectStatusIsSubmitted(mcrObjectId)){
return CREATOR_SUBMITTED_PERMISSIONS.contains(permission);
} else if(objectStatusIsReview(mcrObjectId)) {
return CREATOR_REVIEW_PERMISSIONS.contains(permission);
}
```
to prevent double status check

mycore-base/src/main/java/org/mycore/access/strategies/MCRCreatorRuleStrategy.java
```diff
@@ -118,24 +147,24 @@ public boolean hasRuleMapping(String id, String permission) {
...
-            } catch (RuntimeException e) {
-                if (mcrObjectId == null) {
-                    LOGGER.debug("id is not a valid object ID", e);
-                } else {
-                    LOGGER.warn("Eror while checking permission.", e);
+                if (objectStatusIsReview(mcrObjectId) && CREATOR_REVIEW_PERMISSIONS.contains(permission)) {
+                    return true;
                 }
             }
+        } catch (RuntimeException e) {
```
sebhofmann, line 159: The entire catch block is bad, but it was there before.

- Catching RuntimeException instead of a specific exception
- the mcrObjectId == null can be checked before

sebhofmann (changes requested):

mycore-base/src/main/java/org/mycore/access/strategies/MCRCreatorRuleStrategy.java
toKrause: The code, as proposed, is a more readable version of

```
 if (objectStatusIsSubmitted(mcrObjectId) && CREATOR_SUBMITTED_PERMISSIONS.contains(permission) || (objectStatusIsReview(mcrObjectId) && CREATOR_REVIEW_PERMISSIONS.contains(permission)) {
   return true;
 }
```

I intentionally avoided the use of `else` here, because the two branches are equally significant. The proposed change obfuscates this intent and introduces a subtle difference.

Assume the following configuration:

```
MCR.Access.Strategy.SubmittedCategories         = s1,s2
MCR.Access.Strategy.ReviewCategories            = s2,s3
MCR.Access.Strategy.CreatorSubmittedPermissions = p1,p2
MCR.Access.Strategy.CreatorReviewPermissions    = p1,p3
```

Now, if the object is in state `s2` and the requested permission is `p3`, I' expect a result of `true`, but with the proposed change, the actual result would be `false`.

I've no idea if such a configuration is in any way realistic, but on the other hand, I see no reason to disallow such a configuration.

Should someone have a need for such a configuration, it would be very hard for them to detect that is is not supported. Best case would than probably a lengthy debugging session; worst case would be some unexpected and unnoticed permission-problem. I think we shouldn't unnecessarily introduce such subtleties.

mycore-base/src/main/java/org/mycore/access/strategies/MCRCreatorRuleStrategy.java
toKrause: There is no way for `mcrObjectId` to be `null` anyway. 

My reasoning: `MCRObjectID#getInstance(String)` will throw an exception if the provided string is `null`, Otherwise it will call `MCRObjectIDPool.getMCRObjectID` which will return a cached `MCRObjectID` or try to construct one. Construction will fail, if the provided string is not a syntactically valid object ID.

I'll remove the superfluous `if`-`then`-`else` but keep the `try`-`catch`. This PR should only extend the configuration options for the creator rule strategy, it shouldn't change the externally observed behavior  of this class. A follow-up PR might be appropriate.

toKrause (commented):
Accidentally started a review due to GitHubs call-to-actiony-button-design

sebhofmann (approved):

PR #1597, opened 2022-06-22
MCR-2680 fix exception when checking RestAPI access

rsteph-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2680).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/datamodel/classifications2/MCRCategoryID.java
+++ b/mycore-base/src/main/java/org/mycore/datamodel/classifications2/MCRCategoryID.java
@@ -183,19 +183,17 @@ public class MCRCategoryID implements Serializable {
         }
         this.id = id;
     }
-    
-    
+
     /**
      * @param id
      *              the ID to check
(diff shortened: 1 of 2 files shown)


sebhofmann (approved):

mycore-base/src/main/java/org/mycore/datamodel/classifications2/MCRCategoryID.java
```diff
@@ -179,6 +179,10 @@ private void setID(String id) {
         }
         this.id = id;
     }
+    
+    public static boolean isValid(String id) {
+        return id != null && id.length() > 0 && id.length() <= CATEG_ID_LENGTH && VALID_ID.matcher(id).matches();
+    }
```
yagee-de: Should just call fromString() and catch exceptions. This check too isValid is too simple.
I expect that the call to fromString() is without exception if and only if isValid() returns true

yagee-de: Javadocs should be provided, too

PR #1601, opened 2022-06-24
MCR-2681 RestAPIv2 create parent directory on file upload

rsteph-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2681).

### DIFF
--- a/mycore-restapi/src/main/java/org/mycore/restapi/v2/MCRRestDerivateContents.java
+++ b/mycore-restapi/src/main/java/org/mycore/restapi/v2/MCRRestDerivateContents.java
@@ -375,6 +375,11 @@ public class MCRRestDerivateContents {
         if (mcrPath.getNameCount() > 1) {
             MCRPath parentDirectory = mcrPath.getParent();
             try {
+                if (!Files.exists(parentDirectory)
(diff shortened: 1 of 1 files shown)


mycore-restapi/src/main/java/org/mycore/restapi/v2/MCRRestDerivateContents.java
```diff
@@ -375,6 +375,11 @@ public Response createFileOrDirectory(InputStream contents) {
         if (mcrPath.getNameCount() > 1) {
             MCRPath parentDirectory = mcrPath.getParent();
             try {
+                if (!Files.exists(parentDirectory)
+                    && parentDirectory.toAbsolutePath().startsWith(
+                        MCRPath.getPath(derid.toString(), "/").toAbsolutePath())) {
+                    Files.createDirectories(parentDirectory);
+                }
```
yagee-de: Should be:
```java
Files.createDirectories(parentDirectory);
```
It does the check for existence already and the other checks have no reason I think. Check the call to createDirectories() for a FileAlreadyExistsException if the parent directory is a file to return the right error message later in the code

yagee-de (changes requested):

yagee-de (approved):

PR #1603, opened 2022-06-28
MCR-2684 RestAPIv2: JSON serialization for objects and derivates

rsteph-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2684).


yagee-de (commented):
JSON output should be done via Transformer and is already handled via `org.mycore.restapi.converter.MCRContentAbstractWriter`. But we could include a transformer for MCRDerivate. Feel free to submit on with this PR

yagee-de:
We could also include a Transformer for handling your case, that `MCRObject.createJSON()` is enough.

rsteph-de:
works now with MCRContentWriter and new MCRContentTransformers to create JSON from MCRObjects and MCRDerivates

mycore-base/src/main/java/org/mycore/common/content/transformer/MCRBase2JSONTransformer.java
```diff
@@ -0,0 +1,92 @@
...
+        result.setEncoding(getEncoding());
+        result.setUsingSession(source.isUsingSession());
+        return result;
+    }
+
+    private static JsonObject toJSON(MCRContent source) throws IOException {
+        try {
+            String idAttr = source.asXML().getRootElement().getAttributeValue("ID");
+            MCRObjectID mcrId = MCRObjectID.getInstance(idAttr);
+            MCRBase mcrBase = MCRMetadataManager.retrieve(mcrId);
```
yagee-de: You should use the content provided not read it again.

yagee-de (changes requested):

mycore-base/src/main/java/org/mycore/common/content/transformer/MCRBase2JSONTransformer.java
rsteph-de: solved with new ContentTransformers vor MCRObject and MCRDerivate

yagee-de (approved):
:+1: looks very good

PR #1607, opened 2022-06-29
MCR-2681 RESTv2 - create directory by trailing slash and HTTP-Header X-MCR-IsDirectory

rsteph-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2681).

### DIFF
--- a/mycore-restapi/src/main/java/org/mycore/restapi/v2/MCRRestDerivateContents.java
+++ b/mycore-restapi/src/main/java/org/mycore/restapi/v2/MCRRestDerivateContents.java
@@ -102,7 +102,7 @@ import io.swagger.v3.oas.annotations.responses.ApiResponse;
 
 @Path("/objects/{" + PARAM_MCRID + "}/derivates/{" + PARAM_DERID + "}/contents{" + PARAM_DER_PATH + ":(/[^/]+)*}")
 public class MCRRestDerivateContents {
-    private static final String HTTP_HEADER__IS_DIRECTORY = "X-MCR-IsDirectory";
+    private static final String HTTP_HEADER_IS_DIRECTORY = "X-MCR-IsDirectory";
 
     private static final int BUFFER_SIZE = 8192;
 
@@ -370,7 +370,7 @@ public class MCRRestDerivateContents {
     @Operation(summary = "Creates directory or file. Parent directories will be created if they do not exist.",
         parameters = {
             @Parameter(in = ParameterIn.HEADER,
-                name = HTTP_HEADER__IS_DIRECTORY,
+                name = HTTP_HEADER_IS_DIRECTORY,
                 description = "set to 'true' if a new directory should be created",
                 required = false,
                 schema = @Schema(type = "boolean")) },
@@ -485,7 +485,7 @@ public class MCRRestDerivateContents {
     private boolean isFile() {
         //as per https://tools.ietf.org/html/rfc7230#section-3.3
         MultivaluedMap<String, String> headers = request.getHeaders();
-        return ((!"true".equalsIgnoreCase(headers.getFirst(HTTP_HEADER__IS_DIRECTORY))))
+        return ((!"true".equalsIgnoreCase(headers.getFirst(HTTP_HEADER_IS_DIRECTORY))))
             && (headers.containsKey(HttpHeaders.CONTENT_LENGTH) || headers.containsKey("Transfer-Encoding"));
     }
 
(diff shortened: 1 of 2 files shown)


yagee-de (approved):

mycore-restapi/src/main/java/org/mycore/restapi/v2/MCRRestDerivateContents.java
```diff
@@ -95,11 +95,14 @@
...
 import io.swagger.v3.oas.annotations.Operation;
 import io.swagger.v3.oas.annotations.Parameter;
+import io.swagger.v3.oas.annotations.enums.ParameterIn;
 import io.swagger.v3.oas.annotations.headers.Header;
+import io.swagger.v3.oas.annotations.media.Schema;
 import io.swagger.v3.oas.annotations.responses.ApiResponse;
 
 @Path("/objects/{" + PARAM_MCRID + "}/derivates/{" + PARAM_DERID + "}/contents{" + PARAM_DER_PATH + ":(/[^/]+)*}")
 public class MCRRestDerivateContents {
+    private static final String HTTP_HEADER_IS_DIRECTORY = "X-MCR-IsDirectory";
```
toKrause, line 105: I still think that introducing custom `X`-headers should only be a last resort. And I think the discussion on the mailing list also shows that a solution without such a header would be preferred (option 2).

Disambiguating file and folder request by the presence of a trailing slash (for folders) should be preferred.

The Jakarta-community [suggested](https://github.com/jakartaee/rest/issues/468) injecting `UriInfo` as a workaround (injecting `@Context UriInfo uriInfo` and checking `uriInfo.getRequestUri().getPath().endsWith("/")`).

I tried such an implementation as a proof of concept and it worked fine:

```
public Response createFileOrDirectory(@Context UriInfo uriInfo,InputStream contents) {
    System.out.println("PUT uriInfo.getPath()                 " + uriInfo.getPath());
    System.out.println("PUT uriInfo.getRequestUri().getPath() " + uriInfo.getRequestUri().getPath());
        ...
```

`PUT /api/v2/objects/macau_mods_00000001/derivates/macau_derivate_00000001/contents/foobar` resulted in 
```
PUT uriInfo.getPath()                 objects/macau_mods_00000001/derivates/macau_derivate_00000001/contents/foobar
PUT uriInfo.getRequestUri().getPath() /api/v2/objects/macau_mods_00000001/derivates/macau_derivate_00000001/contents/foobar
```

`PUT /api/v2/objects/macau_mods_00000001/derivates/macau_derivate_00000001/contents/foobar/` resulted in 
```
PUT uriInfo.getPath()                 objects/macau_mods_00000001/derivates/macau_derivate_00000001/contents/foobar/
PUT uriInfo.getRequestUri().getPath() /api/v2/objects/macau_mods_00000001/derivates/macau_derivate_00000001/contents/foobar/
```

The only other thing that hinders this approach is that `MCRNormalizeMCRObjectIDsFilter#filter(ContainerRequestContext)` currently redirect all HTTP methods  from `/api/v2/objects/macau_mods_00000001/derivates/macau_derivate_00000001/contents/foobar/` to `/api/v2/objects/macau_mods_00000001/derivates/macau_derivate_00000001/contents/foobar`; removing the trailing slash.

toKrause: Not sure why an `...ObjectIDsFilter...`  messes with derivate content file paths though, ...

rsteph-de: fixing MCRNormalizeMCRObjectIDsFilter to keep trailing "/" does the "trick"

toKrause (commented):
I'm confused now ...

As far as I can judge, this PR already did what it was supposed to do (allowing to create empty folders) by means of a custom header.

I mentioned that I still think that the whole approach should be avoided. Using a trailing slash as an indicator would be better. 

The changes to `MCRNormalizeMCRObjectIDsFilter` would only be necessary to allow the alternate (trailing slash) approach.

But the original (header) approach is still use, negating the need to modify `MCRNormalizeMCRObjectIDsFilter`.

What am I missing?

---

I'm not going to "request changes" since the modifications seem fine, **IF** the header approach is the approach we're going with.

Or are we just allowing both approaches now? That would also be fine, but the PR or Jira ticket  should reflect that.

rsteph-de:
The last modifications should make it possible to detected directory names by trailing slash. 
The implementation of `MCRNormalizeMCRObjectIDsFilter` was buggy (by removing trailing slashes).
So the prefered way to create a directory via Rest should be the 'trailing slash'. Right now I do not want to give up the optional header param until we have checked other cases (HTTPClients, APIs, proxies, ...) which could also modify the URL.

After the PR-merge I am going to update the documentation regarding these new features.

toKrause:
That seems reasonable.

Apparently I've missed the ... in image. My bad.

PR #1608, opened 2022-06-30
MCR-2688 RestAPIv2 - new PATCH methods to update derivate metadata

rsteph-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2688).

Before you can use / test this new feature , you need to enable it via property:
`MCR.RestApi.Draft.PartialMetadataUpdates=true`


golsch:
How about parsing the operation JSON into a derivate with an object mapper?

rsteph-de:
> How about parsing the operation JSON into a derivate with an object mapper?

We are looking for a simple solution to update one or a few fields (e.g. maindoc). The enduser should not need to know the json structure, or build it from scrach before he can send his modifications.

kkrebs:
@rsteph-de : Please can you check if this branch is still relevant? Than please rebase it and we should talk about it in our next vc.

PR #1615, opened 2022-07-05
MCR-2695 RESTv2 /objects use mcrmetahistory table and add pagination

rsteph-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2695).


golsch:
Why not use `offset` and `limit` with defaults as query params?

rsteph-de:
> Why not use `offset` and `limit` with defaults as query params?
There are problems 
a) if the data is changed from other users, while paginating through the results
b) On large resultsets its expensive for the database or SOLR to retrieve the last pages properly
   (usually there is a cursor which goes through all objects before the one's for the request page)
Using the last-known-id approach is more performant, since it can be coded as part of the query

sebhofmann (approved):

PR #1619, opened 2022-07-11
MCR-2699 new command 'reload mappings in jpa configuration file'

rsteph-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2699).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/frontend/cli/MCRBasicCommands.java
+++ b/mycore-base/src/main/java/org/mycore/frontend/cli/MCRBasicCommands.java
@@ -233,7 +233,7 @@ public class MCRBasicCommands {
         }
     }
 
-    @MCRCommand(syntax = "reload mappings in jpa configuration",
+    @MCRCommand(syntax = "reload mappings in jpa configuration file",
         help = "retrieves the mapping files from MyCoRe jars and adds them to the jpa configuration file.",
         order = 140)
     public static void reloadJPAMappings() throws IOException {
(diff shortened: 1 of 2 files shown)


rsteph-de:
Kathleen suggested to make the command string even more verbose:
reload mappings in jpa configuration **file** ?

yagee-de:
Can we make a convention that all "-mappings.xml" files in META-INF are detected and all present mapping files are kept, if they exist? The command right now is limited to "mycore-" mappings but the command is genric: "reload mappings"

rsteph-de:
1) I was not sure about ${config.dir}\resources\META-INF\mycore-jpa-defaults.xml
This file does not follow the naming convention and I did not want to "loose" it, if anybody has configured it in persistence.xml - Is this file still useful - it handles database schema settings, which in most cases can be configured by properties or jdbc-URL.

2) I did not want to mess with other mappings, which might not be in scope of MyCoRe.
    But I don't see anybody using MyCoRe this way at the moment.
I can check if the mapping files exists before I delete them and only add those new files which are not in persistence.xml

There is a szenario where somebody might remove a dependency (e.g. mycore-pi) later and needs to cleanup the persistence.xml again - but that should be covered.

mycore-base/src/main/java/org/mycore/frontend/cli/MCRBasicCommands.java
```diff
@@ -219,6 +233,74 @@ private static void createSampleConfigFile(String path) throws IOException {
...
+                try (BufferedWriter bw = Files.newBufferedWriter(persistenceXMLFile.toPath())) {
+                    out.output(doc, bw);
+                }
+
+            } else {
+                LOGGER.warn("The config file '" + persistenceXMLFile.toString() + "' does not exist yet!");
+            }
+        } catch (Exception e) {
+            LOGGER.error(e.getMessage(), e);
+        }
```
yagee-de: Do not catch and log exception, throw it! We need the exception to detect if it ran successfully or not.

mycore-base/src/main/java/org/mycore/frontend/cli/MCRBasicCommands.java
```diff
@@ -219,6 +233,74 @@ private static void createSampleConfigFile(String path) throws IOException {
...
+                        ePersistenceUnit.addContent(++pos, eMappingFile);
+                    }
+                    LOGGER
+                        .warn(newMappings.size() + " mapping files added to '" + persistenceXMLFile.toString() + "'.");
+                }
+
+                XMLOutputter out = new XMLOutputter(Format.getPrettyFormat());
+                try (BufferedWriter bw = Files.newBufferedWriter(persistenceXMLFile.toPath())) {
+                    out.output(doc, bw);
+                }
```
yagee-de: It would be good to only update the file if a change requires it

yagee-de (changes requested):

mycore-base/src/main/java/org/mycore/frontend/cli/MCRBasicCommands.java
rsteph-de: rethrown as MCRException

mycore-base/src/main/java/org/mycore/frontend/cli/MCRBasicCommands.java
rsteph-de: use a modified flag to lazyly rewrite the persistence.xml file

yagee-de (approved):

PR #1647, opened 2022-08-18
MCR-2719 fix MCRPICreationEventHandler already exists check

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2719).


sebhofmann:
Fixed by https://github.com/MyCoRe-Org/mycore/pull/1649

PR #1650, opened 2022-08-23
MCR-2721 Support listeners in MCRJobThread

Possommi:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2721).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/services/queuedjob/MCRDefaultJobStatusListener.java
+++ b/mycore-base/src/main/java/org/mycore/services/queuedjob/MCRDefaultJobStatusListener.java
@@ -36,7 +36,7 @@ public class MCRDefaultJobStatusListener implements MCRJobStatusListener {
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


mycore-base/src/main/java/org/mycore/services/queuedjob/MCRDefaultJobStatusListener.java
```diff
@@ -0,0 +1,47 @@
...
+
+    protected static Logger LOGGER = LogManager.getLogger(MCRDefaultJobStatusListener.class);
+
+    @Override
+    public void onProcessing(MCRJob job) {
+        LOGGER.debug("Processing {}", job.getAction().getName());
+    }
+
+    @Override
+    public void onFinish(MCRJob job) {
```
toKrause: `onSuccess` would be a better name, cause this callback is only called, when processing finishes successfully

toKrause (changes requested):

toKrause (approved):

PR #1652, opened 2022-08-26
MCR-2701 use Saxon by default for XSLT

yagee-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2701).

What should be the target branch for this?


rsteph-de (approved):
looks good for me

PR #1655, opened 2022-08-31
MCR-2726 fix Webupload is not working in FF and Safari

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2726).


Possommi (approved):
Upload works now in FF, Safari and Chrome.

PR #1656, opened 2022-08-31
MCR-2698 add mods 3.8 support

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2698).


yagee-de:
MODS 3.8 is [not yet released](https://www.loc.gov/standards/mods/mods-schemas.html) and the XSD is still a draft document. We should wait for the release as it still may change.

fluetze:
> MODS 3.8 is [not yet released](https://www.loc.gov/standards/mods/mods-schemas.html) and the XSD is still a draft document. We should wait for the release as it still may change.

We urgently need the datamodel features coming with 3.8, and Ido not expect major changes in the standard until release. So I'd prefer to have 3.8 draft support as soon as possible in the core.

sebhofmann:
Fixed by https://github.com/MyCoRe-Org/mycore/pull/1659

PR #1662, opened 2022-09-12
MCR-2731 Add configurable list of x-languages that may be edited using the classification editor

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2731).

### DIFF
--- a/mycore-classeditor/src/main/resources/components/classeditor/config/mycore.properties
+++ b/mycore-classeditor/src/main/resources/components/classeditor/config/mycore.properties
@@ -2,3 +2,5 @@ MCR.Jersey.Resource.Packages=%MCR.Jersey.Resource.Packages%,org.mycore.frontend.
 MCR.LayoutTransformerFactory.Default.Ignore=%MCR.LayoutTransformerFactory.Default.Ignore%,classificationEditor
 
 MCR.Module-solr.linkQuery=category.top:{0}
+
+MCR.classeditor.xlanguages=disable,display,group,uri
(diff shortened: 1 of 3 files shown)


mycore-classeditor/src/main/resources/components/classeditor/config/mycore.properties
```diff
@@ -2,3 +2,5 @@ MCR.Jersey.Resource.Packages=%MCR.Jersey.Resource.Packages%,org.mycore.frontend.
 MCR.LayoutTransformerFactory.Default.Ignore=%MCR.LayoutTransformerFactory.Default.Ignore%,classificationEditor
 
 MCR.Module-solr.linkQuery=category.top:{0}
+
+MCR.classeditor.xlanguages=disable,display,group,uri
```
rsteph-de: In my opinion it would be easier to keep the "x-" prefix in the property:
`MCR.classeditor.xlanguages=x-disable,x-display,x-group,x-uri`

I do not remember any usecase where you need the language id without the prefix "x-"?

toKrause: > I do not remember any usecase where you need the language id without the prefix "x-"?

That's my thinking as well. Because all relevant languages always have the `x-`-prefix, it is unnecessary to include this prefix in every configured value. Omitting them, saves some keystrokes and prevents inclusion of non-prefixed languages.

But I have no strong opinion here and can certainly live with the proposed change.

Should the property than be called `MCR.classeditor.languages` (or even `MCR.classeditor.additionalLanguages`) instead of `MCR.classeditor.xlanguages`? Because withot the automatic inclusion, it would be possible to add languages without the `x-`-prefix; say `sgn` (Sign Languages), ...

rsteph-de: I like "additionalLanguages", as you said, that gives us the opportunity to add "our x-languages" and possible other languages without the "x"- prefix.

rsteph-de (dismissed):
looks good for me now

kkrebs:
can you rebase this to 2022.06.x?

toKrause:
> can you rebase this to 2022.06.x?

Rebased. https://github.com/MyCoRe-Org/mir/pull/797 as well.

kkrebs (approved):

PR #1668, opened 2022-09-16
MCR-2735 improve OCFL mycore-slot-layout implementation

rsteph-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2735).


mycore-ocfl/src/main/java/org/mycore/ocfl/layout/MCRStorageLayoutConfig.java
```diff
@@ -56,51 +52,16 @@ public boolean hasParameters() {
      * @param slotLayout MyCoRe Class SlotLayout, see MCRStore for more info
```
Hammer1279: I would probably remove the Class from here since its now generic.
```suggestion
     * @param slotLayout MyCoRe SlotLayout, see MCRStore for more info
```

Hammer1279 (changes requested):
The rest looks good to me.

Hammer1279 (approved):

PR #1675, opened 2022-09-21
MCR-2724 OCFL Users

Hammer1279:
This adds:

Events to MCRUser changes for create, update and delete

OCFL XML UserManger

Commands to manage users in OCFL repository

The UserManager creates, just like with the classifications, backups of the entries within the database, which can be used to write user data back when needed.

This also implements [MCR-2746](https://mycore.atlassian.net/browse/MCR-2746) for user objects.

[Link to jira](https://mycore.atlassian.net/browse/MCR-2724).

### DIFF
--- a/mycore-ocfl/src/main/java/org/mycore/ocfl/commands/MCROCFLCommands.java
+++ b/mycore-ocfl/src/main/java/org/mycore/ocfl/commands/MCROCFLCommands.java
@@ -142,15 +142,13 @@ public class MCROCFLCommands {
         if (MCRUserManager.getUser(userId) == null) {
             throw new MCRUsageException("The User '" + userId + "' does not exist!");
         }
-        new MCROCFLXMLUserManager(MCRConfiguration2.getStringOrThrow("MCR.Users.Manager.Repository"))
-            .updateUser(MCRUserManager.getUser(userId));
+        new MCROCFLXMLUserManager().updateUser(MCRUserManager.getUser(userId));
     }
 
     @MCRCommand(syntax = "delete ocfl user {0}",
         help = "Delete user {0} in the OCFL Store")
     public static void deleteOCFLUser(String userId) {
-        new MCROCFLXMLUserManager(MCRConfiguration2.getStringOrThrow("MCR.Users.Manager.Repository"))
-            .deleteUser(userId);
+        new MCROCFLXMLUserManager().deleteUser(userId);
     }
 
     @MCRCommand(syntax = "sync ocfl users",
@@ -168,8 +166,7 @@ public class MCROCFLCommands {
     @MCRCommand(syntax = "restore database user {0} with version {1} from ocfl",
         help = "restore a specified revision of a ocfl user backup to the database")
     public static void writeUserToDb(String userId, String revision) throws IOException {
-        MCRUser user = new MCROCFLXMLUserManager(MCRConfiguration2.getStringOrThrow("MCR.Users.Manager.Repository"))
-            .retrieveContent(userId, revision);
+        MCRUser user = new MCROCFLXMLUserManager().retrieveContent(userId, revision);
         MCRUserManager.updateUser(user);
     }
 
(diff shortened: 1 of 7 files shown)


mycore-ocfl/src/main/resources/components/ocfl/config/mycore.properties
```diff
@@ -68,3 +68,13 @@ MCR.Classification.Manager.Repository=Main
...
+
+######################################################################
+#                      OCFL Users Configuration                      #
+######################################################################
+
+# bind OCFL event handler
+# MCR.EventHandler.MCRUser.020.Class=org.mycore.ocfl.user.MCROCFLUserEventHandler
+
+# default user repository
+MCR.Users.Manager.Repository=Main
```
rsteph-de, line 80: I would like to see "OCFL" in the property key,
like `MCR.Users.OCFL.Repository ` (prefered) or `MCR.Users.Manager.OCFL.Repository`

But this is a new ticket and should be handled together with classifications and metadata properties after merging this PR

mycore-ocfl/src/main/java/org/mycore/ocfl/commands/MCROCFLCommands.java
```diff
@@ -121,6 +126,50 @@ public static List<String> syncClassificationRepository() {
...
+        List<String> commands = new ArrayList<>();
+        commands.add("update ocfl users");
+        List<String> outOfSync = getStaleOCFLUserIDs();
+        commands.addAll(
+            outOfSync.stream()
+                .map(id -> "delete ocfl user " + id).collect(Collectors.toList()));
+        return commands;
+    }
+
+    @MCRCommand(syntax = "restore database user {0} with version {1} from ocfl",
```
rsteph-de: We never use "database" in the user commands
I would prefer "restore user {0} with version {1} from ocfl"
Or maybe "restore user {0} from ocfl with version {1}"
(Usually we have a parameter at the end of the command)

Or even better to match the other new commands: "restore ocfl user {0} with version {1}:
Do we also need a "restore ocfl user {0}" without version param, which restores the latest version from OCFL?

rsteph-de (changes requested):

mycore-ocfl/src/main/java/org/mycore/ocfl/user/MCROCFLXMLUserManager.java
```diff
@@ -0,0 +1,224 @@
...
+import edu.wisc.library.ocfl.api.model.ObjectVersionId;
+import edu.wisc.library.ocfl.api.model.VersionInfo;
+
+/**
+ * XML Manager to handle MCRUsers in a MyCoRe OCFL Repository
+ * @author Tobias Lenhardt [Hammer1279]
+ */
+public class MCROCFLXMLUserManager {
+
+    private static final String THE_USER = "The User '";
```
rsteph-de: I would not extract a part of a message into a variable, even if code quality tools would complain about redundant strings here.

mycore-ocfl/src/main/java/org/mycore/ocfl/commands/MCROCFLCommands.java
Hammer1279: I think `restore user {0} with version {1} from ocfl` suits this best, since the "ocfl" could be seen as parameter.

> Or even better to match the other new commands: "restore ocfl user {0} with version {1}:

This, at least for me, implies that the user within ocfl gets restored, so it could be misleading.

rsteph-de: O.K. If we see "ocfl" as parameter that's fine with me. 
I would still see the version parameter at the end of the command, because the version is specific to the OCFL repository.
But finally it's your decision.

mycore-ocfl/src/main/resources/components/ocfl/config/mycore.properties
Hammer1279: Made a new Ticket for this: [MCR-2761 - add "OCFL" into OCFL Property Keys](https://mycore.atlassian.net/browse/MCR-2761)

rsteph-de (approved):
Looks good for me - ready to merge

PR #1676, opened 2022-09-22
MCR-2739 prevent parallel usage of SchemaFactory

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2739).


fluetze (approved):
Works excellent for me

PR #1683, opened 2022-09-26
MCR-2743 Only merge titleInfo with matching type attribute

chemicus:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2743).


fluetze (approved):

kkrebs:
can yout take a look at the codacy requests

chemicus:
I don't understand why the Codacy warning is only reported for the two new Tests and not for the existing ones. Anyway, the assert() is implicitly included through MCRMergerTest.test(), so there isn't anything to be fixed.

PR #1684, opened 2022-09-29
MCR-2744 Classification browser with filterCategory=true creates invalid URL query parameters

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2744).


kkrebs (approved):

rsteph-de (approved):
checked, the method is not used in JSPDocportal  or RosDok

PR #1687, opened 2022-09-30
MCR-2748 Added null check

Possommi:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2748).


rsteph-de:
What's the advantage of the IlegalArgumentException instead of the NPE?
How it is handled within the OAI processing?

Possommi:
> What's the advantage of the IlegalArgumentException instead of the NPE? How it is handled within the OAI processing?

Currently it is unclear which line of the code is actually throwing the NPE. Usually you do not catch `NullPointerExceptions`. The NPE/IllegalArgumentException currently breaks the harvesting via OAI (when `modified` is null one could set modified to `new Date()` to avoid this). What do you suggest?

rsteph-de:
I would throw a NullPointerException instead of the IllegalArgumentException.
The later is mainly "Thrown to indicate that a method has been passed an illegal or inappropriate argument"  (Javadoc).
Since the Null-Value comes from internal Solr and not from method parameters a NullpointerException seems to be the better choice for me here.

rsteph-de (approved):
For me the code looks like before (one new line with LOGGER.debug() ) ?
The seperate steps to retrieve the oaiID may help to find the line where the NPE occurs.
This is an improvement and the Java compiler will combine them again.
-> O.K. for me

PR #1689, opened 2022-10-05
MCR-2751 Output classifications as SKOS

rsteph-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2751).


rsteph-de:
The basic  implementation should work, but we need to discuss and agree about the URI that we will use here.
The URI is the foundation in the linked data world and needs to be rock solid!
Some first ideas and considerations can be found in the ticket ([MCR-2751](https://mycore.atlassian.net/browse/MCR-2751)).

sebhofmann:
The converting code should be in an extra class, because its an extra task and it could be reused somewhere else.

rsteph-de:
At the moment this is work in progress.
The currently returned SKOS needs some improvements as well
At the end I see a new mycore-lod (linked open data) module.

PR #1690, opened 2022-10-06
MCR-2752 create mycore-bom Maven module and add <dependencyManagement>

rsteph-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2752).


sebhofmann:
Should be in main.

rsteph-de:
Normally I would agree, but as I see it, we are still in the transition phase between releases.
Therefore we could make an exeption and integrate this breaking change in the current LTS branch.
This feature would improve the build definition for other modules and MyCoRe apps whiche are not migrated to 2022.06 yet.
I do not want to wait another year until it can be used.

fluetze (approved):

PR #1691, opened 2022-10-07
MCR-2753 fix NPE with flags in ObjectService

rsteph-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2753).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/datamodel/metadata/MCRObjectService.java
+++ b/mycore-base/src/main/java/org/mycore/datamodel/metadata/MCRObjectService.java
@@ -509,7 +509,7 @@ public class MCRObjectService {
      */
     public final boolean isFlagSet(String value) {
         return MCRUtils.filterTrimmedNotEmpty(value)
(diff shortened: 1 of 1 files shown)


bt1cu (approved):

sebhofmann (dismissed):
The NPE is actually correct. By changing the equals you only cover up the real error, which is missing validation of the object.

sebhofmann:
A better place to fix that would be in the setFromDOM or in the validate method.

rsteph-de:
That depends ...
The given example was
```
   <servflags class="MCRMetaLangText"> 
     <servflag inherited="0" form="plain">State:original sample</servflag> 
   </servflags>
```

And because the `@type`-Attribute is optional in MCRMetaLangText I would not handle things different for servflags.
Which means that we must rework the API to allow `type=null` where it occurs.

rsteph-de:
> The NPE is actually correct. By changing the equals you only cover up the real error, which is missing validation of the object.

type attribute is optional - see XML Schema for MCRMetaLangText

PR #1695, opened 2022-10-10
MCR-2494 create unique path to xml

Mewel:
Path to content uses UID instead of static path

[Link to jira](https://mycore.atlassian.net/browse/MCR-2494).


kkrebs (approved):

rsteph-de (approved):
looks good for me

PR #1700, opened 2022-10-11
MCR-2757 set modify date to force update

Mewel:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2757).


kkrebs:
should also be fixed in 2021.06.x

PR #1709, opened 2022-10-13
MCR-2438 fix creation of new user

Mewel:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2438).


yagee-de:
Changing user name should be forbidden. I would like this PR not to be merged

PR #1712, opened 2022-10-14
MCR-2767 Dublin Core is an unsupported description type

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2767).

This adds explicit support for Dublin Core as a description type and harmonizes the display of the Rights Definition description.

### DIFF
--- a/mycore-oai/src/main/resources/META-INF/resources/oai/oai2.xsl
+++ b/mycore-oai/src/main/resources/META-INF/resources/oai/oai2.xsl
@@ -274,6 +274,16 @@ p.intro {
 
 
 <!--
+   Identify / Dublin Core Description
+-->
+
+<xsl:template match="oai:description[oai_dc:dc]" xmlns:oai_dc="http://www.openarchives.org/OAI/2.0/oai_dc/">
+  <h2>Description</h2>
+  <xsl:apply-templates select="oai_dc:dc"/>
+</xsl:template>
+
+
+<!--
    Identify / OAI-Identifier
 -->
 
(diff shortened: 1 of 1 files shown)


mycore-oai/src/main/resources/META-INF/resources/oai/oai2.xsl
```diff
@@ -339,20 +349,17 @@ p.intro {
...
-      </xsl:if>
-    </div>
+    <xsl:if test="ri:rightsReference">
+      <h3>Reference</h3>
+      <xsl:variable name="rightsReference" select="ri:rightsReference/@ref" />
+      <div><a href="{$rightsReference}"><xsl:value-of select="$rightsReference" /></a></div>
+    </xsl:if>
+    <xsl:if test="ri:rightsDefinition">
+      <xsl:apply-templates select="ri:rightsDefinition/*" />
+    </xsl:if>
```
yagee-de, line 362: Does not seems to be related to the ticket and removes <div> and @class. Why?

toKrause: As mentioned in the JIRA ticket and the initial post in this PR, I've also harmonized the visuals of the [recently added display of Rights Definition descriptions](https://github.com/MyCoRe-Org/mycore/pull/1694) (something that nobody should rely on as of yet) with the visuals of the newly added support for Dublin Core descriptions.

This could _technically_ be two separate PRs, but that seemed to be overly fine grained for me.

yagee-de (approved):

PR #1717, opened 2022-10-17
MCR-2761 "OCFL" in OCFL Property Keys

Hammer1279:
Changes Property Keys like:
`MCR.<Type>.Manager.Repository=Main`
to
`MCR.OCFL.<Type>.Repository=Main`

- added "OCFL" to ocfl property keys and reformed them
- fixed typo in OCFLPersistenceTransaction method name

[Link to jira](https://mycore.atlassian.net/browse/MCR-2761).

### DIFF
--- a/mycore-ocfl/src/main/java/org/mycore/ocfl/MCROCFLClassificationEventHandler.java
+++ b/mycore-ocfl/src/main/java/org/mycore/ocfl/MCROCFLClassificationEventHandler.java
@@ -18,7 +18,7 @@
 
 package org.mycore.ocfl;
 
-import static org.mycore.ocfl.MCROCFLPersistenceTransaction.addClassficationEvent;
+import static org.mycore.ocfl.MCROCFLPersistenceTransaction.addClassificationEvent;
 
 import java.util.Objects;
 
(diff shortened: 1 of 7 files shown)


rsteph-de (approved):
looks good for me

sebhofmann:
I don`t like this because the whole Idea of the Configurable instance and the MCR.<Type>.Manager classes is, that you can replace the Manager by sth. different e.g. a implementation for s3.
The properties of the manager are only for the specific class.

See also: https://www.mycore.de/documentation/basics/basics_configurable_instance/

Hammer1279:
See #1718

Hammer1279:
@sebhofmann to configure a different Manager, one could simply use a different event handler for a different implementation.
The events are coming from `mycore-base` itself, from the `MCREventedCategoryDAOImpl`.

The property `MCR.Classification.Manager` is not used by anything outside ocfl (and there only a single time) and on second thought, it should also not get used, every implementation should use its own EventHandler.

Its current implementation works more like a "on/off" switch.
While it could work, I don't think we should promote adding things like the ocfl module as a dependency when you can simply make a EventHandler for your own implementation, thus always ensuring it is always compliant with its own implementation.

While it should get removed, it's beyond the scope of the ticket and should be handled as a new one.
If it should be addressed here already, please let me know.

kkrebs:
We should first end up with #1718

yagee-de:
closed as discussed at 2022-10-26

PR #1720, opened 2022-10-19
MCR-2773 Rest api v1 file upload copies two times to derivate

sebhofmann:
Remove the source attribute to prevent the metadata manager from copying the uploaded contents a second time.

[Link to jira](https://mycore.atlassian.net/browse/MCR-2773).


rsteph-de (approved):
works for me

PR #1722, opened 2022-10-19
MCR-2773 do not use sourcepath in restapi

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2773).


rsteph-de (approved):
looks good for me

PR #1736, opened 2022-11-07
MCR-2779 empty nameIdentifier results in exception

eckstueck:
- skip empty nameIdentifier

[Link to jira](https://mycore.atlassian.net/browse/MCR-2779).

### DIFF
--- a/mycore-orcid/src/main/java/org/mycore/orcid/user/MCRORCIDUser.java
+++ b/mycore-orcid/src/main/java/org/mycore/orcid/user/MCRORCIDUser.java
@@ -145,13 +145,12 @@ public class MCRORCIDUser {
 
         List<Element> nameIdentifiers = wrapper.getElements(xPath);
         for (Element nameIdentifier : nameIdentifiers) {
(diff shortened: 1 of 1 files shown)


golsch:
Perhaps a log message would still be useful if the problem occurs. Or are there identifiers which are inserted without type attribute?

fluetze (changes requested):
This is jsut a workaround for inconsistent stored data that actually obfucates the root cause. There never should be a stored publication with empty name identifiers!

kkrebs (approved):

eckstueck:
For my understanding the MODS Schema allows empty nameIdentifier. So it is possible that you run in to this problem.

fluetze:
So, what does that mean if our data contains such an empty name identifier? How do we use and visualize that.
From my point of view, an empty name identifier is just a bug in the data. I'd then assume that some cleanup or import process failed - and we should fix THAT process, not the ORCID code - which should rely on meaningful data.

kkrebs (approved):

kkrebs (changes requested):
please only allow empty type and fix codacy error

kkrebs (approved):

PR #1738, opened 2022-11-11
MCR-2790 bad default filename

bt1cu:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2790).

In class MCRLayoutService method getFileName Line 162/163 the default generate a filename like MCRLpzBasketServlet-1,668,077,879,683 or MCRLpzBasketServlet-1.668.077.879.683 . This is not good. it should be independend from the Locale and it should be a String like MCRLpzBasketServlet-1668077879683 .

### DIFF
--- a/mycore-base/src/main/java/org/mycore/common/xml/MCRLayoutService.java
+++ b/mycore-base/src/main/java/org/mycore/common/xml/MCRLayoutService.java
@@ -161,7 +161,7 @@ public class MCRLayoutService {
             return extractFileName(req.getPathInfo());
         }
         return new MessageFormat("{0}-{1}", Locale.ROOT).format(
-            new Object[] { extractFileName(req.getServletPath()), System.currentTimeMillis() });
+            new Object[] { extractFileName(req.getServletPath()), String.valueOf(System.currentTimeMillis())});
     }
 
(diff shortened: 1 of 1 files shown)


mycore-base/src/main/java/org/mycore/common/xml/MCRLayoutService.java
```diff
@@ -161,7 +161,7 @@ private String getFileName(HttpServletRequest req, MCRParameterCollector paramet
             return extractFileName(req.getPathInfo());
         }
         return new MessageFormat("{0}-{1}", Locale.ROOT).format(
-            new Object[] { extractFileName(req.getServletPath()), System.currentTimeMillis() });
```
rsteph-de, line 164: The change should work and fix the issue.
But I would remove the MessageFormat completely and simply do a String concatenation or use [String.join()](https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#join-java.lang.CharSequence-java.lang.CharSequence...-)
`return String.join("-", extractFileName(req.getServletPath(), System.currentTimeMillis());`

Maybe we should replace the other usage of MessageFormat in this class too.

kkrebs (approved):

PR #1757, opened 2022-11-29
MCR-2802 use local test data

yagee-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2802).


mycore-base/src/test/java/org/mycore/datamodel/objectinfo/MCRObjectInfoEntityQueryResolverTest.java
```diff
@@ -88,8 +89,9 @@ private static void initClassifications() throws SAXParseException, MalformedURL
         DAO = new MCRCategoryDAOImpl();
         CLS = MCRCategLinkServiceFactory.getInstance();
 
-        Document xml = MCRXMLParserFactory.getParser()
-            .parseXML(new MCRURLContent(new URL("https://mycore.de/classifications/mir_licenses.xml")));
+        URL classResourceUrl = MCRCategoryDAOImplTest.class
+            .getResource("/" + MCRCategoryDAOImplTest.class.getSimpleName() + "/mir_licenses.xml");
```
toKrause: Should use `MCRObjectInfoEntityQueryResolverTest.class` in both places.

toKrause (changes requested):

mycore-base/src/test/java/org/mycore/datamodel/objectinfo/MCRObjectInfoEntityQueryResolverTest.java
toKrause: I know that this would require another copy of the (possibly truncated) `mir_licenses.xml`, but as I see it, the latest precedent is, that every test class should have its own resource folder with its own resources. If that means duplicating some stuff, than that is the natural side effect of this approach and should be followed through consistently.

yagee-de: use mycore-classifications as test dependency instead

mycore-base/src/test/java/org/mycore/datamodel/objectinfo/MCRObjectInfoEntityQueryResolverTest.java
```diff
@@ -88,8 +89,8 @@ private static void initClassifications() throws SAXParseException, MalformedURL
         DAO = new MCRCategoryDAOImpl();
         CLS = MCRCategLinkServiceFactory.getInstance();
 
-        Document xml = MCRXMLParserFactory.getParser()
-            .parseXML(new MCRURLContent(new URL("https://mycore.de/classifications/mir_licenses.xml")));
+        URL classResourceUrl = MCRCategoryDAOImplTest.class.getResource("/mycore-classifications/mir_licenses.xml");
```
toKrause: This still uses the wrong class object. `MCRObjectInfoEntityQueryResolverTest` should use `MCRObjectInfoEntityQueryResolverTest.class` as a reference when loading resources.

toKrause (changes requested):

mycore-base/src/test/java/org/mycore/datamodel/objectinfo/MCRObjectInfoEntityQueryResolverTest.java
sebhofmann: Should be fixed by [22c5cc0](https://github.com/MyCoRe-Org/mycore/pull/1757/commits/22c5cc0f9b20ae08c66d4337059e33b1d631059d)

toKrause (approved):

mycore-base/src/test/java/org/mycore/datamodel/objectinfo/MCRObjectInfoEntityQueryResolverTest.java
toKrause: Codacy dosn't like the change though, ...

PR #1771, opened 2022-12-13
MCR-2802 use local test data in viewer tests

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2802).


sebhofmann:
This does not work, because the test files need to have the right format. Because the tests zoom in and take screenshots and compare the pixels.

PR #1774, opened 2022-12-20
MCR-2809-MCR-2810 MCRUploadResource

eckstueck:
- fixed MCRUploadResource will not set mainfile if xml is the only file
- added property to allow selection of ignored mainfiles

[Link to jira](https://mycore.atlassian.net/browse/MCR-MCR-2809).
[Link to jira](https://mycore.atlassian.net/browse/MCR-MCR-2810).

### DIFF
--- a/mycore-webtools/src/main/resources/components/webtools/config/mycore.properties
+++ b/mycore-webtools/src/main/resources/components/webtools/config/mycore.properties
@@ -5,3 +5,6 @@
 
 MCR.Jersey.Resource.Packages=%MCR.Jersey.Resource.Packages%,org.mycore.webtools.session,org.mycore.webtools.upload
 MCR.URIResolver.xslIncludes.components=%MCR.URIResolver.xslIncludes.components%,sessionListing.xsl,processingGUI.xsl,msie-warning.xsl
+
+# filetypes that will be ignored for mainfile, if there are others
+MCR.Upload.IgnoreFiletypeForMainfile=".xml"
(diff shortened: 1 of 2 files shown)


sebhofmann:
If you upload a xml file before, then it would never be set as main file, after this change it would be set to main file if there is no other file. I don´t know if this has some unwanted side effects.

mycore-webtools/src/main/resources/components/webtools/config/mycore.properties
```diff
@@ -5,3 +5,6 @@
 
 MCR.Jersey.Resource.Packages=%MCR.Jersey.Resource.Packages%,org.mycore.webtools.session,org.mycore.webtools.upload
 MCR.URIResolver.xslIncludes.components=%MCR.URIResolver.xslIncludes.components%,sessionListing.xsl,processingGUI.xsl,msie-warning.xsl
+
+# filetypes that will be ignored for mainfile, if there are others
+MCR.Upload.IgnoreFiletypeForMainfile=".xml"
```
sebhofmann: The `"` in the property seems to be wrong

sebhofmann (changes requested):

mycore-webtools/src/main/resources/components/webtools/config/mycore.properties
eckstueck: Yes, thanks for the hint

sebhofmann (approved):

PR #1777, opened 2023-01-05
MCR-2812 MCRAutoDeploy ignores init-params

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2812).


yagee-de (approved):
untested, but looks good

PR #1778, opened 2023-01-05
MCR-2813 Allow automatic PI creation for non-job-based PI services

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2813).


sebhofmann (approved):
Should be ok

PR #1779, opened 2023-01-05
MCR-2814 XPaths resulting in a boolean value don't work as PI predicates

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2814).


sebhofmann (approved):
Seems to be ok

According to this:
https://developer.mozilla.org/en-US/docs/Web/XPath/Functions/boolean#notes
- A number evaluates to false if it is positive or negative zero or NaN. Otherwise, it evaluates true.
- A node-set evaluates to true if it is non-empty.
- A string evaluates to false if it an empty string. Otherwise, it evaluates to true.
- An object of a type other than the four basic types is converted to a boolean in a way that is dependent on that type.

The behavior for numbers is now different.

sebhofmann (approved):

PR #1787, opened 2023-01-18
MCR-2822 IIIF Thumbnail also use linked image

eckstueck:
added function to use linked image if there is no derivate

[Link to jira](https://mycore.atlassian.net/browse/MCR-2822).

### DIFF
--- a/mycore-iview2/src/main/java/org/mycore/iview2/iiif/MCRThumbnailImageImpl.java
+++ b/mycore-iview2/src/main/java/org/mycore/iview2/iiif/MCRThumbnailImageImpl.java
@@ -27,6 +27,7 @@ import java.util.Set;
 
 import org.mycore.access.MCRAccessManager;
 import org.mycore.datamodel.classifications2.MCRCategoryID;
+import org.mycore.datamodel.common.MCRLinkTableManager;
 import org.mycore.datamodel.metadata.MCRDerivate;
 import org.mycore.datamodel.metadata.MCRMetaEnrichedLinkID;
 import org.mycore.datamodel.metadata.MCRMetadataManager;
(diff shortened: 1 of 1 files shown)


mycore-iview2/src/main/java/org/mycore/iview2/iiif/MCRThumbnailImageImpl.java
```diff
@@ -66,15 +67,33 @@ protected MCRTileInfo createTileInfo(String id) throws MCRIIIFImageNotFoundExcep
...
+                                maindoc);
+                            if (tileInfoForFile.isPresent()) {
+                                return tileInfoForFile.get();
                             }
                         }
                     }
                 }
             }
+            Optional<String> linkedImage = MCRLinkTableManager.instance()
+                .getDestinationOf(mcrID, MCRLinkTableManager.ENTRY_TYPE_DERIVATE_LINK).stream().findFirst();
```
sebhofmann: The implementation iterates over all derivates and the first one which has a tiles will be used. 
It should work the same way with derivate_links and not only consider the first link.

mycore-iview2/src/main/java/org/mycore/iview2/iiif/MCRThumbnailImageImpl.java
```diff
@@ -66,15 +67,33 @@ protected MCRTileInfo createTileInfo(String id) throws MCRIIIFImageNotFoundExcep
...
+            if (linkedImage.isPresent()) {
+                MCRObjectID deriID = MCRObjectID.getInstance(
+                    linkedImage.get().substring(0, linkedImage.get().indexOf("/")));
+                MCRDerivate deri = MCRMetadataManager.retrieveMCRDerivate(deriID);
+                if (deri.getDerivate().getClassifications().stream()
+                    .map(classi -> classi.getClassId() + ":" + classi.getCategId())
```
sebhofmann: This can be MCRCategoryID::toString

sebhofmann (changes requested):
Not direct related to the branch but i think `for (String type : derivateTypes) {` in line 62 and `.anyMatch(type::equals)` in line 66 could be replaced with `.anyMatch(derivateTypes::contains)`

mycore-iview2/src/main/java/org/mycore/iview2/iiif/MCRThumbnailImageImpl.java
eckstueck: deri.getDerivate().getClassifications() returns a ArrayList<MCRMetaClassification> and MCRMetaClassification::toString does not return the right format.

rsteph-de: May be we should a the following methods to `MCRMetaClassification`:
```
public final MCRCategoryID getCategoryID(){
        return category;
}

public final void setValue(MCRCategoryID category) throws MCRException {
        this.category = category;
    }
```
In my opinion it is useful to have the MCRCategoryID object available.

eckstueck: Should I add this in this Pull request?

rsteph-de: @sebhofmann  and I came to the conclusion to leave it as it is for now

sebhofmann (approved):

mycore-iview2/src/main/java/org/mycore/iview2/iiif/MCRThumbnailImageImpl.java
```diff
@@ -49,40 +51,90 @@ public MCRThumbnailImageImpl(String implName) {
...
+                }
+            }
+        }
+        return Optional.empty();
+    }
+
+    private Optional<MCRTileInfo> createTileInfoForFile(String derID, String file) {
+        final MCRTileInfo mcrTileInfo = new MCRTileInfo(derID, file, null);
+        final Optional<Path> tileFile = this.tileFileProvider.getTileFile(mcrTileInfo);
+        if (tileFile.isPresent() && Files.exists(tileFile.get())) {
```
sebhofmann: can be written as return tileFile.filter(Files::exists);

eckstueck: I cant, because i need to return the MCRTileInfo and not the Path.

rsteph-de (approved):
O.K. for me

mycore-iview2/src/main/java/org/mycore/iview2/iiif/MCRThumbnailImageImpl.java
sebhofmann: Then you could use:
```
       final MCRTileInfo tileInfo = new MCRTileInfo(derID, file, null);
       return Optional.of(tileInfo).filter(t-> this.tileFileProvider.getTileFile(t).filter(Files::exists).isPresent());
```

PR #1789, opened 2023-01-26
MCR-2823 fix NPE in MCRLogEventProcessor

yagee-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2823).


Possommi (approved):
It works.

PR #1793, opened 2023-01-30
MCR-2829 Use appropriate file extension during export

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2829).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/common/content/transformer/MCRXSLTransformer.java
+++ b/mycore-base/src/main/java/org/mycore/common/content/transformer/MCRXSLTransformer.java
@@ -28,7 +28,6 @@ import java.util.Optional;
 import java.util.Properties;
 
 import javax.xml.parsers.ParserConfigurationException;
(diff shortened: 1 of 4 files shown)


mycore-base/src/main/java/org/mycore/common/xml/MCRXMLUtils.java
```diff
@@ -0,0 +1,62 @@
...
+package org.mycore.common.xml;
+
+import org.mycore.common.config.MCRConfiguration2;
+
+import javax.xml.transform.OutputKeys;
+import javax.xml.transform.Transformer;
+import java.util.Map;
+import java.util.Properties;
+
+public class MCRXMLUtils {
```
yagee-de: class should be named MCRXSLTransformerUtils instead

yagee-de (changes requested):
Rest looks good

yagee-de (approved):

PR #1805, opened 2023-02-12
MCR-2840 URN registering not working

chiliego:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2840).


yagee-de:
please re-format code

mycore-pi/src/main/java/org/mycore/pi/urn/rest/MCRDerivateURNUtils.java
```diff
@@ -83,7 +83,7 @@ public static URL getURL(MCRPIRegistrationInfo piInfo) {
                     String filePath = "/" + file.getOwner() + "/" + file.getFileName();
                     return new URL(
                         MCRFrontendUtil.getBaseURL() + "servlets/" + MCRFileNodeServlet.class.getSimpleName()
-                            + filePath);
+                            + filePath.replaceAll(" ", "%20"));
```
yagee-de: Do not URL encode like this. There will be more cases of characters that need to be percent encoded.

Possommi (approved):

kkrebs (changes requested):
some of this changes are not compatible with 2022.06 and not needed to fix this problem
please minimize this commit to make a merge to 2022.06 possible without a lot of conflicts
renaming eg. should be done in 2022.06 or main

please also resolve comment from yagee

chiliego:
> some of this changes are not compatible with 2022.06 and not needed to fix this problem please minimize this commit to make a merge to 2022.06 possible without a lot of conflicts renaming eg. should be done in 2022.06 or main

Please tell me which changes are not compatible with 2022.06, so I can make adjustments.

chiliego:
> some of this changes are not compatible with 2022.06 and not needed to fix this problem please minimize this commit to make a merge to 2022.06 possible without a lot of conflicts renaming eg. should be done in 2022.06 or main
> 

When merge with 2022.06.x I can see 3 conflicting files: MCRDNBURNRestClient.java, MCRURNGranularRESTRegistrationTask.java and MCRURNGranularRESTServiceTest.java. 

The conflict in MCRURNGranularRESTServiceTest.java comes from changes directly in 2022.06.x, where JMockit was removed and was not introduced by changes in this issue branch.
The conflict in MCRURNGranularRESTRegistrationTask.java is removing two imports which are never used in this class.
 
The changes in MCRDNBURNRestClient.java fix the logs so you can see the real error, which helped me solving the URN registering problem! So this is also a fix!

> please also resolve comment from yagee

is done

PR #1812, opened 2023-03-01
MCR-2847 Allow test case specific test properties

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2847).

### DIFF
--- a/mycore-base/src/test/java/org/mycore/common/MCRTestConfiguration.java
+++ b/mycore-base/src/test/java/org/mycore/common/MCRTestConfiguration.java
@@ -25,7 +25,7 @@ import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
 @Inherited
-@Target({ElementType.METHOD})
+@Target({ElementType.METHOD, ElementType.TYPE})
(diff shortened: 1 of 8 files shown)


mycore-base/src/test/java/org/mycore/common/MCRTestConfiguration.java
```diff
@@ -0,0 +1,34 @@
...
+package org.mycore.common;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Inherited;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Inherited
+@Target({ElementType.METHOD})
```
sebhofmann: Maybe it could be useful to also allow class properties which will be used for all test method runs.

sebhofmann (approved):

mycore-base/src/test/java/org/mycore/common/MCRTestConfiguration.java
toKrause: seems reasonable, will update

sebhofmann (approved):

PR #1830, opened 2023-04-04
MCR-2866 link and cors header are not exposed via cors

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2866).


rsteph-de (approved):
looks good for me

PR #1834, opened 2023-04-19
MCR-2870 Make sure output buffer gets read during wait for end of pro…

chemicus:
…cess to prevent deadlock

[Link to jira](https://mycore.atlassian.net/browse/MCR-2870).


rsteph-de (approved):
looks good fre me

fluetze:
How is this knowledge related to
https://github.com/MyCoRe-Org/mycore/blob/2022.06.x/mycore-base/src/main/java/org/mycore/frontend/cli/MCRExternalProcess.java
including the most famous "MCRStreamSucker" ?
Can that class be improved? Or can it also be used in this case?

kkrebs:
We already discussed this possibility of improvement in our skype session but still decided to merge this pr. Perhaps @chemicus will find some time to take a look into MCRStreamSucker and make a new ticket ...

PR #1839, opened 2023-05-05
MCR-2872 revert to old regexp variant and print warning if variants mismatch

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2872).


Possommi (approved):
Tested in Collections@UrMEL
Links in dashboard (using unencoded queries) and child listing (using a request handler) are working again.

PR #1841, opened 2023-05-17
MCR-2885 proposal for a start of a more controlled handling of "alien objects"…

vs-gsi:
… in data directory – fixes metadata rebuild command while arbitrary files present in data dirs (project/type)

[Link to jira](https://mycore.atlassian.net/browse/MCR-2885).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/datamodel/common/MCRDefaultXMLMetadataManager.java
+++ b/mycore-base/src/main/java/org/mycore/datamodel/common/MCRDefaultXMLMetadataManager.java
@@ -625,11 +625,10 @@ public class MCRDefaultXMLMetadataManager implements MCRXMLMetadataManagerAdapte
      */
     //TODO: consider sanity checks like getObjectTypes() does. (performance penalty?)
     private ProjectDirState projectDirStatus(Path path) {
-    	File p = new File(path.toString());
-    	if (!p.isDirectory()) {
+    	if (Files.isDirectory(path)) {
        		LOGGER.warn(
     			"File '{}' spotted in data-directory where only subdirectories are expected.",
-    			p.getAbsolutePath()
+    			path.toAbsolutePath()
     		);
     		return ProjectDirState.UNUSABLE;
     	}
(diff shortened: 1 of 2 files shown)


yagee-de:
We should discuss this in a video call together and need to improve our contribute documentation a bit.

vs-gsi:
just updated the PR

mycore-base/src/main/java/org/mycore/datamodel/common/MCRDefaultXMLMetadataManager.java
```diff
@@ -592,7 +598,44 @@ public Collection<String> getObjectBaseIds() {
...
+    /**
+     * 
+     * @param path
+     * @return: CLEAN or UNUSABLE
+     */
+    //TODO: consider sanity checks like getObjectTypes() does. (performance penalty?)
+    private ProjectDirState projectDirStatus(Path path) {
+    	File p = new File(path.toString());
+    	if (!p.isDirectory()) {
+       		LOGGER.warn(
```
rsteph-de, line 629: try to avoid `File p` and use `Files.isDirectory(path)` instead

rsteph-de (changes requested):

vs-gsi:
(updated after @rsteph-de pointed out today, that i should use pure nio here)

mycore-base/src/main/java/org/mycore/datamodel/common/ProjectDirState.java
```diff
@@ -0,0 +1,7 @@
+package org.mycore.datamodel.common;
+
+public enum ProjectDirState {
```
rsteph-de, line 3: Is this enum still needed? Or can it be simplified to a boolean?

vs-gsi: The enum was proposed by someone else and i think this solution has both advantages included:
* We have a (minimalistic) stable boolean wrapper function
* The Enum can be freely extended without changing any current code dependent on it
(Or at least minimal&obvious changes if the inequality should go in another direction longer term for whatever reason)
=> to advertise it a bit: maximum flexibility with minimalistic (but still guiding) implementation

sebhofmann:
Stale

PR #1844, opened 2023-05-24
MCR-2877 MCRServlets can be disabled by properties

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2877).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/frontend/servlets/MCRServlet.java
+++ b/mycore-base/src/main/java/org/mycore/frontend/servlets/MCRServlet.java
@@ -38,6 +38,7 @@ import java.util.Properties;
 
 import javax.xml.transform.TransformerException;
 
+import jakarta.servlet.UnavailableException;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.mycore.common.MCRException;
@@ -87,9 +88,6 @@ public class MCRServlet extends HttpServlet {
 
     private static final boolean ENABLE_BROWSER_CACHE = MCRConfiguration2.getBoolean("MCR.Servlet.BrowserCache.enable")
         .orElse(false);
-
-    private boolean disabled;
-
     private static MCRLayoutService LAYOUT_SERVICE;
 
     private static String ACCESS_CONTROL_ALLOW_ORIGIN = "Access-Control-Allow-Origin";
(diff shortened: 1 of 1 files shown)


mycore-base/src/main/java/org/mycore/frontend/servlets/MCRServlet.java
```diff
@@ -103,6 +106,18 @@ public void init() throws ServletException {
...
 
+    @Override
+    public void init(ServletConfig config) throws ServletException {
+        super.init(config);
+        String servletName = config.getServletName();
+        disabled = MCRConfiguration2.getBoolean("MCR.Servlet." + servletName + ".Disabled")
+            .orElse(false);
+
+        if (disabled) {
+            LOGGER.info("Servlet {} is disabled", servletName);
```
yagee-de: Just throw [UnavailableException](https://jakarta.ee/specifications/platform/10/apidocs/jakarta/servlet/unavailableexception) here to remove this Servlet. Also no `disable` field is needed.

mycore-base/src/main/java/org/mycore/frontend/servlets/MCRServlet.java
```diff
@@ -289,6 +304,10 @@ private static boolean isSessionBoundToRequest(HttpServletRequest req) {
      */
     private void doGetPost(HttpServletRequest req, HttpServletResponse res) throws ServletException, IOException,
         SAXException, TransformerException {
+        if (disabled) {
+            res.sendError(HttpServletResponse.SC_NOT_IMPLEMENTED, "This servlet is disabled.");
+            return;
+        }
```
yagee-de: if init() fails with UnavailableException doGetPost() is never called, so here no check is needed.

yagee-de (changes requested):

mycore-base/src/main/java/org/mycore/frontend/servlets/MCRServlet.java
sebhofmann: Good catch, i tried with the Servlet Context API, but changes to the Map returned by jakarta.servlet.ServletRegistration#getMappings are not reflected by tomcat.
Changed to [UnavailableException](https://jakarta.ee/specifications/platform/10/apidocs/jakarta/servlet/unavailableexception) and tomcat returns 404 now for disabled Servlets.

yagee-de (approved):

PR #1846, opened 2023-05-24
MCR-2879 disable date modified check if object is in import mode

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2879).


bt1cu (approved):
it works

PR #1850, opened 2023-06-01
MCR-2882 Provide component sets in both priority orders

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2882).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/common/config/MCRRuntimeComponentDetector.java
+++ b/mycore-base/src/main/java/org/mycore/common/config/MCRRuntimeComponentDetector.java
@@ -29,6 +29,7 @@ import java.util.Enumeration;
 import java.util.Properties;
 import java.util.SortedSet;
 import java.util.TreeSet;
+import java.util.function.Predicate;
 import java.util.jar.Attributes;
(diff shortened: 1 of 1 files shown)


mycore-base/src/main/java/org/mycore/common/config/MCRComponentOrder.java
```diff
@@ -0,0 +1,27 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with MyCoRe.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.mycore.common.config;
+
+public enum MCRComponentOrder {
```
rsteph-de: Does this have to be a separate enum file? Could we integrate it in MCRRuntimeComponentDetector?

toKrause: I don't have a strong opinion about that.

yagee-de (approved):

PR #1851, opened 2023-06-02
MCR-2883 EventType and ObjectType string representation changed leading to not sent mails

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2883).


rsteph-de (approved):
as bugfix O.K., but we should try to use the enum in MCRMailer implementation

PR #1854, opened 2023-06-06
MCR-2883 MCRNormalizeMCRObjectIDsFilter does not handle derivate ids right

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2883).


rsteph-de (approved):
good fix!

PR #1876, opened 2023-06-13
MCR-2902 MCROCFLXMLMetadataManager#getHighestStoredID is to slow

sebhofmann:
Use MCRMetadataHistoryManager.getHighestStoredID(project, type) instead iterating over filesystem

[Link to jira](https://mycore.atlassian.net/browse/MCR-2902).


bt1cu:
We want to check the running duration with and without OCFL for more than 100.000 objects.

rsteph-de:
With [PR MCR-2902](https://github.com/MyCoRe-Org/mycore/pull/1892) we can now switch the implementation for the MCRObjectIDGenerator. This should be the right position in code to increase the performance. 

The OCFL-Store schould be able to retrieve his own higest ID and should not use other components to do so.
At least on initial initialisation it might by necessary for the store get this information from itself.

Maybe it is possible to implement "short cuts" e.g. for the MyCoRe-ID-based OCFL-Layout to improve performance.

PR #1874, opened 2023-06-13
MCR-2546 consolidate id generation in MCRCreateObjectServlet

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2546).

PR #1879, opened 2023-06-15
MCR-2545 consolidate id generation in RestAPI V2

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2545).


rsteph-de (approved):
looks good for me,

Would it make sense, to handle the defaults for label, version and noNamespaceLocation also in the central MyCoRe code, where the real ID is generated. 
(But this would be another PR)

sebhofmann:
> Would it make sense, to handle the defaults for label, version and noNamespaceLocation also in the central MyCoRe code, where the real ID is generated. (But this would be another PR)

The label handling ist done here #1874

rsteph-de (dismissed):

rsteph-de (approved):

PR #1884, opened 2023-06-20
MCR-2907 MCRNotNullResolver should return original XML source

rsteph-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2907).


rsteph-de:
changes dismissed, the actual behaviour works as designed

PR #1885, opened 2023-06-22
MCR-2840 URN registering not working

eckstueck:
fixed path for filenames that need to be escaped

[Link to jira](https://mycore.atlassian.net/browse/MCR-2840).


yagee-de (approved):
:bow: :+1:

PR #1887, opened 2023-06-26
MCR-2909 separate MCRObjectID generation

rsteph-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2909).


mycore-base/src/main/java/org/mycore/common/MCRUtils.java
```diff
@@ -475,6 +475,9 @@ public static String getSizeFormatted(long bytes) {
         return sizeText + " " + sizeUnit;
     }
 
+    //TODO (Robert) The method compareParts was used only once to compare MCRObjectID objects.
+    // There it was replaced by Comparator.comparing().thenComparing().compareTo()
+    // Can we deprecate or even remove it now?
```
rsteph-de: Is this method used elsewhere (MIR, UBO, ...?)
otherwise I would suggest to delete it.

mycore-base/src/main/java/org/mycore/datamodel/metadata/MCRObjectID.java
```diff
@@ -58,13 +57,10 @@ public final class MCRObjectID implements Comparable<MCRObjectID> {
      */
     public static final int MAX_LENGTH = 64;
 
-    private static final MCRObjectIDFormat ID_FORMAT = new MCRObjectIDDefaultFormat();
+    private static NumberFormat NUMBER_FORMAT = initNumberFormat();
```
rsteph-de, line 60: The NUMBER_FORMAT is now a real static variable and can no longer be changed during run time. That leaded to some failed tests, which I fixed below.
I think the new static version is safer, because there would be side effects if you change the format in a running application.

mycore-base/src/main/java/org/mycore/common/MCRUtils.java
Possommi: I prefer to **deprecate** that method since it is **public**.

sebhofmann (dismissed):

sebhofmann (approved):

PR #1890, opened 2023-06-29
MCR-2912 Not all attributes of MCRUser are persisted in the database

Possommi:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2912).

### DIFF
--- a/mycore-user2/src/main/java/org/mycore/user2/MCRUserManager.java
+++ b/mycore-user2/src/main/java/org/mycore/user2/MCRUserManager.java
@@ -195,9 +195,9 @@ public class MCRUserManager {
         CriteriaQuery<Number> query = cb.createQuery(Number.class);
         Root<MCRUser> users = query.from(MCRUser.class);
         return em.createQuery(
-            query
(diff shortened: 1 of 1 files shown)


mycore-user2/src/main/java/org/mycore/user2/MCRUserManager.java
```diff
@@ -195,9 +195,9 @@ public static boolean exists(String userName, String realm) {
         CriteriaQuery<Number> query = cb.createQuery(Number.class);
         Root<MCRUser> users = query.from(MCRUser.class);
         return em.createQuery(
-            query
-                .select(cb.count(users))
-                .where(getUserRealmCriterion(cb, users, userName, realm)))
+                query
+                    .select(cb.count(users))
+                    .where(getUserRealmCriterion(cb, users, userName, realm)))
             .getSingleResult().intValue() > 0;
```
yagee-de: code formatting… Who is right?

kkrebs:
We tried to reproduce this problem using JUnit tests. Seems that this problem comes from application code.

PR #1894, opened 2023-07-03
MCR-2902 Faster GetHighestStoredID

Hammer1279:
This implements a faster, OCFL Optimized method giving more speed.
It is similar to the way IFS2 resolves it.

Compared to https://github.com/MyCoRe-Org/mycore/pull/1876, it does not fully rely on the HistoryManager but actually checks its storage.

Some performance metrics were collected with the old and current implementation and can be found here:
https://speicherwolke.uni-leipzig.de/index.php/s/WD8BW53Wx69nSWX

[Link to jira](https://mycore.atlassian.net/browse/MCR-2902).

### DIFF
--- a/mycore-ocfl/src/main/java/org/mycore/ocfl/metadata/MCROCFLXMLMetadataManager.java
+++ b/mycore-ocfl/src/main/java/org/mycore/ocfl/metadata/MCROCFLXMLMetadataManager.java
@@ -56,6 +56,8 @@ import org.mycore.datamodel.common.MCRXMLMetadataManagerAdapter;
 import org.mycore.datamodel.ifs2.MCRObjectIDDateImpl;
 import org.mycore.datamodel.metadata.MCRObjectID;
 import org.mycore.datamodel.metadata.history.MCRMetadataHistoryManager;
+import org.mycore.ocfl.repository.MCROCFLHashRepositoryProvider;
+import org.mycore.ocfl.repository.MCROCFLMCRRepositoryProvider;
 import org.mycore.ocfl.layout.MCRStorageLayoutConfig;
 import org.mycore.ocfl.layout.MCRStorageLayoutExtension;
 import org.mycore.ocfl.repository.MCROCFLRepositoryProvider;
@@ -329,24 +331,31 @@ public class MCROCFLXMLMetadataManager implements MCRXMLMetadataManagerAdapter {
 
     @Override
     public int getHighestStoredID(String project, String type) {
-        Path path
-            = Path.of(MCRConfiguration2.getStringOrThrow("MCR.OCFL.Repository." + repositoryKey + ".RepositoryRoot"));
         int highestStoredID = 0;
         int maxDepth = Integer.MAX_VALUE;
-        OcflExtensionConfig config = MCRConfiguration2.getSingleInstanceOf("MCR.OCFL.Repository." + repositoryKey)
-            .map(MCROCFLRepositoryProvider.class::cast).orElseThrow().getExtensionConfig();
+        MCROCFLRepositoryProvider oclfRepoProvider
+            = MCRConfiguration2.getSingleInstanceOf("MCR.OCFL.Repository." + repositoryKey)
+                .map(MCROCFLRepositoryProvider.class::cast).orElseThrow();
+
+        OcflExtensionConfig config = oclfRepoProvider.getExtensionConfig();
+        Path basePath = null;
+
         // optimization for known layouts
         if (Objects.equals(config.getExtensionName(), MCRStorageLayoutExtension.EXTENSION_NAME)) {
             maxDepth = ((MCRStorageLayoutConfig) config).getSlotLayout().split("-").length;
-            path = Path.of(path.toString(), MCROCFLObjectIDPrefixHelper.MCROBJECT.replace(":", ""), project, type);
-            highestStoredID = traverseDirectory(path, maxDepth);
+            basePath = ((MCROCFLMCRRepositoryProvider) oclfRepoProvider).getRepositoryRoot()
+                .resolve(MCROCFLObjectIDPrefixHelper.MCROBJECT.replace(":", ""))
+                .resolve(project).resolve(type);
+            highestStoredID = traverseMCRStorageDirectory(basePath, maxDepth);
         } else if (Objects.equals(config.getExtensionName(),
             HashedNTupleIdEncapsulationLayoutExtension.EXTENSION_NAME)) {
             maxDepth = ((HashedNTupleIdEncapsulationLayoutConfig) config).getNumberOfTuples() + 1;
+            basePath = ((MCROCFLHashRepositoryProvider) oclfRepoProvider).getRepositoryRoot();
         }
-        if (highestStoredID == 0) {
+        
+        if (basePath != null && highestStoredID == 0) {
             Pattern pattern = Pattern.compile("^.*" + project + "_{1}" + type + "_{1}\\d+$");
-            try (Stream<Path> stream = Files.find(path, maxDepth,
+            try (Stream<Path> stream = Files.find(basePath, maxDepth,
                 (filePath, fileAttr) -> pattern.matcher(filePath.getFileName().toString()).matches())) {
                 highestStoredID = stream.map(entry -> entry.getFileName().toString())
                     .map(fileName -> Integer.parseInt(fileName.substring(fileName.lastIndexOf('_') + 1)))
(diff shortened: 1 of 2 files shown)


sebhofmann:
The MCR-Variant could be optimized to search "tuples" Recursive like the [old Store implementation](https://github.com/MyCoRe-Org/mycore/blob/c0a601caa9e1dbfe6748c4258c77187b8bdc6196/mycore-base/src/main/java/org/mycore/datamodel/ifs2/MCRStore.java#L558) does.

Instead of Searching through all files, this Implemenation will only search the highest "tuple" on each level. This will reduce the time drastically, probably from 5055ms for 28989 Objects to ~500ms.

mycore-ocfl/src/main/java/org/mycore/ocfl/metadata/MCROCFLXMLMetadataManager.java
```diff
@@ -317,7 +328,34 @@ public IntStream getStoredIDs(String project, String type) throws MCRPersistence
...
+            .map(MCROCFLRepositoryProvider.class::cast).orElseThrow().getExtensionConfig();
+        // optimization for known layouts
+        if (Objects.equals(config.getExtensionName(), MCRStorageLayoutExtension.EXTENSION_NAME)) {
+            maxDepth = ((MCRStorageLayoutConfig) config).getSlotLayout().split("-").length;
+            path = Path.of(path.toString(), MCROCFLObjectIDPrefixHelper.MCROBJECT.replace(":", ""), project, type);
+        } else if (Objects.equals(config.getExtensionName(),
+            HashedNTupleIdEncapsulationLayoutExtension.EXTENSION_NAME)) {
+            maxDepth = ((HashedNTupleIdEncapsulationLayoutConfig) config).getNumberOfTuples() + 1;
+        }
+        try (Stream<Path> stream
```
rsteph-de: The performance is improve by 10-12% if you work with a precompiled Pattern:
 ```
  Pattern pattern = Pattern.compile("^.*" + TEST_PROJECT + "_{1}" + TEST_TYPE + "_{1}\\d+$");
        try (Stream<Path> stream
            = Files.find(path, maxDepth, (filePath, fileAttr) -> pattern.matcher(filePath.getFileName().toString()).matches())) 
```

Comment (rsteph-de) [reply mycore-ocfl/src/main/java/org/mycore/ocfl/metadata/MCROCFLXMLMetadataManager.java]: use of String.startsWith instead of pattern could bring another 1% improvement:
```
Stream<Path> stream
            = Files.find(path, maxDepth, (filePath, fileAttr) -> filePath.getFileName().toString().startsWith(TEST_PREFIX))
````

File: mycore-ocfl/src/main/java/org/mycore/ocfl/metadata/MCROCFLXMLMetadataManager.java [hunk 2]
```diff
@@ -317,7 +328,34 @@ public IntStream getStoredIDs(String project, String type) throws MCRPersistence
...
+            = Files.find(path, maxDepth, (filePath, fileAttr) -> filePath.getFileName().toString()
+                .matches("^.*" + project + "_{1}" + type + "_{1}\\d+$"))) {
+            highestStoredID = stream.map(entry -> entry.getFileName().toString())
+                .map(fileName -> Integer.parseInt(fileName.substring(fileName.lastIndexOf('_') + 1)))
+                .max(Comparator.comparing(Integer::valueOf)).orElse(0);
```
rsteph-de: This could be simplified to   `.max(Integer::compareTo)`

rsteph-de (changes requested):
some improvements to this algorithm ...

mycore-ocfl/src/main/java/org/mycore/ocfl/metadata/MCROCFLXMLMetadataManager.java
rsteph-de: If we can assume that the number_length of the identifier are equal / do not change in the whole OCFL repository, we could  compare on Path instead of Integer and do the `.max()` earlier in the stream:
```
try (Stream<Path> stream
            = Files.find(path, maxDepth, (filePath, fileAttr) -> filePath.getFileName().toString().startsWith(TEST_PREFIX))) {
            return stream
                .map(p -> p.getFileName())
                .max(Path::compareTo)
                .map(entry -> entry.getFileName().toString())
                .map(smax -> Integer.parseInt(smax.substring(smax.lastIndexOf('_') + 1)))
                .orElse(0);
         }
```
This brings only a very little performance, but may make the algorithm more understandable.

mycore-ocfl/src/main/java/org/mycore/ocfl/metadata/MCROCFLXMLMetadataManager.java
```diff
@@ -317,7 +329,60 @@ public IntStream getStoredIDs(String project, String type) throws MCRPersistence
 
     @Override
     public int getHighestStoredID(String project, String type) {
-        return getStoredIDs(project, type).max().orElse(0);
+        Path path
+            = Path.of(MCRConfiguration2.getStringOrThrow("MCR.OCFL.Repository." + repositoryKey + ".RepositoryRoot"));
+        int highestStoredID = 0;
+        int maxDepth = Integer.MAX_VALUE;
+        OcflExtensionConfig config = MCRConfiguration2.getSingleInstanceOf("MCR.OCFL.Repository." + repositoryKey)
```
yagee-de: Will not work with S3

Hammer1279: The default Repository Providers don't allow the use of S3, since they are all `fileSystem` storage.
```java
this.repository = new OcflRepositoryBuilder()
            .defaultLayoutConfig(getExtensionConfig())
            .storage(storage -> storage.fileSystem(repositoryRoot))
            .workDir(workDir)
            .build();
```
`storage.cloud([...])` would be required here.
See: [ocfl-java Amazon-S3 Usage](https://github.com/OCFL/ocfl-java/blob/main/docs/USAGE.md#amazon-s3)
When creating an S3 RepositoryProvider, the MetadataManagers method can also be replaced for the S3 store with something like `MCROCFLS3XMLMetadataManager`.

mycore-ocfl/src/main/java/org/mycore/ocfl/metadata/MCROCFLXMLMetadataManager.java
```diff
@@ -317,7 +329,58 @@ public IntStream getStoredIDs(String project, String type) throws MCRPersistence
...
+            } catch (Exception e) {
+                // nothing found, let the HistoryManager take over
+            }
+        }
+        return Math.max(highestStoredID, MCRMetadataHistoryManager.getHighestStoredID(project, type)
+            .map(MCRObjectID::getNumberAsInteger)
+            .orElse(0));
+    }
+
+    private int traverseDirectory(Path path, int depth) {
```
rsteph-de: Please provide a short comment, what this method does.
For people how do not understand the regex so quickly

mycore-ocfl/src/main/java/org/mycore/ocfl/metadata/MCROCFLXMLMetadataManager.java
rsteph-de: So we ignore the "S3-case" until we start with an S3 implementation in the future.

mycore-ocfl/src/main/java/org/mycore/ocfl/metadata/MCROCFLXMLMetadataManager.java
```diff
@@ -317,7 +331,99 @@ public IntStream getStoredIDs(String project, String type) throws MCRPersistence
...
+            HashedNTupleIdEncapsulationLayoutExtension.EXTENSION_NAME)) {
+            maxDepth = ((HashedNTupleIdEncapsulationLayoutConfig) config).getNumberOfTuples() + 1;
+            basePath = ((MCROCFLHashRepositoryProvider) oclfRepoProvider).getRepositoryRoot();
+        } else {
+            //for other repository provider implementation start with root directory
+            basePath = MCRConfiguration2.getString("MCR.OCFL.Repository." + repositoryKey + ".RepositoryRoot")
+                .map(Path::of).orElse(null);
+        }
+
+        if (basePath != null && highestStoredID == 0) {
```
Hammer1279, line 360: After talking with robert about it, we found it, that it _might_ be possible that `traverseMCRStorageDirectory` _could_ return `-1`, so in that case this would then also run into the fallback:
```suggestion
        if (basePath != null && highestStoredID <= 0) {
```

rsteph-de (approved):
looks good for me now

Hammer1279:
Some idea I just had that might also improve speed thanks to less I/O operations might be to create an (OCFL) file index (+ more info if wanted, max id used for example), this would also speed up the hashed layout significantly.
I'll probalby give this one a try.

PR #1896, opened 2023-07-10
MCR-2914 sass does not work with jetty

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2914).


kkrebs (approved):
successfully tested with MIR

PR #1903, opened 2023-07-12
MCR-2918 new MCRSolrClassificationEventHandler

rsteph-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2918).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/datamodel/classifications2/MCRCategoryDAO.java
+++ b/mycore-base/src/main/java/org/mycore/datamodel/classifications2/MCRCategoryDAO.java
@@ -32,6 +32,18 @@ import java.util.SortedSet;
 public interface MCRCategoryDAO {
 
     /**
+     * key for update type as payload in MCREvent
+     */
+    public static final String UPDATE_TYPE_KEY = "type";
+
+    /**
+     * different update types, used in MCRSolrClassificationEventHandler
+     */
+    public enum UpdateType {
+        MOVE, REPLACE
+    };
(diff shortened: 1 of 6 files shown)


Mewel (changes requested):
- [x] MCRSolrClassificationEventHandler#solrMove() and MCRSolrClassificationEventHandler#solrDelete() are just copied from MCRSolrCategoryDAO.
Put them into MCRSolrClassificationUtil.

- [x] Also solrMove() doesn't require the index position anymore. So the whole position thing can be deleted:
                    int pos = -1;
                    if (evt.get("index") instanceof Integer) {
                        pos = (Integer) evt.get("index");
                    }
Including passing it through the MCREvent.

- [x] The switch statement should be an extended switch:
switch (evt.getEventType()) {
            case CREATE -> MCRSolrClassificationUtil.reindex(categ, categParent);
            case UPDATE -> {
...

- [x] Also the whole switch inside of a switch looks very ugly. The inside switch should be extracted to a method.

- [x] MCRSolrClassificationUtil.reindex(replaced.toArray(new MCRCategory[replaced.size()])); can be replaced by MCRSolrClassificationUtil.reindex(replaced.toArray(new MCRCategory[0]));

- [x] For the event types (and the type itself) you use strings like "move" and "replace", maybe use enum's instead.

rsteph-de:
If everything works properly, I plan to remove the MCRSolrCategoryDAO in the next release.

Mewel (approved):

PR #1930, opened 2023-07-31
fix dependencies (fixes IDE errors)

yagee-de:
Should fix complaints in IDE (at least in IntelliJ).
Adds an optional orcid-model dependency with scope provided.


golsch (approved):
My IDE (Eclipse) is happy and module works.

PR #1934, opened 2023-08-01
MCR-2938 rewrite jackson.core.JsonFactory with ObjectNode

rsteph-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2938).

### DIFF
--- a/mycore-base/pom.xml
+++ b/mycore-base/pom.xml
@@ -167,10 +167,6 @@
     </dependency>
     <dependency>
       <groupId>com.fasterxml.jackson.core</groupId>
-      <artifactId>jackson-core</artifactId>
-    </dependency>
-    <dependency>
-      <groupId>com.fasterxml.jackson.core</groupId>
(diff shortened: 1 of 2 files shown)


yagee-de (changes requested):
We should drop JSON libraries that are not in use out there. Jackson is the JSON library out there. I don't like a shift towards GSON if it's not discussed in a broader group and our goal to drop Jackson as a dependency - which I think it's imposible.

rsteph-de:
I totally agree that we should consolidate our use of JSON libraries
This was the only place where a jackson `JSONGenerator` was used. 
Currently you can find more GSON code in MyCoRe like I used it in the replacement.
And as you can see in the failed test build, it looks like we can remove the dependency to `com.fasterxml.jackson.core:jackson-core:jar:2.14.2:compile`in `mycore base`

yagee-de (approved):

PR #1935, opened 2023-08-02
MCR-2939 improve MCRUserAgentFilter

yagee-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2939).

### DIFF
--- a/mycore-base/src/main/resources/config/mycore.properties
+++ b/mycore-base/src/main/resources/config/mycore.properties
@@ -36,6 +36,14 @@
 # MCR.XMLParser.ValidateSchema=false
 
 ##############################################################################
+# Servlet Filter configuration
+##############################################################################
+
+MCR.Filter.UserAgent.Pattern=(bot|spider|crawler|mercator|slurp|seek|nagios|Java|curl|wget|fetch)
+MCR.Filter.UserAgent.MinLength=4
+MCR.Filter.UserAgent.AcceptInvalid=false
+
+##############################################################################
 # The MCRURIResolver configuration
 ##############################################################################
 
--- a/mycore-base/src/main/resources/config/mycore.properties
+++ b/mycore-base/src/main/resources/config/mycore.properties
@@ -39,7 +39,7 @@
 # Servlet Filter configuration
 ##############################################################################
 
-MCR.Filter.UserAgent.Pattern=(bot|spider|crawler|mercator|slurp|seek|nagios|Java|curl|wget|fetch)
+MCR.Filter.UserAgent.BotPattern=(bot|spider|crawler|mercator|slurp|seek|nagios|Java|curl|wget|fetch)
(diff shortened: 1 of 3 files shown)


mycore-base/src/main/resources/config/mycore.properties
```diff
@@ -35,6 +35,14 @@
   MCR.XMLParser.ValidateSchema=true
 # MCR.XMLParser.ValidateSchema=false
 
+##############################################################################
+# Servlet Filter configuration
+##############################################################################
+
+MCR.Filter.UserAgent.Pattern=(bot|spider|crawler|mercator|slurp|seek|nagios|Java|curl|wget|fetch)
```
rsteph-de: ` .PatternInvalid` ? oder `.Invalid.Pattern`

toKrause: Maybe `MCR.Filter.UserAgent.Pattern.Bot` or more generally  `MCR.Filter.UserAgent.Pattern.Sessionless`.

Something like `MCR.Filter.UserAgent.Pattern.Invalid` would be bad, because, according to class comment and, indeed, the code, matching agents are actually valid. A match just results in a closed session, but the request is served anyways.

This confused me too at first, but the code basically does the following:
```
if (userAgent == null || isInvaild(userAgent)) {
   if (!ACCEPT_INVALID_USER_AGENTS) {
      return rejectRequest();
   }
}

if (isInvaild(userAgent) || isBot(userAgent)) {
   closeSession();
}

return acceptReuest();
```

yagee-de: renamed to `MCR.Filter.UserAgent.BotPattern` in https://github.com/MyCoRe-Org/mycore/pull/1935/commits/bb9ca9654f7c2c16eb5e7a8189e5ed294ba214e5

toKrause:
Just out of interest, what is the reason to consider short user agents to be invalid?

eckstueck:
Can you make the BotPattern not case sensitiv. To recognize User-Agents like "DAMFileCheckCrawler" or bots like "SemrushBot"

mycore-base/src/main/resources/config/mycore.properties
toKrause: see https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html#CASE_INSENSITIVE

toKrause:
> Can you make the BotPattern not case sensitiv. To recognize User-Agents like "DAMFileCheckCrawler" or bots like "SemrushBot"

`MCR.Filter.UserAgent.Pattern=(?i)(bot|spider|crawler|mercator|slurp|seek|nagios|java|curl|wget|fetch)` should do

see https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html#CASE_INSENSITIVE

toKrause (approved):

PR #1942, opened 2023-08-10
MCR-2949 make method in MCRMetaEnrichedLinkID public

rsteph-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2949).


rsteph-de:
For me it would further help if the static constants of the class become public, but that's only "nice to have"

rsteph-de:
seems to be unecessary if we modify the API to work with MCREditableMetaEnrichedLinkID

rsteph-de:
solved by improving the API

PR #1947, opened 2023-08-18
MCR-2954 Improve logging of cron schedule

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2954).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/common/log/MCRTableMessage.java
+++ b/mycore-base/src/main/java/org/mycore/common/log/MCRTableMessage.java
@@ -1,0 +1,163 @@
+/*
+ * This file is part of ***  M y C o R e  ***
+ * See http://www.mycore.de/ for details.
+ *
+ * MyCoRe is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * MyCoRe is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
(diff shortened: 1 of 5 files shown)


mycore-base/src/main/java/org/mycore/common/log/MCRTableMessage.java
```diff
@@ -0,0 +1,145 @@
+package org.mycore.common.log;
```
sebhofmann, line 19: Missing copyright

mycore-cronjob/src/main/java/org/mycore/mcr/cronjob/MCRCronjobManager.java
```diff
@@ -119,11 +135,8 @@ private void scheduleNext(String jobID) {
         });
     }
 
-    public <T extends MCRCronjob> T getJob(String id) {
```
sebhofmann, line 122: Did you check if this method is used in MIR or other mycore projects?

sebhofmann (changes requested):

mycore-cronjob/src/main/java/org/mycore/mcr/cronjob/MCRCronjobManager.java
toKrause: AFAIA it is not used. TBH, I didn't notice that this method is public, as I don't see its value in the public API. Should I keep it anyways?

sebhofmann (approved):

mycore-cronjob/src/main/java/org/mycore/mcr/cronjob/MCRCronjob.java
```diff
@@ -139,6 +141,16 @@ public Cron getCron() {
         return cron;
     }
 
+    @MCRPostConstruction
```
sebhofmann, line 144: Maybe consider using id or replace id. Since they are almost the same. (id is only the last part of the property)

PR #1948, opened 2023-08-18
MCR-2955 Friends are outdated and should be able to be wiped out

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2955).

### DIFF
--- a/mycore-oai/src/main/java/org/mycore/oai/MCROAIIdentify.java
+++ b/mycore-oai/src/main/java/org/mycore/oai/MCROAIIdentify.java
@@ -20,6 +20,7 @@ package org.mycore.oai;
 import java.time.Instant;
 import java.util.Collection;
 import java.util.Map;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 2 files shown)


rsteph-de:
Can we extend this issue and add a filter for empty Friends-Properties here:
https://github.com/MyCoRe-Org/mycore/blob/1ca2311523ec98cadeb4a1733dc678032d820e11/mycore-oai/src/main/java/org/mycore/oai/MCROAIIdentify.java#L115

```java
.filter(Predicate.not(String::isBlank))
````
This would allow to override settings with empty properties:
```
MCR.OAIDataProvider.OAI2.Friends.DBThueringen=
```

toKrause:
> Can we extend this issue and add a filter for empty Friends-Properties

That is a good idea, but should probably be a separate issue.

rsteph-de:
Since it's only one additional line of code I would rephrase the issue title
"improve configuration of OAI friends"

rsteph-de (approved):

PR #1950, opened 2023-08-24
MCR-2958 simplify sass import

yagee-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2958).


mycore-sass/src/main/java/org/mycore/sass/MCRResourceImporter.java
```diff
@@ -0,0 +1,140 @@
...
+            })
+            .or(() -> Optional.ofNullable(MCRConfigurationDir.getConfigResource("META-INF/resources/" + resourcePath)))
+            .orElseGet(() -> {
+                try {
+                    return servletContext.getResource("/" + resourcePath);
+                } catch (MalformedURLException e) {
+                    throw new UncheckedIOException(e);
+                }
+            });
+    }
```
yagee-de, line 146: needs to be replaced by a central resource locating method. @toKrause anything ready yet?

toKrause: Draft PR: https://github.com/MyCoRe-Org/mycore/pull/1849

toKrause: This would be replaced by `MCRResourceHelper#getWebResourceUrl(resourcePath)`. I will include that change in my PR  once this PR is merged and I rebase my PR.

mycore-sass/src/main/java/org/mycore/sass/MCRResourceImporter.java
```diff
@@ -0,0 +1,142 @@
+package org.mycore.sass;
```
toKrause, line 19: License header is missing.

mycore-sass/src/main/java/org/mycore/sass/MCRSassCompiler.java
```diff
@@ -1,13 +1,9 @@
 package org.mycore.sass;
```
toKrause, line 19: License header is also missing; and has been missing for a while.

mycore-sass/src/main/java/org/mycore/sass/MCRResourceImporter.java
```diff
@@ -0,0 +1,142 @@
...
+                } catch (MalformedURLException e) {
+                    throw new UncheckedIOException(e);
+                }
+            });
+    }
+
+    /**
+     * Helper class for working with URLs.
+     */
+    private static class UrlHelper extends CustomUrlImporter {
```
toKrause, line 153: Could use a better name, i.e. `{$something}Importer` where ${something} describes the purpose of this class. The `Helper`-suffix is, as per PMD-configuration, usually used for utility classes (i.e. a class that has only static methods and doesn't need to be/shouldn't be instantiated).

Alternatively, this could be implemented as an actual utility class that hides the sadly necessary weirdness of using an instance-level utility method instance-independently. Something like

```
    /**
     * Helper class that exposes {@link CustomizedUrlImporter#isFile(URL)} as a static method.
     */
    private static final class UrlHelper {
        
        private static final CustomizedUrlImporter IMPORTER = new CustomizedUrlImporter();
        
        private UrlHelper(){
        }

        public static boolean isFile(URL url) throws IOException {
            return IMPORTER.isFile(url);
        }

        private static final class CustomizedUrlImporter extends CustomUrlImporter {

            @Nullable
            @Override
            public URL canonicalizeUrl(String url) {
                return null;
            }


            @Override
            public boolean isFile(URL url) throws IOException {
                return super.isFile(url);
            }

        }
        
    }
```

mycore-sass/src/main/java/org/mycore/sass/MCRResourceImporter.java
```diff
@@ -0,0 +1,142 @@
...
+            //with jetty the directory does not end with /, so we need to REAL check
+            return null;
+        }
+        LOGGER.debug("Resolved {} to {}", url, resource);
+        return resource == null ? null : url;
+    }
+
+    /**
+     * Handles the import of a URL and returns the success of the import along with its content.
+     *
```
toKrause: missing `<br>` or `<p>`

mycore-sass/src/main/java/org/mycore/sass/MCRResourceImporter.java
```diff
@@ -0,0 +1,142 @@
...
+    @Override
+    public ImportSuccess handleImport(String url) throws IOException {
+        LOGGER.debug(() -> "handle import: " + url);
+        if (!url.startsWith(SASS_URL_PREFIX)) {
+            return null;
+        }
+        ImportSuccess.Builder result = ImportSuccess.newBuilder();
+        String resourcePath = url.substring(SASS_URL_PREFIX.length());
+        LOGGER.debug(() -> "resource: " + resourcePath);
+        URL configResource = getWebResourceUrl(resourcePath);
```
toKrause: Seems to me like 

```java
        if (!url.startsWith(SASS_URL_PREFIX)) {
            return null;
        }
        String resourcePath = url.substring(SASS_URL_PREFIX.length());
        LOGGER.debug(() -> "resource: " + resourcePath);
        URL configResource = getWebResourceUrl(resourcePath);
        LOGGER.debug(() -> "resource url: " + configResource);
```
(and possible the special handling of Jettys behaviors) should be a common method used by `canonicalize` as well as `handleImport`. 

The necessity for the following differences isn't immediately obvious and I'd guess that there actually is no such necessity.

- Whats the significance of naming the result of `getWebResourceUrl` differently in the two methods. Is there actually a difference?
- Why is the logging only relevant in `handleImport`?
- Why is the special handling of Jettys behaviors not present in `handleImport`? Is there a difference or is this an oversight?

mycore-sass/src/main/java/org/mycore/sass/MCRSassCompiler.java
```diff
@@ -169,33 +98,28 @@ protected EmbeddedSass.InboundMessage.CompileRequest.Builder compileRequestBuild
...
-                .setImporter(EmbeddedSass.InboundMessage.CompileRequest.Importer.newBuilder()
-                    .setImporterId(importer.getId())
-                    .build())
-                .setSyntax(syntax)
-                .build();
+    public CompileSuccess compile(String realFileName, OutputStyle outputStyle)
+        throws SassCompilationFailedException, IOException {
+        String ourURL = MCRResourceImporter.SASS_URL_PREFIX + realFileName;
+        MCRResourceImporter importer = new MCRResourceImporter(servletContext);
+        var importSuccess = importer.handleImport(ourURL);
```
toKrause: I'm not a fan of this arbitrary use of `var` in a long list of otherwise explicitly typed variables.

toKrause (changes requested):

mycore-sass/src/main/java/org/mycore/sass/MCRResourceImporter.java
yagee-de: ok

mycore-sass/src/main/java/org/mycore/sass/MCRSassCompiler.java
yagee-de: ok

mycore-sass/src/main/java/org/mycore/sass/MCRResourceImporter.java
yagee-de: ok

mycore-sass/src/main/java/org/mycore/sass/MCRSassCompiler.java
yagee-de: ok

mycore-sass/src/main/java/org/mycore/sass/MCRResourceImporter.java
yagee-de: have extracted common code into new method
Jetty behavior is only relevant in canonicalize, as this will result in null so there is no handleImport call afterwards.

mycore-sass/src/main/java/org/mycore/sass/MCRResourceImporter.java
yagee-de: I prefer to keep it small and simple, it's an internal class only

toKrause: The point of the `URLHelper`  is to expose and use a protected instance method where it is not intended to be used. This is both unusual and subtle and therefore hard to understand at a glance. Both points require some remedy to make easier for the reader. "Being internal" is not a good reason to omit this. That `URLHelper` actually is a `CustomUrlImporter` makes things worse still. At least a better name and some explanatory comment is necessary here.

toKrause (approved):

PR #1954, opened 2023-08-30
 MCR-2827 MCRConcurrentHashMap.get wrong behavior

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2827).

### DIFF
--- a/mycore-base/src/test/java/org/mycore/common/config/MCRConfigurationTest.java
+++ b/mycore-base/src/test/java/org/mycore/common/config/MCRConfigurationTest.java
@@ -18,21 +18,26 @@
 
 package org.mycore.common.config;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertTrue;
-
 import java.util.Arrays;
 
+import org.apache.commons.lang3.tuple.Pair;
 import org.apache.logging.log4j.LogManager;
 import org.junit.Test;
 import org.mycore.common.MCRTestCase;
+import org.mycore.common.config.MCRConfiguration2.SingletonKey;
+import org.mycore.common.config.singletoncollision.MCRConfigurationSingletonCollisionClassA;
+import org.mycore.common.config.singletoncollision.MCRConfigurationSingletonCollisionClassB;
+
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
 
 /**
  * @author Thomas Scheffler (yagee)
  * @author Tobias Lenhardt [Hammer1279]
  */
 public class MCRConfigurationTest extends MCRTestCase {
+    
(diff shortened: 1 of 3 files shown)


mycore-base/src/test/java/org/mycore/common/config/MCRConfigurationTest.java
```diff
@@ -18,131 +18,156 @@
...
+
+    private Pair<String, String> createConflictConfiguration() {
+
         char[] letters = "abcdefghijklmnopqrstuvwxyz".toCharArray();
 
         int limit = 50; // initial size of char arrays
 
-        char[][] className = new char[2][limit];
-        char[][] property = new char[2][limit];
+        String classNameA = MCRConfigurationSingletonCollisionClassA.class.getName();;
```
Hammer1279:
```suggestion
        String classNameA = MCRConfigurationSingletonCollisionClassA.class.getName();
```

mycore-base/src/test/java/org/mycore/common/config/MCRConfigurationTest.java
```diff
@@ -18,131 +18,156 @@
...
-            if (value.matches(".+MCRConfigurationSingletonCollisionClass[AB]")) {
-                assertNotNull(MCRConfiguration2.instanceHolder.get(new MCRConfiguration2.SingletonKey(key, value)));
-            }
-        });
+    private int spread(int h) {
+        return (h ^ (h >>> 16)) & 0x7fffffff /* all usable bits */;
     }
+
+    private static String toSuitableString(char[] value) {
+        return String.valueOf(value).trim().replaceAll(Pattern.quote(" "), "");
```
Hammer1279: Is the `Pattern.quote()` really necessary?

toKrause: I think it's best practice to always use `Pattern.quote` when using a string literal as a pattern.

THB, I don't really see why the the `replaceAll` is necessary at all, but it's in the original code. If necessary, a `replaceAll("\\s+", "")` would probably be best?

mycore-base/src/test/java/org/mycore/common/config/singletoncollision/MCRConfigurationSingletonCollisionClassA.java
```diff
@@ -1,36 +1,34 @@
...
+ * (at your option) any later version.
+ *
+ * MyCoRe is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with MyCoRe.  If not, see <http://www.gnu.org/licenses/>.
+ */
```
Hammer1279, line 17: Note for later: this has to also be done on class B

Hammer1279 (approved):
Looks good to me.

PR #1971, opened 2023-10-05
MCR-2978 Added org.mycore.services.i18n.MCRTranslation.translate(java…

Possommi:
….lang.String, java.lang.String, java.lang.String) to MCRTranslation

[Link to jira](https://mycore.atlassian.net/browse/MCR-2978).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/services/i18n/MCRTranslation.java
+++ b/mycore-base/src/main/java/org/mycore/services/i18n/MCRTranslation.java
@@ -133,7 +133,7 @@ public class MCRTranslation {
      *            target locale of translation
      * @return translated String
      */
-    public static String translate(String label, Locale locale) {
+    public static String translateToLocale(String label, Locale locale) {
         return translate(label, locale, MESSAGES_BUNDLE);
     }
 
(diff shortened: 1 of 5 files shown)


mycore-base/src/main/java/org/mycore/services/i18n/MCRTranslation.java
```diff
@@ -285,6 +285,22 @@ public static String translate(String label, String argument, Locale locale) {
...
+     * @param argument
+     *            String that is inserted instead of placeholders in the property values
+     * @param locale target locale of translation
+     * @return translated String
+     * @see #translate(String, Object[])
+     */
+    public static String translate(String label, String argument, String locale) {
+        return translate(label, argument, MCRTranslation.getLocale(locale));
+    }
+
```
yagee-de: How can this method be distinguished from `public static String translate(String label, Object... arguments)`? It will probably break old code that expect the older method to be called.

Possommi: That's true. Unfortunately it is currently not possible to invoke MCRTranslation#translate() from xsl with a locale other than CurrentLang without writing a delegate method.

sebhofmann: Could rename the method to translateToLanguage

rsteph-de: We've discussed the issue, the new methods should be name translateToLocale(....)
Please do also rename the new methods from [MCR-2970](https://github.com/MyCoRe-Org/mycore/pull/1965/files)

Possommi: @rsteph-de Renamed methods to ```translateToLocale(...)```as requested.

kkrebs (approved):

PR #1972, opened 2023-10-05
MCR-2979 fact acl system interprets webpage targets as metadata

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2979).


rsteph-de (approved):
looks good for me / needs to get into release

PR #1982, opened 2023-10-18
MCR-2985 validate store keys

yagee-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2985).


yagee-de:
Does not seem to solve my problem

PR #1984, opened 2023-10-19
MCR-2986 fix classeditor dnd bug

Mewel:
-add a childloader to disallow parallel loading
-categories will be only loaded once
-the dnd drop event will wait till all children are successfully loaded

[Link to jira](https://mycore.atlassian.net/browse/MCR-2986).


sebhofmann:
What about the transaction issues on the server side? The transaction was committed even if a error happened.

Mewel:
There will be another Ticket/PR for that. This is just for the classeditor.

yagee-de (approved):

PR #1986, opened 2023-10-25
MCR-2988 Add sourceUri validation in xEditor

Possommi:
Added property `MCR.XEditor.MCRXEditorTransformer.URISchema.disallowList=callJava`:

[Link to JIRA](https://mycore.atlassian.net/browse/MCR-2988).

### DIFF
--- a/mycore-xeditor/src/main/java/org/mycore/frontend/xeditor/MCRXEditorTransformer.java
+++ b/mycore-xeditor/src/main/java/org/mycore/frontend/xeditor/MCRXEditorTransformer.java
@@ -20,7 +20,9 @@ package org.mycore.frontend.xeditor;
 
 import java.io.IOException;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 import javax.xml.parsers.DocumentBuilderFactory;
 import javax.xml.parsers.ParserConfigurationException;
@@ -41,6 +43,7 @@ import org.jdom2.Parent;
 import org.mycore.common.MCRClassTools;
 import org.mycore.common.MCRConstants;
 import org.mycore.common.MCRException;
+import org.mycore.common.config.MCRConfiguration2;
 import org.mycore.common.content.MCRContent;
 import org.mycore.common.content.MCRWrappedContent;
 import org.mycore.common.content.transformer.MCRContentTransformer;
@@ -76,6 +79,10 @@ public class MCRXEditorTransformer {
 
     private boolean withinSelectMultiple = false;
 
+    private static final List<String> DISALLOWED_URI_SCHEMAS = MCRConfiguration2.getOrThrow(
+        "MCR.XEditor.MCRXEditorTransformer.URISchema.disallowList",
+        MCRConfiguration2::splitValue).collect(Collectors.toList());
+
     public MCRXEditorTransformer(MCREditorSession editorSession, MCRParameterCollector transformationParameters) {
         this.editorSession = editorSession;
         this.transformationParameters = transformationParameters;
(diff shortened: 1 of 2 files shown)


mycore-xeditor/src/main/java/org/mycore/frontend/xeditor/MCRXEditorTransformer.java
```diff
@@ -107,9 +114,25 @@ public void addNamespace(String prefix, String uri) {
...
+        if (!validateURI(uriRe)) {
+            throw new TransformerException("URI '" + uriRe + "' contains illegal uri scheme");
+        }
+
         editorSession.setEditedXML(uri);
     }
 
+    protected boolean validateURI(String uri) {
+        for (String scheme : MCRXEditorTransformer.DISALLOWED_URI_SCHEMAS) {
+            if (uri.contains(scheme)) {
```
rsteph-de, line 128: startsWith()?

mycore-xeditor/src/main/resources/components/xeditor/config/mycore.properties
```diff
@@ -44,6 +44,9 @@ MCR.ContentTransformer.xeditor.Class=org.mycore.common.content.transformer.MCRXS
 MCR.ContentTransformer.xeditor.TransformerFactoryClass=org.apache.xalan.processor.TransformerFactoryImpl
 MCR.ContentTransformer.xeditor.Stylesheet=xsl/xeditor-templates.xsl,xsl/xeditor.xsl
 
+# Comma separated list of uri resolvers disallowed in the xed:source/@uri
+MCR.XEditor.MCRXEditorTransformer.URISchema.disallowList=callJava:
+
```
rsteph-de, line 49: with or without colon - I am not sure
If we keep the colon,  the value works like an URI prefix and we could define more complex URIs including server with server and pathes if needed
If we remove the colon, the value is a "proper" URI schema and we should at the colon in the validating Java code.

mycore-xeditor/src/main/java/org/mycore/frontend/xeditor/MCRXEditorTransformer.java
Possommi: Does not catch xslStyle:xeditor-preprocessor:callJava:org.mycore.common.xml.MCRXMLFunctions:foo:bar

fluetze (changes requested):
I will not approve this. It is not the responsibility of internal XEditor code to check URIs to be "allowed" or not. URIs are URIs, there can not be any assumption about their meanings or intended usage here. Setting allowed/disallowed patterns for the complete application will strictly prevent useful URIs even for admin use, for debugging etc. 

The misuse of URIs must be fixed by the application that uses the frameworks. There simply must be no line of code like xed:form/@uri="$foo" in the application. Instead, the application should only use parameters like in xed:form/@uri="mcrobject:{$id}" - and the mcrobject uri resolver will, hopefully, defend against nonsense object IDs.

Possommi:
> I will not approve this. It is not the responsibility of internal XEditor code to check URIs to be "allowed" or not. URIs are URIs, there can not be any assumption about their meanings or intended usage here. Setting allowed/disallowed patterns for the complete application will strictly prevent useful URIs even for admin use, for debugging etc.
> 
> The misuse of URIs must be fixed by the application that uses the frameworks. There simply must be no line of code like xed:form/@uri="$foo" in the application. Instead, the application should only use parameters like in xed:form/@uri="mcrobject:{$id}" - and the mcrobject uri resolver will, hopefully, defend against nonsense object IDs.

It is true that one could check for a certain uri scheme and I could do that for applications under my control. But there are mycore based apps out there where their developers are not aware of this potential security issue. This fix would prevent at least the invocation of the callJava resolver.

fluetze:
> It is true that one could check for a certain uri scheme and I could do that for applications under my control. But there are mycore based apps out there where their developers are not aware of this potential security issue. This fix would prevent at least the invocation of the callJava resolver.

The fix will only become effective it the application admin actually installs it. 
If he/she can do so, he/she could also fix the forms that contain the dangeroues line of code.
If mycore/mir/ubo use complete sourceURIs in provided code, that forms have to be fixed.

Do we also "fix" that in any other MyCoRe code that uses URIs? 
How do we know there isn't any application that calls that code?
What about custom URI resolvers? We should also check every XSL file.

rsteph-de:
> Do we also "fix" that in any other MyCoRe code that uses URIs?
> How do we know there isn't any application that calls that code?
> What about custom URI resolvers? We should also check every XSL file.

We should at least try to identify the places where URIs are generated by user-input, e.g. HTML form input, request params, ...

kkrebs:
is no xeditor problem, should not be fixed here, fix in application instead

PR #1987, opened 2023-10-25
MCR-2989 rollback on jersey exception

Mewel:
-add setRollbackOnly support
-jersey exception mapper uses setRollbackOnly
-add test cases
-improve jpa test api

[Link to jira](https://mycore.atlassian.net/browse/MCR-2989).


mycore-ocfl/src/main/java/org/mycore/ocfl/MCROCFLPersistenceTransaction.java
```diff
@@ -156,6 +156,15 @@ public boolean getRollbackOnly() {
         return rollbackOnly;
     }
 
+    @Override
+    public void setRollbackOnly() throws IllegalStateException {
+        LOGGER.debug("TRANSACTION {} SET ROLLBACK - {}", threadId, rollbackOnly);
+        if (!isActive()) {
+            throw new IllegalStateException("TRANSACTION NOT ACTIVE");
+        }
+        this.rollbackOnly = true;
```
Hammer1279: I don't think we need to explicitly tell it to use the `this` context here, right?
It wasn't used in the other methods, so I'm mentioning it for uniformity.
```suggestion
        rollbackOnly = true;
```

yagee-de (approved):

PR #1988, opened 2023-10-27
MCR-2990 Added support of <xinclude @uri="..."/> element.

Possommi:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2990).

### DIFF
--- a/mycore-base/src/main/resources/xsl/MyCoReWebPage.xsl
+++ b/mycore-base/src/main/resources/xsl/MyCoReWebPage.xsl
@@ -139,4 +139,9 @@
 
   <!-- =============================================================================== -->
 
+  <!-- Dynamic includes -->
+
+  <xsl:template match="xinclude">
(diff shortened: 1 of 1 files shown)


fluetze:
Maybe we should improve the implementation and use/introduce a namespace for xinclude here.

fluetze (changes requested):
Maybe we should improve the implementation and use/introduce a namespace for xinclude here?

Possommi:
> Maybe we should improve the implementation and use/introduce a namespace for xinclude here?

like ```mcrwebpage:xinclude``` but I think it is not needed since other elements do not have namespaces either

fluetze:
> > Maybe we should improve the implementation and use/introduce a namespace for xinclude here?
> 
> like `mcrwebpage:xinclude` but I think it is not needed since other elements do not have namespaces either

Would help distinguish it from the "official" xinclude, and from the (x)html namespace.

fluetze (approved):

PR #1995, opened 2023-11-08
MCR-2995 always return Source with systemId

yagee-de:
uses original systemId if an URIResolver does not provide one on return

[Link to jira](https://mycore.atlassian.net/browse/MCR-2995).


rsteph-de (approved):
looks good for me

PR #2006, opened 2023-11-21
MCR-3002 disable log4j2 javadoc url

rsteph-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3002).


yagee-de (commented):
What about https://github.com/MyCoRe-Org/mycore/pull/2008?

PR #2013, opened 2023-11-23
MCR-3005 remove test scope for com.zaxxer:HikariCP in mycore-bom

rsteph-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3005).

### DIFF
--- a/mycore-bom/pom.xml
+++ b/mycore-bom/pom.xml
@@ -186,6 +186,11 @@
         <version>2.1.0</version>
       </dependency>
       <dependency>
+        <groupId>com.zaxxer</groupId>
+        <artifactId>HikariCP</artifactId>
+        <version>5.0.1</version>
+      </dependency>
+      <dependency>
         <groupId>commons-beanutils</groupId>
         <artifactId>commons-beanutils</artifactId>
         <version>1.9.4</version>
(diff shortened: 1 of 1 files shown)


yagee-de (changes requested):
I think the test scope is OK. HikariCP is not required and can be used like H2 in your app, just set the scope to "compile" there.

rsteph-de:
> I think the test scope is OK. ... just set the scope to "compile" there.

Just tried it - does not work for me.
I added the dependency in the myapp-common Maven-module (with scope=compile) to be available in myapp-webapp and myapp-cli. But somehow it is not copied into the webapp's  WAR file or the cli's lib directory.

rsteph-de:
> HikariCP is not required and can be used like H2 in your app

Since C3P0 is not working with the Jakarta configuration, HikariCP is the new connection pooling for all of us. The definition via `dependencyManagement does not say whether the library is finally used or not.

kkrebs (approved):
We had the same problem with mycore-classification, see https://mycore.atlassian.net/browse/MCR-2905

only removing test scope helps here

yagee-de:
OK

yagee-de (approved):

PR #2021, opened 2023-11-27
MCR-2978 fix MCRLocaleResource and add Junit tests

yagee-de:
also adding warning log msg if a Locale instance is and argument to translate()

[Link to jira](https://mycore.atlassian.net/browse/MCR-2978).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/services/i18n/MCRTranslation.java
+++ b/mycore-base/src/main/java/org/mycore/services/i18n/MCRTranslation.java
@@ -238,6 +238,9 @@ public class MCRTranslation {
      * @return translated String
      */
     public static String translate(String label, Object... arguments) {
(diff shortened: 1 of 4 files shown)


mycore-base/src/main/java/org/mycore/services/i18n/MCRTranslation.java
```diff
@@ -238,6 +238,9 @@ public static Map<String, String> translatePrefixToLocale(String prefix, Locale
      * @return translated String
      */
     public static String translate(String label, Object... arguments) {
+        if (arguments.length > 0 && arguments[0] instanceof Locale locale) {
```
sebhofmann: locale is not used and should be removed

sebhofmann (approved):

PR #2023, opened 2023-11-29
MCR-3010 Duplicate warnings about old absolute include mechanism congest logs

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3010).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/common/xml/MCRURIResolver.java
+++ b/mycore-base/src/main/java/org/mycore/common/xml/MCRURIResolver.java
@@ -38,6 +38,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Hashtable;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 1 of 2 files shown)


sebhofmann:
You can do similar things with log4j.
```
<Logger name="org.mycore.common.xml.MCRURIResolver" level="INFO">
    <BurstFilter level="WARN" rate="0.5" maxBurst="10" />
</Logger>
```
Maybe also with CompositeFilter and RegExp to only match this specific message. [Log4j Filter Documentation](https://logging.apache.org/log4j/2.x/manual/filters.html)

EDIT:
If this is not enough i would still prefer a log4j filter Implementation for this case, which could be reused in other classes

toKrause:
If I understand the Log4j documentation correctly, this, if applied in combination with a properly configured `RegexFilter`, would rate-limit and drop messages matching the string `The Stylesheet .* has include .* which only works with an old absolute include mechanism. Please change the include to relative!`.

With the suggested PR, every combination of the two parameters will be logged at least once. With the burst filter, it might very well be possible, that some combinations are always dropped (b/c the are at the end of burst and all bursts will have the same structure), while other combinations are often logged (b/c the are at the beginning of burst and all bursts will have the same structure).

toKrause:
I've created a set of draft PRs (https://github.com/MyCoRe-Org/mycore/pull/2025, https://github.com/MyCoRe-Org/mir/pull/936) for a logging filter approach. Not sure if that is really better, as it is a bit cumbersome (to use and to configure).

PR #2024, opened 2023-11-30
MCR-2647 fix database configuration for JUnit tests

yagee-de:
remove hsqldb as a test dependency
minor bug fixes

[Link to jira](https://mycore.atlassian.net/browse/MCR-2647).


mycore-base/pom.xml
```diff
@@ -207,6 +207,10 @@
       <groupId>com.h2database</groupId>
       <artifactId>h2</artifactId>
     </dependency>
+    <dependency>
```
rsteph-de: duplicate entry?

mycore-mods/pom.xml
```diff
@@ -57,6 +57,10 @@
       <groupId>com.h2database</groupId>
       <artifactId>h2</artifactId>
     </dependency>
+    <dependency>
```
rsteph-de: duplicate dependency?
please check all other pom.xml  too

mycore-packer/src/test/resources/META-INF/persistence.xml
```diff
@@ -8,13 +8,15 @@
...
-      <property name="jakarta.persistence.jdbc.driver" value="org.hsqldb.jdbcDriver" />
-      <property name="jakarta.persistence.jdbc.url" value="jdbc:hsqldb:mem:mycore" />
+      <property name="jakarta.persistence.jdbc.driver" value="org.h2.Driver" />
+      <property name="jakarta.persistence.jdbc.url" value="jdbc:h2:mem:mycore" />
       <property name="jakarta.persistence.jdbc.user" value="sa" />
       <property name="jakarta.persistence.jdbc.password" value="" />
       <property name="hibernate.cache.use_second_level_cache" value="false" />
       <property name="hibernate.jdbc.batch_size" value="0" />
       <property name="hibernate.show_sql" value="false" />
+      <property name="hibernate.globally_quoted_identifiers" value="true" />
```
rsteph-de, line 18: Should these properties be mentioned in the persistence.xml template file for the "real" application?
Are they only for H2 or the other dabase configurations as well?

rsteph-de (changes requested):

mycore-base/pom.xml
yagee-de: removed it in 3 locations

mycore-mods/pom.xml
yagee-de: removed in 3 locations

mycore-packer/src/test/resources/META-INF/persistence.xml
yagee-de: In https://github.com/MyCoRe-Org/mycore/blob/40d3deaf86776e84a43756aa1b31db124c4d580e/mycore-base/src/main/resources/configdir.template/resources/META-INF/persistence.xml#L44 another property is used that can be always enabled and MUST be enabled when H2 is used.

rsteph-de:
We should remove hsqldb settings from `mycore-pi/src/test/resources/META-INF/persistence.xml`
Is the following file still in use or can we delete it: `mycore-base\src\test\resources\META-INF\hibernate.xml`

yagee-de:
> We should remove hsqldb settings from `mycore-pi/src/test/resources/META-INF/persistence.xml` Is the following file still in use or can we delete it: `mycore-base\src\test\resources\META-INF\hibernate.xml`

done!

sebhofmann (approved):
I think this is fine. We need to check if this requires a migration for existing tables.

rsteph-de (approved):
looks good for me

PR #2025, opened 2023-11-30
 MCR-3010 Duplicate warnings about old absolute include mechanism congest logs 

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3010).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/common/log4j2/filter/MCRUniqueFilter.java
+++ b/mycore-base/src/main/java/org/mycore/common/log4j2/filter/MCRUniqueFilter.java
@@ -6,10 +6,8 @@ import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 import java.util.Set;
 
-import org.apache.logging.log4j.Marker;
 import org.apache.logging.log4j.core.Filter;
 import org.apache.logging.log4j.core.LogEvent;
 import org.apache.logging.log4j.core.config.Node;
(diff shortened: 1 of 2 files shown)


mycore-base/src/main/java/org/mycore/common/log4j2/filter/MCRUniqueFilter.java
```diff
@@ -0,0 +1,85 @@
...
+    private MCRUniqueFilter(String targetMarker, int cacheSizePerFormat, Result onMatch, Result onMismatch) {
+        super(onMatch, onMismatch);
+        this.targetMarker = targetMarker;
+        this.cacheSizePerFormat = cacheSizePerFormat;
+    }
+
+    @Override
+    public Result filter(LogEvent event) {
+        if (event.getMessage() instanceof ParameterizedMessage message) {
+            Marker marker = event.getMarker();
```
sebhofmann: I would remove the Marker since there is already a [MarkerFilter](https://logging.apache.org/log4j/2.x/manual/filters.html#markerfilter) in log4j. It makes the Filter more flexible since it could be also Combined with RegExp.

sebhofmann (changes requested):

mycore-base/src/main/java/org/mycore/common/log4j2/filter/MCRUniqueFilter.java
sebhofmann: It would look like, if i understand the log4j api correct:
```
   <logger name="org.mycore.common.xml.MCRURIResolver" >
      <Filters>
        <MarkerFilter marker="tryResolveXML" onMatch="NEUTRAL" onMismatch="ACCEPT"/>
         <MCRUniqueFilter />
      </Filters>
    </logger>
```
I think to match the log4j logic the MCRUniqueFilter should return the onMatch and the onMismatch fields, instead of Result.Accept and Result.Deny.

toKrause: Background: I've added the marker to reduce the resource load. Each "message format string" will create a cache for the sets of parameter values used in combination with that "message format string", so I'd like to have a mechanism to restrict the amount of caches to the bare minimum.

Log4js [documentation](https://logging.apache.org/log4j/2.x/manual/filters.html#CompositeFilter) on the composite filter is quite sparse.

> The CompositeFilter provides a way to specify more than one filter. It is added to the configuration as a filters element and contains other filters to be evaluated. The filters element accepts no parameters. 

The class comment on the [implementation](https://logging.apache.org/log4j/log4j-2.17.2/log4j-core/apidocs/src-html/org/apache/logging/log4j/core/filter/CompositeFilter.html) is not helpful either.

> Composes and invokes one or more filters.

Looking at the implementation shows that the first filter result that yields a non-neutral result is used.

```
@Override
public Result filter(final LogEvent event) {
    Result result = Result.NEUTRAL;
    for (int i = 0; i < filters.length; i++) {
        result = filters[i].filter(event);
        if (result == Result.ACCEPT || result == Result.DENY) {
            return result;
        }
    }
    return result;
}
```

So, I think your suggestion should work and, importantly, also doesn't create unnecessary caches.

sebhofmann (approved):

PR #2027, opened 2023-12-05
MCR-3013 Improve speed of MODS merging

fluetze:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3013).


chemicus (approved):
The speed improvement is impressive and in almost all practical cases the scaling behavior is reduced from quadradically to linear. In fact this is also a bugfix in some respect, since seemingly duplicate authors (there are plenty in http://dx.doi.org/10.1080/15548627.2020.1797280) are not merged anymore.
I will add a suitable testcase for this in a commit. This test should succeed with this PR but fails with the old implementations

PR #2029, opened 2023-12-06
MCR-3008 remove jpa-defaults.xml from property

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3008).

The jpa-defaults is present in the property which leads to errors when creating the configuration dir.


rsteph-de (approved):
O.K.

PR #2045, opened 2024-01-16
MCR-3027 Ability to store references to user information in configuration properties

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3027).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/common/MCRUserInformationResolver.java
+++ b/mycore-base/src/main/java/org/mycore/common/MCRUserInformationResolver.java
@@ -1,0 +1,141 @@
+/*
+ * This file is part of ***  M y C o R e  ***
+ * See http://www.mycore.de/ for details.
+ *
+ * MyCoRe is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * MyCoRe is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with MyCoRe.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.mycore.common;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+
+import org.apache.logging.log4j.Logger;
+import org.apache.logging.log4j.status.StatusLogger;
+import org.mycore.common.config.MCRConfiguration2;
+import org.mycore.common.config.MCRConfigurationException;
+import org.mycore.common.log.MCRListMessage;
+
+/**
+ * A {@link MCRUserInformationResolver} can be used to obtain {@link MCRUserInformation}, without knowledge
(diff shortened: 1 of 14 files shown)


Mewel:
What is the purpose of MCRListMessage? Seems to me like a limited version of MCRTreeMessage. Think one should be enough cause you only use one and not both.

toKrause:
`MCRListMessage` is a message formatter for "list-like" data.

`MCRTreeMessage` is a message formatter for "tree-like" data.

`MCRListMessage` internaly uses a `MCRTreeMessage` for its implementation.

`MCRUserInformationResolver` uses a `MCRListMessage` to log the configured *list* of providers. Using a `MCRTreeMessage` here would be semantically confusing. Why use something to render a tree in order to render a list?

mycore-base/src/main/java/org/mycore/common/MCRUserInformationResolver.java
```diff
@@ -0,0 +1,141 @@
...
+
+    private final Map<String, MCRUserInformationProvider> providers;
+
+    public MCRUserInformationResolver(Map<String, MCRUserInformationProvider> providers) {
+        this.providers = Objects.requireNonNull(providers);
+    }
+
+    private static Map<String, MCRUserInformationProvider> getProviders() {
+
+        List<MCRUserInformationProvider> providers = MCRConfiguration2
```
sebhofmann: Maybe it would be better if we use org.mycore.common.config.MCRConfiguration2#getInstances .

1. It would be the same principle as uri resolvers
2. It would avoid the issue that Providers can have the same schema
3. It would make it easier to exchange a specific provider
4. It would make the getProviders Method easier. 
5. The getSchema method could be removed

sebhofmann (dismissed):
It looks good. Maybe consider using the other property style (like uri resolvers)

Mewel:
> `MCRListMessage` is a message formatter for "list-like" data.
> 
> `MCRTreeMessage` is a message formatter for "tree-like" data.
> 
> `MCRListMessage` internaly uses a `MCRTreeMessage` for its implementation.
> 
> `MCRUserInformationResolver` uses a `MCRListMessage` to log the configured _list_ of providers. Using a `MCRTreeMessage` here would be semantically confusing. Why use something to render a tree in order to render a list?

I can see that. But you never use the tree functionality. Or do I miss it? Why implement something you never use.
It makes the code harder to understand.

rsteph-de:
> Why implement something you never use.
Because MyCoRe is a Framework and this feature could/will  be used in our own applications.

toKrause:
> But you never use the tree functionality. Or do I miss it?

Now I've got you. `MCRTreeMessage` is a part of [MCR-2881](https://github.com/MyCoRe-Org/mycore/pull/1849) and is used there. But it might be a while, until that PR is merged, so I've copied it from there in order to implement `MCRListMessage` w/o much effort. `MCRTreeMessage` will be removed from MCR-2881 once this PR is merged.

[MCR-2881]: https://mycore.atlassian.net/browse/MCR-2881?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

mycore-base/src/main/java/org/mycore/common/MCRUserInformationResolver.java
toKrause: I'll look into that.

sebhofmann (dismissed):

toKrause:
I've made an improvement/simplification in the naming schema.

Now, if `MCRXXX` is an implementation of `MCRUserInformation`, the corresponding provider is consistently called `MCRXXXProvider`.

sebhofmann (approved):

PR #2046, opened 2024-01-17
MCR-2948 move name declaration

Mewel:
* fix createXML() changes model
* move name declaration from validation() to setters
* cleanup class
* add test case

[Link to jira](https://mycore.atlassian.net/browse/MCR-2948).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/datamodel/metadata/MCRMetaPersonName.java
+++ b/mycore-base/src/main/java/org/mycore/datamodel/metadata/MCRMetaPersonName.java
@@ -121,11 +121,8 @@ public final class MCRMetaPersonName extends MCRMetaDefault {
      * This method set the first name text element.
      */
     public void setFirstName(String firstname) {
-        if (firstname != null) {
-            this.firstname = firstname.trim();
-        } else {
-            this.firstname = "";
-        }
+        this.firstname = firstname != null ? firstname.trim() : "";
+        updateName();
     }
 
     /**
@@ -141,11 +138,8 @@ public final class MCRMetaPersonName extends MCRMetaDefault {
      * This method set the call name text element.
      */
     public void setCallName(String callname) {
-        if (callname != null) {
-            this.callname = callname.trim();
-        } else {
-            this.callname = "";
-        }
+        this.callname = callname != null ? callname.trim() : "";
+        updateName();
     }
 
     /**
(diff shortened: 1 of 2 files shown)


mycore-base/src/main/java/org/mycore/datamodel/metadata/MCRMetaPersonName.java
```diff
@@ -301,10 +268,22 @@ public String getAffix() {
...
      */
     public void setAffix(String affix) {
-        if (affix != null) {
-            this.affix = affix.trim();
-        } else {
-            this.affix = "";
+        this.affix = affix != null ? affix.trim() : "";
+    }
+
+    private void updateName() {
```
rsteph-de: I think the whole update method is unnecessary
I would move the checks for empty and the specified calculated fallbacks into the corresponding getters
getFirstname(), getCallName(), getFullName();

``` java
public String getCallname(){
  callname = callname.trim();
  return callname.isEmpty() ? callname : firstname.trim();
}
```
Otherwise you can not distinguish between explicitely set values and calculated defaults.

rsteph-de:
Isn't the null-check in the beginning of the validate() method "dead code", if all values are initialized with empty strings?
You might check for empty strings, but that should be decided by someone who uses this class.

Mewel:
I kept the changes as close to the original code as possible. I don't want to change behavior which can backfire in the applications (like setting names to null via setters).
I just wanted to fix the "bug" that createXML() updates the model itself.

mycore-base/src/main/java/org/mycore/datamodel/metadata/MCRMetaPersonName.java
Mewel: I removed the update method.

mycore-base/src/main/java/org/mycore/datamodel/metadata/MCRMetaPersonName.java
```diff
@@ -363,38 +333,20 @@ public Element createXML() throws MCRException {
...
-            callname = firstname;
-        }
-        fullname = fullname.trim();
-        if (fullname.length() == 0) {
-            String sb = academic + ' ' + peerage + ' ' + firstname + ' ' + prefix + ' ' + surname;
-            fullname = sb.trim();
-            if (fullname.length() == 0) {
-                throw new MCRException(getSubTag() + ": full name / first name or surname is empty");
-            }
+        if (getFullName().isEmpty()) {
```
bt1cu, line 343: The setting of full name should be every time for valid entry.

bt1cu (approved):

kkrebs (changes requested):
please rebase to 2023.06

kkrebs (approved):

PR #2047, opened 2024-01-18
MCR-2968 fix ArrayIndexOutOfBoundsException

Mewel:
MCRLanguageResolver - return empty language tag when language code is empty

[Link to jira](https://mycore.atlassian.net/browse/MCR-2968).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/datamodel/language/MCRLanguageResolver.java
+++ b/mycore-base/src/main/java/org/mycore/datamodel/language/MCRLanguageResolver.java
@@ -24,6 +24,8 @@ import javax.xml.transform.Source;
 import javax.xml.transform.TransformerException;
 import javax.xml.transform.URIResolver;
 
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.jdom2.Document;
 import org.jdom2.Element;
 import org.jdom2.transform.JDOMSource;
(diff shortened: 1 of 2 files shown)


Possommi (approved):

mycore-base/src/main/java/org/mycore/datamodel/language/MCRLanguageResolver.java
```diff
@@ -24,21 +24,30 @@
...
  */
 public class MCRLanguageResolver implements URIResolver {
 
+    private static final Logger LOGGER = LogManager.getLogger();
+
     public Source resolve(String href, String base) throws TransformerException {
         try {
-            String code = href.split(":")[1];
+            String[] hrefContent = href.split(":");
+            if (hrefContent.length < 2) {
```
sebhofmann, line 46: I think it would be better if this method throws an error if the parameter is missing. The error should be IllegalArgumentException. The iso code should not be optional. If you want to ignore the exception, than you should use notNull:language:... , which should catch the exception.

mycore-base/src/main/java/org/mycore/datamodel/language/MCRLanguageResolver.java
```diff
@@ -24,21 +24,30 @@
...
+                LOGGER.warn("Empty language code found while resolving URI 'language:'. Returning empty language tag.");
+                return new JDOMSource(new Document(new Element("language")));
```
sebhofmann: I think it would be better if this method throws an error if the parameter is missing. The error should be IllegalArgumentException. The iso code should not be optional. If you want to ignore the exception, than you should use notNull:language:... , which should catch the exception.

sebhofmann (changes requested):

mycore-base/src/main/java/org/mycore/datamodel/language/MCRLanguageResolver.java
Mewel: My comment: https://mycore.atlassian.net/browse/MCR-2968?focusedCommentId=49071

sebhofmann (approved):

PR #2049, opened 2024-01-22
MCR-3028 make MyCoRe compile with Java 21

yagee-de:
update to PMD 7.0.0
fix newly detected PMD violations

[Link to jira](https://mycore.atlassian.net/browse/MCR-3028).

PR #2050, opened 2024-01-24
MCR-3029 use java 21 features

yagee-de:
Requires MCR-3028 (https://github.com/MyCoRe-Org/mycore/pull/2049)

[Link to jira](https://mycore.atlassian.net/browse/MCR-3029).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/common/MCRCalendar.java
+++ b/mycore-base/src/main/java/org/mycore/common/MCRCalendar.java
@@ -898,15 +898,12 @@ public class MCRCalendar {
         }
         try {
             formatter.setCalendar(calendar);
-            if (calendar instanceof IslamicCalendar) {
-                return formatter.format(calendar.getTime()) + " h.";
-            } else if (calendar instanceof CopticCalendar) {
-                return formatter.format(calendar.getTime()) + " A.M.";
-            } else if (calendar instanceof EthiopicCalendar) {
-                return formatter.format(calendar.getTime()) + " E.E.";
-            } else {
-                return formatter.format(calendar.getTime());
-            }
+            return switch (calendar) {
+                case IslamicCalendar islamicCalendar -> formatter.format(calendar.getTime()) + " h.";
+                case CopticCalendar copticCalendar -> formatter.format(calendar.getTime()) + " A.M.";
+                case EthiopicCalendar ethiopicCalendar -> formatter.format(calendar.getTime()) + " E.E.";
+                default -> formatter.format(calendar.getTime());
+            };
         } catch (Exception e) {
             return "";
         }
@@ -953,23 +950,16 @@ public class MCRCalendar {
         if (calendar == null) {
             return "";
         }
-        if (calendar instanceof IslamicCalendar) {
-            return TAG_ISLAMIC;
-        } else if (calendar instanceof BuddhistCalendar) {
-            return TAG_BUDDHIST;
-        } else if (calendar instanceof CopticCalendar) {
-            return TAG_COPTIC;
-        } else if (calendar instanceof EthiopicCalendar) {
-            return TAG_ETHIOPIC;
-        } else if (calendar instanceof HebrewCalendar) {
-            return TAG_HEBREW;
-        } else if (calendar instanceof JapaneseCalendar) {
-            return TAG_JAPANESE;
-        } else if (calendar instanceof GregorianCalendar) {
-            return TAG_GREGORIAN;
-        } else {
-            return TAG_JULIAN;
-        }
+        return switch (calendar) {
+            case IslamicCalendar islamicCalendar -> TAG_ISLAMIC;
+            case BuddhistCalendar buddhistCalendar -> TAG_BUDDHIST;
+            case CopticCalendar copticCalendar -> TAG_COPTIC;
+            case EthiopicCalendar ethiopicCalendar -> TAG_ETHIOPIC;
+            case HebrewCalendar hebrewCalendar -> TAG_HEBREW;
+            case JapaneseCalendar japaneseCalendar -> TAG_JAPANESE;
+            case GregorianCalendar gregorianCalendar -> TAG_GREGORIAN;
+            default -> TAG_JULIAN;
+        };
     }
 
     /**
--- a/mycore-base/src/main/java/org/mycore/common/MCRCalendar.java
+++ b/mycore-base/src/main/java/org/mycore/common/MCRCalendar.java
@@ -651,76 +651,58 @@ public class MCRCalendar {
             final boolean bb = beforeZero(dateTrimmed, CalendarType.Persic);
             final String cleanDate = cleanDate(dateTrimmed, CalendarType.Persic);
             final int[] fields = parseDateString(cleanDate, last, CalendarType.Persic);
-            final int year = fields[0];
+            int year = fields[0];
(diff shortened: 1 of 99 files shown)


mycore-base/src/main/java/org/mycore/common/content/MCRSourceContent.java
```diff
@@ -117,6 +69,58 @@ public MCRSourceContent(Source source) {
         this.setBaseContent(baseContent);
     }
 
+    private static MCRContent getBaseContent(StreamSource streamSource) {
```
Mewel, line 73: Any reason why those getBaseContent() methods are static?

mycore-base/src/main/java/org/mycore/common/MCRCalendar.java
```diff
@@ -953,23 +950,16 @@ public static String getCalendarTypeString(Calendar calendar) {
         if (calendar == null) {
```
rsteph-de: this can be added to switch as "case null"

rsteph-de (changes requested):
found a "case null" case for witch

mycore-base/src/main/java/org/mycore/common/content/MCRSourceContent.java
rsteph-de: probably because they are independent of the state of the object ?

mycore-base/src/main/java/org/mycore/common/content/MCRSourceContent.java
```diff
@@ -117,6 +69,58 @@ public MCRSourceContent(Source source) {
...
+        Document xml = src.getDocument();
+        if (xml != null) {
+            return new MCRJDOMContent(xml);
+        } else {
+            for (Object node : src.getNodes()) {
+                if (node instanceof Element element) {
+                    Document doc = element.getDocument();
+                    if (doc == null) {
+                        return new MCRJDOMContent(element);
+                    } else {
```
Mewel: Not required else case

yagee-de: Sorry. Do not understand why. Can you explain please?

mycore-base/src/main/java/org/mycore/common/MCRCalendar.java
yagee-de: Should be addressed by https://github.com/MyCoRe-Org/mycore/pull/2050/commits/78d3dbbcbca301a539778217a374cc08a77f3758

mycore-base/src/main/java/org/mycore/common/content/MCRSourceContent.java
Mewel: When doc is equal null, it is returned, therefore the else is not required:
```
if (doc == null) {
  return new MCRJDOMContent(element);
} 
if (doc.getRootElement() == element) {
  return new MCRJDOMContent(doc);
} else {
  return new MCRJDOMContent(element.clone());
}
```
is sufficient, removes one nesting

mycore-base/src/main/java/org/mycore/common/content/MCRSourceContent.java
Mewel: True, but it also has no benefit. The methods are never used in any static way.

mycore-base/src/main/java/org/mycore/common/content/MCRSourceContent.java
yagee-de: Should be addressed in https://github.com/MyCoRe-Org/mycore/pull/2050/commits/41defe31da69bae47ba323e2c9337f46ae90e3a6

mycore-base/src/main/java/org/mycore/common/MCRCalendar.java
```diff
@@ -651,76 +651,58 @@ protected static GregorianCalendar getCalendarFromPersicDate(String datestr, boo
             final boolean bb = beforeZero(dateTrimmed, CalendarType.Persic);
             final String cleanDate = cleanDate(dateTrimmed, CalendarType.Persic);
             final int[] fields = parseDateString(cleanDate, last, CalendarType.Persic);
-            final int year = fields[0];
+            int year = fields[0];
```
Mewel: Same as:
`final year = bb ? 622 - fields[0] : 621 + fields[0] ;`
Or?

mycore-base/src/main/java/org/mycore/common/content/MCRSourceContent.java
yagee-de: There is a HUGE 🧐 benefit. Static methods a resolved at compile time and can be optimized (inlined), while instance methods can only be resolved at runtime. So marking method as static provides a benefit, if no state is required. What else should "private static" methods be used for, when there is no benefit?

rsteph-de (dismissed):
from my point of view the remaining open issues ares resolved - The PR is ready to be merged.

mycore-base/src/main/java/org/mycore/common/content/MCRSourceContent.java
```diff
@@ -117,6 +70,52 @@ public MCRSourceContent(Source source) {
...
+    private static MCRJDOMContent getBaseContent(JDOMSource src) {
+        Document xml = src.getDocument();
+        if (xml != null) {
+            return new MCRJDOMContent(xml);
+        }
+        return src.getNodes().stream()
+            .filter(Parent.class::isInstance) //only Element or Document matter
+            .map(Parent.class::cast)
+            .findFirst()
+            .map(p -> switch (p){
```
Mewel, line 109: Looks correct. This method is still very confusing though. In one case you want the document, in another the element and in the third a cloned element.

yagee-de: It's not what I want, but what I get. The resulting content should not return more "content" than that is returned by the Source. These are all cases that occurred in the wild and had to be handled like this. The amount of cases is mainly to reduce the work that has to be done to get the proper result. One could have done a simple "clone()" on Element to have a detached copy of the given element, but this involves a copy every time even if it's not really needed.

mycore-base/src/main/java/org/mycore/common/MCRCalendar.java
yagee-de: is addressed in https://github.com/MyCoRe-Org/mycore/pull/2050/commits/a80138fbebbc50ad293b2d31a6649aa63cc0d35c

mycore-base/src/main/java/org/mycore/common/content/MCRSourceContent.java
Mewel: I can accept that, but I highly doubt there is any difference in the real world. The JIT-Compiler is probably optimizing this anyway.
In my opinion such decision should never be based on performance (except there is a real reason for it). I would always choose readability, testability and extensibility over it.
But I guess its ok in this case cause we don't have any of that.

Mewel (approved):

mycore-base/src/main/java/org/mycore/common/content/MCRSourceContent.java
sebhofmann: These methods should be in StreamSource, JDOMSource...

PR #2056, opened 2024-02-01
MCR-3033 MCRPIJobService should use MCRUserInformationResolver

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3033).

### DIFF
--- a/mycore-pi/pom.xml
+++ b/mycore-pi/pom.xml
@@ -121,11 +121,6 @@
       <artifactId>mycore-jobqueue</artifactId>
     </dependency>
     <dependency>
-      <groupId>org.mycore</groupId>
-      <artifactId>mycore-user2</artifactId>
-      <optional>true</optional>
-    </dependency>
-    <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
       <scope>test</scope>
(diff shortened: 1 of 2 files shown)


mycore-pi/pom.xml
```diff
@@ -120,11 +120,6 @@
       <groupId>org.mycore</groupId>
       <artifactId>mycore-jobqueue</artifactId>
     </dependency>
-    <dependency>
-      <groupId>org.mycore</groupId>
-      <artifactId>mycore-user2</artifactId>
-      <optional>true</optional>
```
sebhofmann, line 126: I think that has to be a optional runtime dependency. See [Maven Optional](https://maven.apache.org/guides/introduction/introduction-to-optional-and-excludes-dependencies.html).

toKrause: I don't see why that dependency would be needed anymore. With this PR, no direct references to classes from `mycore-user2` exist in `mycore-pi` anymore. Using `MCRUserInformationResolver`  effectively decouples `mycore-pi` from `mycore-user2`. I'm pretty sure that the Maven dependency plugin would rightfully flag `mycore-user2` as an unused dependency and fail the build, if I kept it as an dependency.

sebhofmann (approved):

PR #2058, opened 2024-02-02
MCR-3037 Added getUsableSpace() to MCRXMLFunctions

Possommi:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3037).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/common/xml/MCRXMLFunctions.java
+++ b/mycore-base/src/main/java/org/mycore/common/xml/MCRXMLFunctions.java
@@ -865,7 +865,7 @@ public class MCRXMLFunctions {
      * @throws IOException
      * */
     public static long getUsableSpace() throws IOException{
-        Path dataDir = Paths.get(MCRConfiguration2.getString("MCR.datadir").get());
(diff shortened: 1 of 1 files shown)


mycore-base/src/main/java/org/mycore/common/xml/MCRXMLFunctions.java
```diff
@@ -856,6 +858,20 @@ public static String stripHtml(final String s) {
...
     }
 
+    /**
+     * Returns approximately the usable space in the MCR.datadir in bytes as in {@link FileStore#getUsableSpace()}.
+     *
+     * @return approximately the usable space in the MCR.datadir
+     * @throws IOException
+     * */
+    public static long getUsableSpace() throws IOException{
+        Path dataDir = Paths.get(MCRConfiguration2.getString("MCR.datadir").get());
```
toKrause: `MCR.datadir` should always be set in an ordinary MyCoRe instance. Still, should use `MCRConfiguration2.getStringOrThrow("MCR.datadir")` instead of `MCRConfiguration2.getString("MCR.datadir").get()` in order to generate a meaningful error message in the unusual case where no value is set.

toKrause (changes requested):

toKrause (approved):

PR #2059, opened 2024-02-05
MCR-3039 fix upload size detection and fix error messages

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3039).


toKrause:
I've tested the proposed changes by cherry-picking onto release version 2023.06 of MyCoRe and using release version 2023.06 of MIR.

In case of missing rights:
- Improved user-facing error message ("Das Hochladen ist nicht erlaubt.") is working
- **No error UI displayed when trying to upload with "select file"**
- **No error UI displayed when trying to upload with "drag&drop"**

In case of invalid file name
- Improved user-facing error message ("Die Datei ... ist nicht valide: Der Datei-Name enthält ungültige Zeichen.") is working
- Error UI displayed when trying to upload with "select file"
- Error UI displayed when trying to upload with "drag&drop"

In case of unacceptable file size
- Improved user-facing error message ("Die Datei ... ist nicht valide: Die Datei ist zu groß: ... / ...") is working
- Error UI displayed when trying to upload with "select file"
- **No Error UI displayed when trying to upload with "drag&drop"**

sebhofmann:
> In case of missing rights:
> 
> * Improved user-facing error message ("Das Hochladen ist nicht erlaubt.") is working
> * **No error UI displayed when trying to upload with "select file"**
> * **No error UI displayed when trying to upload with "drag&drop"**
> In case of invalid file name

This should work now.

> * Improved user-facing error message ("Die Datei ... ist nicht valide: Die Datei ist zu groß: ... / ...") is working
> * Error UI displayed when trying to upload with "select file"
> * **No Error UI displayed when trying to upload with "drag&drop"**

This works on my pc with chrome.
I just saw that it does not work with firefox and i fixed it.

toKrause:
I've tested again on my system (Ubuntu 22.04) with Firefox (122.0) and Chrome (121.0.6167.160) and, generally speaking, the issues appear to be fixed now.

But on Chrome, it appears to be a bit flaky. With both methods (drag&drop, file chooser) in about 10% of the cases, the upload doesn't even start. The process hangs in this state:
image

According to the network tab of the developer tools, no request to the upload handler is ever even made. Sadly, I haven't been able to discern any meaningful pattern.

toKrause:
Stack trace in failing cases is:

```
upload-api.js:1 Uncaught (in promise) TypeError: Cannot read properties of null (reading 'isDirectory')
    at t.<anonymous> (upload-api.js:1:15455)
    at upload-api.js:1:12507
    at Object.next (upload-api.js:1:12612)
    at upload-api.js:1:11548
    at new Promise (<anonymous>)
    at a (upload-api.js:1:11293)
    at t.validateTraverse (upload-api.js:1:15349)
    at t.<anonymous> (upload-api.js:1:15024)
    at upload-api.js:1:12507
    at Object.next (upload-api.js:1:12612)
```
But I have a hard time correlating the minified JS to the original TS. The first entry points to a function named `validateTraverse`. Hope, this helps.

sebhofmann:
I am not sure how this can happen and i can not reproduce it, can you try the new commit if the error gets catched by the validation? There should also appear a log message in the console.

toKrause:
image


```
upload-api.js:1 File entry is null null
(anonymous) @ upload-api.js:1
(anonymous) @ upload-api.js:1
(anonymous) @ upload-api.js:1
(anonymous) @ upload-api.js:1
a @ upload-api.js:1
t.validateTraverse @ upload-api.js:1
(anonymous) @ upload-api.js:1
(anonymous) @ upload-api.js:1
(anonymous) @ upload-api.js:1
(anonymous) @ upload-api.js:1
a @ upload-api.js:1
t.processDrop @ upload-api.js:1
(anonymous) @ upload-api.js:1
```

sebhofmann:
I dont understand how that can happen. The validateTraverse function is called recursive that means the children are read using `const reader fileEntry.createReader()` and `reader.readEntries()`. 
The docs of readEntries state the following:

> successCallback – A function which is called when the directory's contents have been retrieved. The function receives a single input parameter: an array of file system entry objects, each based on FileSystemEntry. Generally, they are either FileSystemFileEntry objects, which represent standard files, or FileSystemDirectoryEntry objects, which represent directories. If there are no files left, or you've already called readEntries() on this FileSystemDirectoryReader, the array is empty.

But in your case there is a Array returned which contains a null object and not a FileSystemEntry.

toKrause:
Maybe an issue with Chrome, rather then your code…?

toKrause:
I've added some logging to `processDrop`:
```
const webkitEntries: any[] = [];
for (let i = 0; i < items.length; i++) {
    console.log("index: ", i);
    console.log("before: ", items[i]);
    const x = items[i].webkitGetAsEntry();
    console.log("after: ", x);
    webkitEntries.push(x);
}
```

When successful:

```
index:  0
before:  DataTransferItem {kind: 'file', type: ''}
after:  FileEntry {isFile: true, isDirectory: false, name: 'test--pdf', fullPath: '/test--pdf', filesystem: DOMFileSystem}
```

When unsuccessful:

```
index:  0
before:  DataTransferItem {kind: 'string', type: 'text/plain'}
after:  null
```

Chrome reports the dropped file as a string, ... which is weird, and not conforming to the docs you quoted. 

[`DataTransferItem:webkitGetAsEntry()`](https://developer.mozilla.org/en-US/docs/Web/API/DataTransferItem/webkitGetAsEntry) then works a documented:

> A [`FileSystemEntry`](https://developer.mozilla.org/en-US/docs/Web/API/FileSystemEntry)-based object describing the dropped item. This will be either [`FileSystemFileEntry`](https://developer.mozilla.org/en-US/docs/Web/API/FileSystemFileEntry) or [`FileSystemDirectoryEntry`](https://developer.mozilla.org/en-US/docs/Web/API/FileSystemDirectoryEntry). The method aborts and **returns `null` if the dropped item isn't a file**, or if the [`DataTransferItem`](https://developer.mozilla.org/en-US/docs/Web/API/DataTransferItem) object is not in read or read/write mode.

sebhofmann:
I filtered so only file drops are used, which should fix your issue. Also allow to select multiple files with the file selector  and optimized how often the validation message is shown and hidden.

toKrause (approved):

PR #2064, opened 2024-02-14
MCR-3044 Files with certain special characters in their names are not properly supported

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3044).


yagee-de:
There should be no MIR ticket in MyCoRe code base.

toKrause:
If a MIR issues requires groundwork to be done in MyCoRe, this should be fine. We've discussed and approved this approach with MIR-1132, see 83d5c5cf7ff537c2262d108c02565c712d07b35c.

Nevertheless, I've created a MyCoRe issue and updated the commit, because, as it turns out, there was more work to be done that affected MyCoRe proper.

yagee-de (approved):

PR #2066, opened 2024-02-16
MCR 3045 Neo4J XSL does not produce HTML

michael-becker:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3045).

With saxon, we need to add a media-type statement to create HTML output from XSL.


kkrebs:
seems to be a bug and should be fixed in 2023.06.x, we will merge this into main

mycore-neo4j/src/main/resources/xsl/mycoreobject-neo4j.xsl
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 
-<xsl:stylesheet version="1.0" xmlns:xsl="http://www.w3.org/1999/XSL/Transform">
-  <xsl:output method="html" encoding="UTF-8" indent="yes"/>
+<xsl:stylesheet version="2.0" xmlns:xsl="http://www.w3.org/1999/XSL/Transform">
```
toKrause: This file resides in the `xsl` folder and should therefore be processable by Xalan. Afaik, that precludes a version other then `1.0`, or am I confused about that?

michael-becker:
I had a little hazzle switching the base version; the updated PR is available at https://github.com/MyCoRe-Org/mycore/pull/2066

PR #2068, opened 2024-02-19
MCR-3043 new text editor

Mewel:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3043).


mycore-webtools/src/main/vue/texteditor/src/apis/BaseContentHandler.ts
```diff
@@ -0,0 +1,44 @@
...
+     *
+     * @param id the id
+     * @param response http response
+     */
+    handleWriteAccessResponse(id: string, response: Response) {
+        if (!response.ok) {
+            throw this.buildError(`Unable get OPTIONS of ${id}.`, response);
+        }
+        //let allow = response.headers.get("Allow");
+        const allow = "HEAD,DELETE,GET,OPTIONS,PUT";
```
sebhofmann, line 37: Add a // TODO or use the response.headers.get...

mycore-webtools/src/main/vue/texteditor/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "vue": "^3.4.15",
+    "vue-prism-editor": "^2.0.0-alpha.2",
+    "vue-router": "^4.2.5"
+  },
+  "devDependencies": {
+    "@tsconfig/node20": "^20.1.2",
+    "@types/node": "^20.11.10",
+    "@vitejs/plugin-vue": "^5.0.3",
+    "@vue/tsconfig": "^0.5.1",
+    "npm-run-all2": "^6.1.1",
```
sebhofmann: Is npm-run-all2 really required?

mycore-webtools/src/main/vue/texteditor/src/apis/ClassificationsContentHandler.ts
```diff
@@ -0,0 +1,65 @@
...
+        // TODO - rest API v2 does not provide support for checking write access
+        return true;
+    }
+
+    /**
+     * https://stackoverflow.com/questions/376373/pretty-printing-xml-with-javascript
+     *
+     * @param xml xml to prettify
+     */
+    prettifyXml(xml: string) {
```
sebhofmann, line 53: I don't like string based xml modification. Can you try https://github.com/riversun/xml-beautify ? It should be easy usable.

mycore-webtools/src/main/vue/texteditor/src/views/TextEditorView.vue
```diff
@@ -0,0 +1,134 @@
...
+let success: any = ref();
+let error: any = ref();
+let updateEnabled = ref(false);
+let writeAccess = ref(false);
+let undoKeyPressed = false;
+let originalModel: string = "";
+
+watch(() => model.value.data, async (newModel) => {
+  // prevent ctrl+z clearing the textarea
+  setTimeout(() => {
```
sebhofmann: use nextTick() or even better prevent the ctrl+z browser function by catching the event and return false

sebhofmann (changes requested):

mycore-webtools/src/main/vue/texteditor/src/apis/ClassificationsContentHandler.ts
Mewel: In my opinion the xml from server should be prettified. But this is not supported for classification's yet. So this is more of a workaround for classification's only. I would keep it simple like this, without adding a new dependency.
I will comment it to make it more clear, that this should be removed in the future.

mycore-webtools/src/main/vue/texteditor/src/views/TextEditorView.vue
Mewel: Will use nextTick().
Cannot prevent ctrl+z because I don't know the model state at this point. Ctrl+z should only be prevented if it clears the textarea.

sebhofmann (approved):

PR #2077, opened 2024-02-23
MCR-3050 add "xor" support

yagee-de:
* fixes bug in order of child conditions
* add JUnit tests

[Link to jira](https://mycore.atlassian.net/browse/MCR-3050).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/access/facts/condition/combined/MCRAbstractCombinedCondition.java
+++ b/mycore-base/src/main/java/org/mycore/access/facts/condition/combined/MCRAbstractCombinedCondition.java
@@ -17,11 +17,13 @@
  */
 package org.mycore.access.facts.condition.combined;
 
-import java.util.HashSet;
+import java.util.LinkedHashSet;
+import java.util.Objects;
 import java.util.Set;
 
 import org.jdom2.Element;
 import org.mycore.access.facts.MCRFactsAccessSystemHelper;
+import org.mycore.access.facts.MCRFactsHolder;
 import org.mycore.access.facts.condition.MCRAbstractCondition;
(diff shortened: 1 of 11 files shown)


mycore-base/src/main/java/org/mycore/access/facts/condition/combined/MCRAbstractCombinedCondition.java
```diff
@@ -64,4 +68,14 @@ public void debugInfoForMatchingChildElement(MCRCondition c, boolean matches) {
         }
     }
 
+    boolean addDebugInfoIfRequested(MCRCondition c, MCRFactsHolder facts) {
+        if (!isDebug()) {
+            return c.matches(facts);
+        }
+        boolean matches = c.matches(facts);
```
toKrause, line 75: Maybe:
```
boolean matches = c.matches(facts);
if (!isDebug()) {
    return matches;
}
```
This would make it more obvious at a glance that debug- and non-debug-mode actually perform the same thing.

mycore-base/src/main/java/org/mycore/access/facts/condition/combined/MCRAbstractCombinedCondition.java
```diff
@@ -64,4 +68,14 @@ public void debugInfoForMatchingChildElement(MCRCondition c, boolean matches) {
...
+        Objects.requireNonNull(c.getBoundElement(), "Condition is not bound to an element.")
```
toKrause, line 76: Is `Objects#requireNonNull` really necessary here. This introduces a new possibility for an exception that can only show itself when debug is enabled. The old implementation seems more robust.

toKrause (approved):

PR #2080, opened 2024-02-23
MCR-3051 allow to configure classifications for MODS import

yagee-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3051).

### DIFF
--- a/mycore-mods/src/main/java/org/mycore/mods/MCRMODSCommands.java
+++ b/mycore-mods/src/main/java/org/mycore/mods/MCRMODSCommands.java
@@ -47,7 +47,9 @@ import org.mycore.common.MCRPersistenceException;
 import org.mycore.common.config.MCRConfiguration2;
 import org.mycore.common.xml.MCRXMLHelper;
 import org.mycore.datamodel.common.MCRActiveLinkException;
+import org.mycore.datamodel.classifications2.MCRCategoryID;
 import org.mycore.datamodel.metadata.MCRDerivate;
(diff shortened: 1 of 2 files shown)


mycore-mods/src/main/java/org/mycore/mods/MCRMODSCommands.java
```diff
@@ -195,6 +199,14 @@ private static MCRDerivate createDerivate(MCRObjectID documentID, File fileDir)
             firstRegularFile.ifPresent(ifs::setMainDoc);
             derivate.getDerivate().setInternals(ifs);
         }
+        MCRConfiguration2.getString("MCR.MODS.Import.Derivate.Categories")
+            .map(MCRConfiguration2::splitValue)
+            .ifPresent(s -> {
+                s.map(MCRCategoryID::fromString)
+                    .forEach(categId -> derivate.getDerivate().getClassifications()
+                        .add(new MCRMetaClassification("classification", 0, null,
```
toKrause, line 207: How does that work? The `classId` parameter of `MCRMetaClassification` is always `null` and the entire configuration value chunk (`derivate_types:content`) is passed as the `categId`, which will eventually be used as the value for `MCRCategoryID#setId(String)` where it is checked against the pattern `[^:$\{\}]+`, i.e. it is not allowed to contain a colon. Shouldn't the configuration value chunk be split at the colon ant the parts be used as `classId` and `categId`?

sebhofmann (approved):

PR #2083, opened 2024-02-26
MCR-3053 Redesign API to create configured instances

toKrause:

### DIFF
--- a/mycore-base/src/main/java/org/mycore/common/config/MCRConfigurableInstanceHelper.java
+++ b/mycore-base/src/main/java/org/mycore/common/config/MCRConfigurableInstanceHelper.java
@@ -67,11 +67,45 @@ class MCRConfigurableInstanceHelper {
      */
     public static boolean isSingleton(String property) {
         return MCRConfiguration2.getString(property).stream()
-            .anyMatch(propertyVal -> getClass(property, propertyVal).getDeclaredAnnotation(Singleton.class) != null);
+            .anyMatch(propertyVal -> isSingleton(getClass(property, propertyVal)));
     }
 
     /**
-     * Creates a configured instance of a class .
+     * Checks if a class is annotated with {@link Singleton}.
+     *
+     * @param targetClass the class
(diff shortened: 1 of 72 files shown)


mycore-base/src/main/java/org/mycore/common/config/MCRConfigurableInstanceHelper.java
```diff
@@ -67,11 +67,46 @@ class MCRConfigurableInstanceHelper {
...
+
+    /**
+     * Creates a configured instance of a class.
+     *
+     * @param superClass the intended super class of the instantiated class
+     * @param name the property which contains the class name
+     * @return the configured instance of T
+     * @throws MCRConfigurationException if the property is not right configured.
+     */
+    @SuppressWarnings("unchecked")
```
sebhofmann: This Warning can be removed by changing the Code to:
```
        if (superClass.isInstance(instance)) {
            return Optional.of(superClass.cast(instance));
        } else {
```

Source: https://stackoverflow.com/questions/28921833/why-is-this-an-unchecked-cast-and-how-to-fix-it#answer-28921986

mycore-webtools/src/main/java/org/mycore/webtools/upload/MCRUploadResource.java
```diff
@@ -157,13 +157,8 @@ public Response commit(@PathParam("buckedID") String buckedID) {
     }
 
     public static MCRUploadHandler getUploadHandler(String uploadHandlerID) {
-        Optional<MCRUploadHandler> uploadHandler = MCRConfiguration2
```
sebhofmann, line 160: This is still a bad request and not a configuration issue

sebhofmann (changes requested):

sebhofmann (dismissed):

toKrause:
Rebased to include current changes of `main`

sebhofmann (approved):

PR #2090, opened 2024-02-28
MCR-3021 Remove mapping of ORCID common:language-code

chemicus:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3021).

### DIFF
--- a/mycore-orcid2/src/main/resources/xsl/orcid2/v3/work2mcr_generic.xsl
+++ b/mycore-orcid2/src/main/resources/xsl/orcid2/v3/work2mcr_generic.xsl
@@ -25,7 +25,6 @@
         <xsl:apply-templates select="work:short-description" />
         <xsl:apply-templates select="common:url" />
         <xsl:apply-templates select="common:external-ids/common:external-id[common:external-id-relationship='self']" />
-        <xsl:apply-templates select="common:language-code" />
       </mods:mods>
       <xsl:apply-templates select="work:citation[work:citation-type='bibtex']" />
     </xsl:copy>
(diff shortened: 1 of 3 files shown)


chemicus:
The reported error is due to a xsd-validation failure on the transformed orcid work. According to the scheme https://github.com/ORCID/orcid-model/blob/733021de324030e66bed0e4c821dc8cd30a3ef00/src/main/resources/record_3.0/work-3.0.xsd#L96  only one common:language-code is allowed in the scheme. This makes sense, since this field is meant to store the “language used for the metadata” https://github.com/ORCID/orcid-model/blob/733021de324030e66bed0e4c821dc8cd30a3ef00/src/main/resources/record_3.0/work-3.0.xsd#L98 (and not the language of the work/publication itself).
Therefore the correct mapping of common:language-code is mods:recordInfo/mods:languageOfCataloging rather than mods:language

mycore-orcid2/src/main/resources/xsl/orcid2/v3/work2mcr_generic.xsl
```diff
@@ -208,11 +208,13 @@
...
   <xsl:template match="common:language-code">
-    <mods:language>
-      <mods:languageTerm authority="rfc5646" type="code">
-        <xsl:value-of select="document(concat('language:',text()))/language/@xmlCode" />
-      </mods:languageTerm>
-    </mods:language>
+    <mods:recordInfo>
+      <mods:languageOfCataloging>
+        <mods:languageTerm authority="rfc5646" type="code">
+          <xsl:value-of select="document(concat('language:',text()))/language/@xmlCode" />
```
toKrause, line 214: Should check if a corresponding language classification value exists and handle the case when it doesn't.

toKrause (changes requested):

mycore-orcid2/src/main/resources/xsl/orcid2/v3/work2mcr_generic.xsl
toKrause: Should check if a corresponding language classification value exists and handle the case when this is not the case.

kkrebs (changes requested):
Currently there doesn't seem to be a suitable field for mods:language in ORCID, see https://support.orcid.org/hc/en-us/articles/360006971353-Metadata-in-the-Works-section

That's why we decided to leave it out completely.

kkrebs (approved):

toKrause (approved):

PR #2093, opened 2024-02-28
MCR-3056 Reverse resource lookup causes performance issues

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3056).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/resource/provider/MCRResourceProvider.java
+++ b/mycore-base/src/main/java/org/mycore/resource/provider/MCRResourceProvider.java
@@ -22,11 +22,12 @@ import java.io.File;
 import java.net.URI;
 import java.net.URL;
 import java.util.Collections;
-import java.util.LinkedHashSet;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
-import java.util.Set;
+import java.util.concurrent.ExecutorService;
+import java.util.function.Supplier;
 
 import org.apache.logging.log4j.Level;
 import org.mycore.common.hint.MCRHints;
@@ -34,6 +35,8 @@ import org.mycore.common.log.MCRTreeMessage;
 import org.mycore.resource.MCRResourcePath;
 
 import io.github.classgraph.ClassGraph;
+import io.github.classgraph.ScanResult;
+import nonapi.io.github.classgraph.concurrency.AutoCloseableExecutorService;
 
 /**
  * A {@link MCRResourceProvider} implements a resource lookup strategy.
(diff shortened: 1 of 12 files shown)


mycore-base/src/main/java/org/mycore/resource/provider/MCRResourceProvider.java
```diff
@@ -102,30 +110,57 @@ public BaseDirPrefixStripper(File baseDir) {
...
+        public ClassLoaderPrefixStripper(ClassLoader classLoader) {
+            this.classLoader = Objects.requireNonNull(classLoader);
+        }
+
+        @Override
+        public List<String> getStrippedPaths(String value) {
+            List<String> potentialPaths = new LinkedList<>();
+            ClassGraph classGraph = new ClassGraph();
+            classGraph.overrideClassLoaders(classLoader);
+            try (ScanResult scanResult = classGraph.scan(EXECUTOR_SERVICE, CLASS_GRAPH_THREADS)) {
```
Mewel: This looks weird to me. Using a static EXECUTOR_SERVICE which is autoclosable. Doesn't this imply that once it's used, it is closed and therefore not avaiable anymore in other instances of ClassLoaderPrefixStripper? Or do I see that wrong?

mycore-base/src/main/java/org/mycore/resource/provider/MCRResourceProvider.java
```diff
@@ -102,30 +110,57 @@ public BaseDirPrefixStripper(File baseDir) {
...
-            classpath.forEach(uri -> {
-                if (uri.getScheme().equals("file")) {
-                    File file = new File(uri.getPath());
-                    if (file.isDirectory()) {
-                        strippers.add(new BaseDirPrefixStripper(file));
+    }
+
+    final class ClassLoaderPrefixStripper extends PrefixStripperBase {
+
+        private static final int CLASS_GRAPH_THREADS = 4;
```
Mewel: Why 4? Shouldn't that be configurable?

toKrause: Good point, wanted to add a configuration value and then forgot. The ClassGraph library has a nifty formula to calculate `ClassGraph#DEFAULT_NUM_WORKER_THREADS`, but it can not be reused, because that field is package-private.

mycore-base/src/main/java/org/mycore/resource/provider/MCRResourceProvider.java
toKrause: The `ExecutorService` is not used/closed in the  try-with-resourcse-statement, the `ScanResult` is. Only the `ScanResult` gets closed.

I keept using the `AutoCloseableExecutorService` while deconstructing the formally used method `ClassGraph#getClasspathURIs`, but the auto-closing feature isn't needed now. A simple `ThreadPoolExecutor` should suffice. Will change.

Mewel: ah ok, thanks for clarification.

Mewel (approved):

toKrause:
Rebased to include current changes of `main`

PR #2099, opened 2024-03-11
MCR-3064 support for {CurrentUser} placeholder

yagee-de:
replaces {CurrentUser} in navigation.xml with userID
removes item if user is not logged in

[Link to jira](https://mycore.atlassian.net/browse/MCR-3064).


sebhofmann (approved):

toKrause (commented):
Might need changes, see [MIR-1288](https://github.com/MyCoRe-Org/mir/pull/974)

PR #2100, opened 2024-03-12
MCR-3063 Searching by attributes leads display of same records multiple times

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3063).


Possommi (commented):
the error seems still to be present

Possommi (approved):
Tested with the latest commits and its working as expected.

PR #2106, opened 2024-03-15
MCR-3064 do not delete items if user is guest

yagee-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3064).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/frontend/MCRLayoutUtilities.java
+++ b/mycore-base/src/main/java/org/mycore/frontend/MCRLayoutUtilities.java
@@ -25,6 +25,7 @@ import java.io.File;
 import java.io.IOException;
 import java.io.UnsupportedEncodingException;
 import java.net.MalformedURLException;
+import java.net.URISyntaxException;
 import java.net.URL;
 import java.net.URLConnection;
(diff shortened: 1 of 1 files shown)


toKrause:
I vaguely remember that we wanted to change what happens in `MCRLayoutUtilities`, although I don't remember the exact plan.

But something doesn't add up here. According to the description of the JIRA-ticket the intended behavior is: "If an `@href` contains the placeholder `{CurrentUser}` and the user is not logged in, the `<item />` should be removed from the resulting personal navigation." Isn't that the exact behavior that this PR removes?

Wrong description or wrong implementation?

yagee-de:
> I vaguely remember that we wanted to change what happens in `MCRLayoutUtilities`, although I don't remember the exact plan.

This is the follow-up PR with the change as it was planned.

> 
> But something doesn't add up here. According to the description of the JIRA-ticket the intended behavior is: "If an `@href` contains the placeholder `{CurrentUser}` and the user is not logged in, the `<item />` should be removed from the resulting personal navigation." Isn't that the exact behavior that this PR removes?

Thanks, changed the ticket description so it reflect the changes.

toKrause (approved):

PR #2107, opened 2024-03-15
MCR-3067 reuse JVM forks for junit tests

yagee-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3067).


mycore-base/src/main/java/org/mycore/common/xml/MCRURIResolver.java
```diff
@@ -155,6 +154,11 @@ public final class MCRURIResolver implements URIResolver {
      * Creates a new MCRURIResolver
      */
     private MCRURIResolver() {
+        reInit();
```
Mewel: move reInit() to static {} block and remove constructor.

mycore-base/src/main/java/org/mycore/common/events/MCRShutdownHandler.java
```diff
@@ -159,6 +161,30 @@ void runClosables() {
         }
     }
 
+    /**
+     * only used to make {@link MCRShutdownHandler#resetCloseables()} testable
```
Mewel: Javadoc comment is ambiguous. It should explain that this method (and the following) should only be called by a test framework and are not for applications to use.

Mewel (changes requested):

mycore-base/src/main/java/org/mycore/common/xml/MCRURIResolver.java
yagee-de: is fixed

mycore-base/src/main/java/org/mycore/common/events/MCRShutdownHandler.java
yagee-de: is fixed

Mewel (dismissed):

sebhofmann:
I think its hard or almost impossible to implement a "reset" of mycore. The configured instances in MCRConfiguration2#instanceHolder do not get resetted and there are many other place e.g. singletons, which do not get resetted. The Tests are now unstable and they get more unstable with this pr. Wouldn't it be a better approach to run the tests parallel to increase the build performance?

mycore-base/src/main/java/org/mycore/common/events/MCRShutdownHandler.java
```diff
@@ -159,6 +161,38 @@ void runClosables() {
...
+
+    /**
+     * Resets the {@link Closeable} requests in the shutdown handler.
+     * This method checks if there is a previous state ({@link #mark()}) and, if present, resets the shuttingDown
+     * and requests fields to their values from the mark.
+     * The operation is performed under a write lock to ensure thread safety.
+     * <p>
+     * <b>Warning:</b> It should be used only in unit tests to test this class while keeping a clean state.
+     */
+    protected void resetCloseables() {
```
Mewel: Should be inside of junit.

Mewel (changes requested):

Mewel (dismissed):

Mewel (approved):

sebhofmann (approved):

PR #2115, opened 2024-03-27
MCR-3053 reverted unintended changes in MCRCategoryDAOFactory

sebhofmann:

### DIFF
--- a/mycore-base/src/main/java/org/mycore/datamodel/classifications2/MCRCategoryDAOFactory.java
+++ b/mycore-base/src/main/java/org/mycore/datamodel/classifications2/MCRCategoryDAOFactory.java
@@ -21,6 +21,7 @@ package org.mycore.datamodel.classifications2;
 import java.util.Objects;
 
 import org.mycore.common.config.MCRConfiguration2;
+import org.mycore.datamodel.classifications2.impl.MCRCategoryDAOImpl;
 
 /**
  * @author Thomas Scheffler (yagee)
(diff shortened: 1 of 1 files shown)


mycore-base/src/main/java/org/mycore/datamodel/classifications2/MCRCategoryDAOFactory.java
```diff
@@ -50,8 +51,9 @@ public static synchronized void set(Class<? extends MCRCategoryDAO> daoClass) th
     // encapsulate the INSTANCE in an inner static class to avoid issues with class loading order
     // this is known as "Bill Pugh singleton"
     private static class MCRCategoryDAOHolder {
-        private static MCRCategoryDAO INSTANCE = MCRConfiguration2.getInstanceOfOrThrow(
-            MCRCategoryDAO.class, "MCR.Category.DAO");
+        private static MCRCategoryDAO INSTANCE = MCRConfiguration2.getInstanceOf(
+            MCRCategoryDAO.class, "MCR.Category.DAO")
+            .orElseGet(MCRCategoryDAOImpl::new);
```
toKrause, line 56: this PR makes no sense. the whole point of MCR-3053 is to get rid of fallback definition in Java code. `MCR.Category.DAO` should have a default value in the properties instead.

This was in fact done as part of MCR-3053: https://github.com/MyCoRe-Org/mycore/blob/c774afbfe8303fdda98a4bd9c309d111347a7409/mycore-base/src/main/resources/config/mycore.properties#L651

toKrause (changes requested):

PR #2121, opened 2024-04-11
MCR-3078 Remove unused redundant property

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3078).


rsteph-de (approved):
solved as discussed on last telko

PR #2124, opened 2024-04-12
MCR-3079 fix property check

erodde:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3079).

In the current implementation MCRPIService.requireNotEmptyProperty throws an NPE when a property is not set:

java.lang.NullPointerException: Cannot invoke "String.length()" because the return value of "java.util.Map.get(Object)" is null
[INFO] [talledLocalContainer] 	at org.mycore.pi.MCRPIService.requireNotEmptyProperty(MCRPIService.java:502)

 
It is not an urgent fix, as a missing property is supposed to throw an exception anyway. But the NPE obscures the intended error message.


sebhofmann:
Good catch, can you rebase this to 2023?

sebhofmann (approved):

toKrause (approved):

PR #2129, opened 2024-04-19
MCR-3085 Upgrade citeproc-java to version 3.1.0 (supports new variable citation-key)

Possommi:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3085).


kkrebs (approved):

kkrebs (changes requested):
we should wait a moment, because we have found another citeproc-java bug. perhaps this will be fixed fast so that we can include this here as well
(and test builds doesn't work atm ...)

kkrebs (approved):

PR #2138, opened 2024-04-24
MCR-3073 Use local/configurable Javascript source for vis network library

michael-becker:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3073).


sebhofmann (changes requested):
We should not include third party library dependencies as minified js. 
There is a webjar:
```
<dependency>
    <groupId>org.webjars.npm</groupId>
    <artifactId>vis-network</artifactId>
    <version>9.1.2</version>
</dependency>
```

rsteph-de (approved):
looks good for me - "correct" use of webjars!

sebhofmann (approved):

PR #2142, opened 2024-04-25
MCR-3093 new enrichment command

erodde:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3093).


fluetze (changes requested):
Pls use MCRMetadataManager.update() instead of fireUpdateEvent().

fluetze (approved):
Looks good for me

PR #2143, opened 2024-04-25
MCR-3053 fix instantiation of MCROAISolrSetHandler

kkrebs:

comes from https://github.com/MyCoRe-Org/mycore/commit/ec5de0a8ec4faf998b7e28ad2ddc7d4f6b7ef36a#diff-63774fe31cf1f8d417b64d01680d7715211199cbb259f3e5699188a4c040b81e

### DIFF
--- a/mycore-oai/src/main/java/org/mycore/oai/set/MCROAISolrSetConfiguration.java
+++ b/mycore-oai/src/main/java/org/mycore/oai/set/MCROAISolrSetConfiguration.java
@@ -52,6 +52,7 @@ public class MCROAISolrSetConfiguration implements MCROAISetConfiguration<SolrQu
                 }
                 return MCRConfiguration2.instantiateClass(fallbackHandlerClassName);
             });
+        handler.init(configPrefix, setId);
         this.id = setId;
         this.uri = getURI(setConfigPrefix);
         this.handler = handler;
--- a/mycore-oai/src/main/java/org/mycore/oai/set/MCROAISolrSetConfiguration.java
+++ b/mycore-oai/src/main/java/org/mycore/oai/set/MCROAISolrSetConfiguration.java
@@ -43,8 +43,15 @@ public class MCROAISolrSetConfiguration implements MCROAISetConfiguration<SolrQu
     public MCROAISolrSetConfiguration(String configPrefix, String setId) {
         String setConfigPrefix = configPrefix + SETS_PREFIX + setId;
         MCROAISolrSetHandler handler = MCRConfiguration2.getInstanceOf(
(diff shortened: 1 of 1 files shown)


mycore-oai/src/main/java/org/mycore/oai/set/MCROAISolrSetConfiguration.java
```diff
@@ -44,8 +44,7 @@ public MCROAISolrSetConfiguration(String configPrefix, String setId) {
         String setConfigPrefix = configPrefix + SETS_PREFIX + setId;
         MCROAISolrSetHandler handler = MCRConfiguration2.getInstanceOf(
                 MCROAISolrSetHandler.class, setConfigPrefix + ".Handler")
-            .orElseGet(() -> MCRConfiguration2.getInstanceOfOrThrow(
-                MCROAISolrSetHandler.class, getFallbackHandler(configPrefix, setId)));
+            .orElseGet(() -> MCRConfiguration2.instantiateClass(getFallbackHandler(configPrefix, setId)));
```
toKrause: The fallback handler might be be `null`, support for which was unintentionally dropped in MCR-3053.

Maybe:

```java
MCROAISolrSetHandler handler = MCRConfiguration2
    .getInstanceOf(MCROAISolrSetHandler.class, setConfigPrefix + ".Handler")
    .orElseGet(() -> {
        String fallbackHandler = getFallbackHandler(configPrefix, setId);
        if (fallbackHandler == null) {
            throw new MCRConfigurationException("Configuration property " +  setConfigPrefix 
                + ".Handler is not set and no fallback handler could be determined");
        }
        return MCRConfiguration2.instantiateClass(fallbackHandler);
    });
```

toKrause (changes requested):

mycore-oai/src/main/java/org/mycore/oai/set/MCROAISolrSetConfiguration.java
rsteph-de: changed as suggested

toKrause (dismissed):

sebhofmann:
The .init method call is gone: https://github.com/MyCoRe-Org/mycore/commit/ec5de0a8ec4faf998b7e28ad2ddc7d4f6b7ef36a#diff-63774fe31cf1f8d417b64d01680d7715211199cbb259f3e5699188a4c040b81eL51

kkrebs (commented):
works now with current MIR main

rsteph-de:
the mix of calls to constructor and init() is a little bit confusing

toKrause (approved):

PR #2148, opened 2024-05-02
MCR-3096 prepare for ocfl

Mewel:
-Support different digest algorithms like SHA-512
-Extend NIOFS API
-Support Versioning Paths
-Improve Transaction Handling

[Link to jira](https://mycore.atlassian.net/browse/MCR-3096).


mycore-base/src/main/java/org/mycore/common/digest/MCRDigest.java
```diff
@@ -0,0 +1,111 @@
...
+
+/**
+ * Representing a message digest.
+ * This class is sealed to specific digest implementations, ensuring
+ * that only permitted classes can extend it.
+ *
+ * @see MCRMD5Digest
+ * @see MCRSHA512Digest
+ */
+public abstract sealed class MCRDigest permits MCRMD5Digest, MCRSHA512Digest {
```
sebhofmann: I don't see any reason why MCRDigest should not allow other Implementations.

mycore-base/src/main/java/org/mycore/common/digest/MCRMD5Digest.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    @Override
+    public Algorithm getAlgorithm() {
+        return Algorithm.MD5;
+    }
+
+    @Override
+    protected void validate() throws MCRDigestValidationException {
+        if (this.value.length() != 32 || !MD5_HEX_PATTERN.matcher(value).matches()) {
```
sebhofmann: Better use HexFormat and parseHex.

mycore-base/src/main/java/org/mycore/common/digest/MCRDigest.java
```diff
@@ -0,0 +1,111 @@
...
+            return name.toLowerCase(Locale.ROOT);
+        }
+
+        public String toUpperCase() {
+            return name.toUpperCase(Locale.ROOT);
+        }
+
+    }
+
+    protected String value;
```
sebhofmann: A digest should be bytes and there should be a toHexString method.

mycore-base/src/main/java/org/mycore/datamodel/niofs/MCRAbstractFileSystem.java
```diff
@@ -144,14 +146,14 @@ public boolean verifies(MCRPath path, MCRFileAttributes<?> attrs) throws NoSuchF
...
-            md5Sum = MCRUtils.getMD5Sum(Files.newInputStream(path));
+            digest = MCRUtils.getDigest(attrs.digest().getAlgorithm(), Files.newInputStream(path));
         } catch (IOException e) {
             LogManager.getLogger(getClass()).error("Could not verify path: {}", path, e);
             return false;
         }
-        boolean returns = md5Sum.matches(attrs.md5sum());
+        boolean returns = digest.matches(attrs.digest().getValue());
         if (!returns) {
             LogManager.getLogger(getClass()).warn("MD5sum does not match: {}", path);
```
sebhofmann: Should use digest algorithm here and not asume it is md5 (in the log line)

mycore-base/src/main/java/org/mycore/datamodel/niofs/MCRAbstractFileSystemProvider.java
```diff
@@ -0,0 +1,178 @@
...
+    /* (non-Javadoc)
+     * @see java.nio.file.spi.FileSystemProvider#getPath(java.net.URI)
+     */
+    @Override
+    public Path getPath(final URI uri) {
+        PathInformation pathInfo = getPathInformation(uri);
+        return getPath(pathInfo.owner(), pathInfo.path(), getFileSystem());
+    }
+
+    public MCRPath getPath(String owner, String path, MCRAbstractFileSystem fs) {
```
sebhofmann: Is this really required? It can default to getFileSystem()

mycore-base/src/main/java/org/mycore/datamodel/niofs/MCRBasicFileAttributeViewProperties.java
```diff
@@ -0,0 +1,115 @@
...
+
+    private static final String IS_DIRECTORY_NAME = "isDirectory";
+
+    private static final String IS_REGULAR_FILE_NAME = "isRegularFile";
+
+    private static final String IS_SYMBOLIC_LINK_NAME = "isSymbolicLink";
+
+    private static final String IS_OTHER_NAME = "isOther";
+
+    public static final HashSet<String> ALLOWED_ATTRIBUTES = Sets.newHashSet(ALL, SIZE_NAME, CREATION_TIME_NAME,
```
sebhofmann: Should be immutable.

mycore-base/src/main/java/org/mycore/datamodel/niofs/MCRBasicFileAttributeViewProperties.java
```diff
@@ -0,0 +1,115 @@
...
+import java.util.Objects;
+import java.util.Set;
+
+import com.google.common.collect.Sets;
+
+public class MCRBasicFileAttributeViewProperties<V extends BasicFileAttributeView> {
+
+    private static final String ALL = "*";
+
+    private static final String SIZE_NAME = "size";
```
sebhofmann: The attributes should be an enum.

mycore-base/src/main/java/org/mycore/datamodel/niofs/MCRBasicFileAttributesProxy.java
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    @Override
+    public MCRDigest digest() {
+        return proxy().digest();
+    }
+
+    protected MCRFileAttributes<T> proxy() {
+        if (this.proxy == null) {
+            synchronized (lock) {
```
sebhofmann: The synchronized should use this.proxy and lock should be removed.

mycore-base/src/main/java/org/mycore/datamodel/niofs/MCRDefaultFileAttributes.java
```diff
@@ -0,0 +1,147 @@
...
+            lastAccessTime);
+    }
+
+    public static MCRFileAttributes fromAttributes(BasicFileAttributes attrs, MCRDigest digest) {
+        return new MCRDefaultFileAttributes<>(FileType.fromAttribute(attrs), attrs.size(), attrs.fileKey(), digest,
+            attrs.creationTime(), attrs.lastModifiedTime(), attrs.lastAccessTime());
+    }
+
+    /* (non-Javadoc)
+     * @see java.nio.file.attribute.BasicFileAttributes#creationTime()
```
sebhofmann: I think its better to use @inheritDoc

mycore-base/src/main/java/org/mycore/common/digest/MCRDigest.java
```diff
@@ -0,0 +1,111 @@
...
+ *
+ * @see MCRMD5Digest
+ * @see MCRSHA512Digest
+ */
+public abstract sealed class MCRDigest permits MCRMD5Digest, MCRSHA512Digest {
+
+    /**
+     * Enumeration of supported digest algorithms.
+     */
+    public enum Algorithm {
```
sebhofmann: This enum is okay to avoid magic strings, but i would not make it a requirement for MCRDigest.

mycore-base/src/main/java/org/mycore/datamodel/niofs/MCRVersionedFileSystemProvider.java
```diff
@@ -0,0 +1,232 @@
...
+            return false;
+        }
+        MCRVersionedPath versionedPath1 = MCRVersionedPath.toVersionedPath(path1);
+        MCRVersionedPath versionedPath2 = MCRVersionedPath.toVersionedPath(path2);
+        return Objects.equals(versionedPath1, versionedPath2);
+    }
+
+    @Override
+    public boolean isHidden(Path path) {
+        MCRVersionedPath.toVersionedPath(path);
```
sebhofmann: Is this required?

mycore-restapi/src/main/java/org/mycore/restapi/v1/utils/MCRJSONFileVisitor.java
```diff
@@ -104,7 +104,7 @@ private void addBasicAttributes(Path path, BasicFileAttributes attrs) throws IOE
         if (attrs.isRegularFile()) {
             jw.name("contentType").value(MCRContentTypes.probeContentType(path));
             if (attrs instanceof MCRFileAttributes<?> fileAttrs) {
-                jw.name("md5").value(fileAttrs.md5sum());
+                jw.name("md5").value(fileAttrs.digest().getValue());
```
sebhofmann: I think its ok to deliver a sha512 hash as md5, because the api would be more stable for old applications. Alternative would be using the digest algo name as key.

mycore-restapi/src/main/java/org/mycore/restapi/v2/MCREventHandler.java
```diff
@@ -287,7 +287,7 @@ public void sendEvent(MCREvent evt, Path path, BasicFileAttributes attrs) {
             file.addProperty("object", objId);
             file.addProperty("size", attrs.size());
             file.addProperty("modified", attrs.lastModifiedTime().toInstant().toString());
-            file.addProperty("md5", ((MCRFileAttributes) attrs).md5sum());
+            file.addProperty("md5", ((MCRFileAttributes) attrs).digest().getValue());
```
sebhofmann: I think its ok to deliver a sha512 hash as md5, because the api would be more stable for old applications. Alternative would be using the digest algo name as key.

sebhofmann (changes requested):

mycore-base/src/main/java/org/mycore/datamodel/niofs/MCRAbstractFileSystemProvider.java
Mewel: @yagee-de Can you elaborate on this thomas? Should we add a FileSystem member variable for MCRPath to avoid the anonymous class?

mycore-base/src/main/java/org/mycore/datamodel/niofs/MCRBasicFileAttributesProxy.java
Mewel: Would lead to: synchronization on a non-final field 'this.proxy'

mycore-base/src/main/java/org/mycore/datamodel/niofs/MCRAbstractFileSystemProvider.java
yagee-de: MCRAbstractFileSystemProvider already has a method "getFileSystem()" It should be used instead of a parameter in getPath(). Makes little sense to construct an MCRPath in FileSystem "A" with the Implementation of FileSystem "B".

sebhofmann (dismissed):

mycore-base/src/main/java/org/mycore/common/digest/MCRMD5Digest.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    @Override
+    protected void validate() throws MCRDigestValidationException {
+        if (this.value.length != 16) {
+            throw new MCRDigestValidationException("Not a valid digest: " + toHexString());
+        }
+    }
+
+    public static class Md5Algorithm extends Algorithm {
```
yagee-de: naming: should be MD5… not Md5…

mycore-base/src/main/java/org/mycore/datamodel/niofs/MCRBasicFileAttributeViewProperties.java
```diff
@@ -0,0 +1,195 @@
...
+    /**
+     * Set of allowed attributes that can be accessed.
+     */
+    public static final Set<String> ALLOWED_ATTRIBUTES;
+
+    static {
+        ALLOWED_ATTRIBUTES = Arrays.stream(Attribute.values())
+            .map(Attribute::getName)
+            .collect(Collectors.toSet());
+    }
```
yagee-de: initialize when declerating ALLOWED_ATTRIBUTES

mycore-base/src/main/java/org/mycore/datamodel/niofs/MCRBasicFileAttributeViewProperties.java
```diff
@@ -0,0 +1,195 @@
...
+        ALL("*"),
+        SIZE_NAME("size"),
+        CREATION_TIME_NAME("creationTime"),
+        LAST_ACCESS_TIME_NAME("lastAccessTime"),
+        LAST_MODIFIED_TIME_NAME("lastModifiedTime"),
+        FILE_KEY_NAME("fileKey"),
+        IS_DIRECTORY_NAME("isDirectory"),
+        IS_REGULAR_FILE_NAME("isRegularFile"),
+        IS_SYMBOLIC_LINK_NAME("isSymbolicLink"),
+        IS_OTHER_NAME("isOther");
```
yagee-de, line 52: Isn't really used as enum anywhere. It's extended in sub classes and always converted to String or from String

mycore-base/src/main/java/org/mycore/datamodel/niofs/MCRBasicFileAttributesProxy.java
```diff
@@ -0,0 +1,98 @@
...
+                if (this.proxy == null) {
+                    this.proxy = loadProxy();
```
yagee-de: if loadProxy() fails every call to it will result in a new request to loadProxy(). So it will not load once but many times and can result in unforeseen exceptions.

mycore-base/src/main/java/org/mycore/datamodel/niofs/MCRBasicFileAttributesProxy.java
```diff
@@ -0,0 +1,98 @@
...
+
+    private volatile MCRFileAttributes<T> proxy;
+
+    private final Object lock = new Object();
+
+    protected final P path;
+
+    public MCRBasicFileAttributesProxy(P path) {
+        this.path = path;
+    }
```
yagee-de: This is not how attributes work. Call to the methods should not result in exceptions once they are constructed:
`Files.readAttributes(file, MyFileAttributes.class)` should throw any exception and not a call to lastAccessTime(). Often a check for existence of a file is implemented as reading the file attributes.

mycore-base/src/main/java/org/mycore/frontend/fileupload/MCRUploadHandlerIFS.java
```diff
@@ -199,7 +199,7 @@ public boolean acceptFile(String path, String checksum, long length) throws Exce
                 @SuppressWarnings("rawtypes")
                 MCRFileAttributes attrs = Files.readAttributes(child, MCRFileAttributes.class);
                 shouldAcceptFile = attrs.size() != length
-                    || !(checksum.equals(attrs.md5sum()) && child.getFileSystem().verifies(child, attrs));
+                    || !(checksum.equals(attrs.digest()) && child.getFileSystem().verifies(child, attrs));
```
yagee-de: String.equals(MCRDigest) should always be false

yagee-de (changes requested):

mycore-base/src/main/java/org/mycore/datamodel/niofs/MCRBasicFileAttributeViewProperties.java
Mewel: Any reason why you don't like static blocks?

mycore-base/src/main/java/org/mycore/datamodel/niofs/MCRBasicFileAttributeViewProperties.java
Mewel: I just copied this code from the old class. What do you want to have?
```
private static final String ALL = "*";
...
```

mycore-base/src/main/java/org/mycore/datamodel/niofs/MCRBasicFileAttributesProxy.java
Mewel: Ok, will rm this class entirely and solve the attribute loading in ocfl

mycore-base/src/main/java/org/mycore/datamodel/niofs/MCRBasicFileAttributeViewProperties.java
yagee-de: I <3 static blocks! Use of static blocks is only required when catching an exception is required. Usually it's best to initialize static fields right away.

yagee-de (approved):

PR #2150, opened 2024-05-07
 MCR-3087 Allow selection of XSLT Processor when using MCRXslStyleResolver

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3087).


yagee-de (approved):
Looks good and seems to be a good solution to clean up after tests

PR #2151, opened 2024-05-08
MCR-3098 Limit length of short description to 5000 characters

Possommi:
[MCR-3098](https://mycore.atlassian.net/browse/MCR-3098).

[MCR-3098]: https://mycore.atlassian.net/browse/MCR-3098?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mycore-orcid2/src/main/resources/xsl/orcid2/v3/mcr2work_generic.xsl
+++ b/mycore-orcid2/src/main/resources/xsl/orcid2/v3/mcr2work_generic.xsl
@@ -88,7 +88,7 @@
 
   <xsl:template match="mods:abstract">
     <work:short-description>
-      <xsl:value-of select="mcrstring:shorten(text(), $short-description-max-length, '…')"/>
+      <xsl:value-of select="mcrstring:shorten(text(), ($short-description-max-length - 1), '…')"/>
     </work:short-description>
   </xsl:template>
 
--- a/mycore-orcid2/src/main/resources/xsl/orcid2/v3/mcr2work_generic.xsl
+++ b/mycore-orcid2/src/main/resources/xsl/orcid2/v3/mcr2work_generic.xsl
@@ -4,12 +4,14 @@
   xmlns="http://www.w3.org/TR/REC-html40"
   xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
   xmlns:fn="http://www.w3.org/2005/xpath-functions"
+  xmlns:mcrstring="http://www.mycore.de/xslt/stringutils"
   xmlns:mods="http://www.loc.gov/mods/v3"
   xmlns:work="http://www.orcid.org/ns/work"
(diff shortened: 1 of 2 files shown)


mycore-orcid2/src/main/resources/xsl/orcid2/v3/mcr2work_generic.xsl
```diff
@@ -84,7 +86,14 @@
 
   <xsl:template match="mods:abstract">
     <work:short-description>
-      <xsl:value-of select="text()" />
+      <xsl:choose>
```
toKrause: should make use of `mcrstring:shorten(string, int)` (defined in `xslt/functions/stringutils.xsl`, namespace: `http://www.mycore.de/xslt/stringutils`) instead of reinventing the wheel. That method, for example, takes better care of placing the ellipsis at word eindings.

toKrause (changes requested):

toKrause (approved):

golsch (changes requested):
It doesn't look right to me. The edge case appears to be critical.
`mcrstring:shorten('123456', 5, '…')` => 12345…
In General, ellipsis seems to me to be added after the permitted length. The resulting length would then be n + 1 and not n.
But `mcrstring:shorten('12345', 5-1, '…')` => 1234…  doesn't seem so elegant either.

Possommi:
> It doesn't look right to me. The edge case appears to be critical. `mcrstring:shorten('123456', 5, '…')` => 12345… In General, ellipsis seems to me to be added after the permitted length. The resulting length would then be n + 1 and not n. But `mcrstring:shorten('12345', 5-1, '…')` => 1234… doesn't seem so elegant either.

I can rollback to the previous version which works and doesnt require an extra include.

toKrause:
Using an existing utility function still seem right to me, but, apparently, that function shortens the text to the desired length _and then_ adds the ellipsis, requiring a call like `<xsl:value-of select="mcrstring:shorten(text(), $short-description-max-length - 1, '…')"/>` (which, imho, is a questionable design decision regarding that utility function that might warrant a separate issue to be opened).

golsch (commented):
This would solve the problem on the part of ORCID. However, I would also question the behavior of `mcrstring:shorten`. If an ellipsis is specified, I would expect the final length to be equal to the passed length including the ellipsis.

Possommi:
> This would solve the problem on the part of ORCID. However, I would also question the behavior of `mcrstring:shorten`. If an ellipsis is specified, I would expect the final length to be equal to the passed length including the ellipsis.

But that could be an extra ticket to speed up the merge of this ticket

yagee-de (approved):
looks good and reasonable to accomplish to goal

PR #2169, opened 2024-06-05
MCR-3121 Improve session related logging and information display

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3121).

### DIFF
--- a/mycore-webtools/src/main/resources/META-INF/resources/modules/webtools/session/js/sessionListing.js
+++ b/mycore-webtools/src/main/resources/META-INF/resources/modules/webtools/session/js/sessionListing.js
@@ -145,6 +145,11 @@ mycore.session.listing = {
   },
 
   showStacktrace: function(id) {
+    var escapeHtml = function(text) {
+      let element = document.createElement('div');
+      element.textContent = text;
+      return element.innerHTML;
+    }
     var dialog = $("#stacktraceModalBody");
     dialog.empty();
     var session = mycore.session.listing.getSession(id);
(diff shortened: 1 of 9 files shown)


mycore-webtools/src/main/resources/META-INF/resources/modules/webtools/session/js/sessionListing.js
```diff
@@ -142,12 +142,25 @@ mycore.session.listing = {
     var session = mycore.session.listing.getSession(id);
     var stacktrace = session.constructingStacktrace.stacktrace;
     var header = dialog.parents(".modal-dialog").find(".modal-header");
+    var html = "ID: " + session.id;
```
sebhofmann: The whole function is very dangerous and allows to inject javascript code in to the session listing. Example:
```
curl -H "User-Agent: <script>alert('test')</script>" http://localhost:8291/mir/content/index.xml
```
Open the http://localhost:8291/mir/modules/webtools/session/sessionListing.xml as admin and click the colored item of the right session and you get a funny message box. 


To avoid this you need to escape the content. 
```
    let cleanerElement = document.createElement('div');
    let badText = '<script>alert("XSS");</script>'
    cleanerElement.textContent = badText;
    let cleanText = badText.innerHTML;
```

sebhofmann (changes requested):

mycore-webtools/src/main/resources/META-INF/resources/modules/webtools/session/js/sessionListing.js
toKrause: You're absolutely right. Will fix.

sebhofmann (dismissed):

yagee-de (approved):

PR #2174, opened 2024-06-10
MCR-3125 Set max length to MCRJob parameter value to 16384

Possommi:
[MCR-3125](https://mycore.atlassian.net/browse/MCR-3125).

[MCR-3125]: https://mycore.atlassian.net/browse/MCR-3125?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mycore-jobqueue/src/main/java/org/mycore/services/queuedjob/MCRJob.java
+++ b/mycore-jobqueue/src/main/java/org/mycore/services/queuedjob/MCRJob.java
@@ -245,7 +245,7 @@ public class MCRJob implements Cloneable {
     @ElementCollection(fetch = FetchType.EAGER)
     @CollectionTable(name = "MCRJobParameter", joinColumns = @JoinColumn(name = "jobID"))
     @MapKeyColumn(name = "paramKey", length = 128)
-    @Column(columnDefinition = "TEXT")
+    @Column(name="paramValue", length = 16384)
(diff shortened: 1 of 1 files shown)


sebhofmann (approved):

kkrebs (changes requested):
need migration step, so this should be a change in main branch

yagee-de:
MS SQL Server
> The ntext, text, and image data types will be removed in a future version of SQL Server. Avoid using these data types in new development work, and plan to modify applications that currently use them.

– [Source](https://learn.microsoft.com/en-us/sql/t-sql/data-types/ntext-text-and-image-transact-sql?view=sql-server-ver16)

yagee-de:
Type „TEXT“ is also not part of SQL or JDBC standard – but present in many DBMS. So I would prefer to make reasonable big VARCHAR out of it. "columDefinition" also makes this not portable as this definition is used on every DBMS.

toKrause (approved):

toKrause (approved):

kkrebs (approved):

kkrebs:
Please add migration documentation to https://www.mycore.de/documentation/migrate/migrate_mcr2024_06/

PR #2175, opened 2024-06-10
MCR-3126 OCFL filesystem

Mewel:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3126).


sebhofmann:
I found a conceptional Problem, which leads to errors in the tiling process. The MCRTreeCopier commits and restarts database transactions. The tiling queue runs in another thread is now able to find the commited tiling jobs and now starts tiling. The ocfl transaction of the Thread in with the MCRTreeCopier did not yet commit its OCFL Transaction, so the files which should be tiled do not exist yet.
We introduced Transaction commits in the MCRTreeCopier, so that in a large upload the connection to the database is released. In my opinion the cleanest aproach would be to remove the transaction restarts from the MCRTreeCopier, to always keep the file transaction and the database transaction in sync and also prevent weird rollback errors.

Meanwhile the uploads happen in another transaction and are stored in a local directory, so the "long" transaction only runs long to upload from the mycore machine to s3, or to copy from one local to another local directory, which should not take that long.

mycore-ocfl/src/main/java/org/mycore/ocfl/metadata/MCROCFLXMLMetadataManagerAdapter.java
```diff
@@ -338,23 +344,28 @@ public IntStream getStoredIDs(String project, String type) throws MCRPersistence
     public int getHighestStoredID(String project, String type) {
         int highestStoredID = 0;
         int maxDepth = Integer.MAX_VALUE;
-        MCROCFLRepositoryProvider oclfRepoProvider = MCRConfiguration2.getSingleInstanceOfOrThrow(
-            MCROCFLRepositoryProvider.class, "MCR.OCFL.Repository." + repositoryKey);
 
-        OcflExtensionConfig config = oclfRepoProvider.getExtensionConfig();
+        MCROCFLRepositoryProvider ocflRepoProvider = MCROCFLRepositoryProvider.getProvider(repositoryKey);
+        if (!(ocflRepoProvider instanceof MCROCFLLocalRepositoryProvider localRepositoryProvider)) {
```
sebhofmann, line 346: This should still be possible, otherwise the user needs to rebuild the id cache by hand, which is a impossible task with the default layout.

sebhofmann (dismissed):

mycore-ocfl/src/main/java/org/mycore/ocfl/metadata/MCROCFLXMLMetadataManagerAdapter.java
Mewel: I don't think it's a good idea to deactivate this check. I would rather have a migration command which builds the cache. Otherwise the remote repo is not useable anyway.

mycore-ocfl/src/test/java/org/mycore/ocfl/metadata/migration/MCROCFLCombineIgnoreXPathPrunerTest.java
```diff
@@ -28,16 +31,22 @@
 import org.jdom2.Document;
 import org.jdom2.Element;
 import org.jdom2.JDOMException;
-import org.junit.Assert;
-import org.junit.Test;
-import org.mycore.common.MCRTestCase;
+import org.junit.jupiter.api.Test;
```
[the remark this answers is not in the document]
Mewel: This is incorrect.

mycore-ocfl/src/main/java/org/mycore/ocfl/metadata/MCROCFLXMLMetadataManagerAdapter.java
sebhofmann: There is a command which builds the cache, but it uses this method. The method should not be called anywhere, except in the command. All other places should use the cache or the database.

Mewel: Lets talk about this tomorrow, cause this method works directly with the Path API. It wants to call getRepositoryRoot(). This is not available on a remote S3 repository. There is no root path.

So either this method is wrongly implemented and shouldn't work with Path's or you would need to migrate to the MCRFileBaseCacheObjectIDGenerator first in order to use OCFL S3 repos.

mycore-ocfl/src/main/java/org/mycore/ocfl/repository/MCROCFLRepositoryProvider.java
```diff
@@ -19,24 +19,66 @@
...
+     * <p>
+     * This method uses MyCoRe's configuration system to dynamically locate and return the provider
+     * instance matching the given repository ID.
+     * </p>
+     *
+     * @param id the unique identifier for the desired repository provider.
+     * @return the {@link MCROCFLRepositoryProvider} instance for the specified ID.
+     * @throws MCRConfigurationException if no provider is configured for the specified ID.
+     */
+    static MCROCFLRepositoryProvider getProvider(String id) {
```
toKrause: Name `obtainInstance` would be more in line with the rest of MyCoRe.

mycore-ocfl/pom.xml
```diff
@@ -63,11 +87,48 @@
       <groupId>org.mycore</groupId>
       <artifactId>mycore-user2</artifactId>
     </dependency>
+    <dependency>
+      <groupId>software.amazon.awssdk</groupId>
```
toKrause: Presumably, the dependencies for S3-support could be marked as optional. 

Not everyone who wishes to use OCFL is going to also use S3. Theese dependecies could therefore be keept off the classpath for, presumably, a lot of users.

As far as I  can see, S3-supoport is implementes soley in `MCROCFLS3RepositoryProvider`.
Have you consideres moving S3-support into a separate module `mycore-ocfl-s3` that contains only that class, the corresponding entries in `mycore.properties` and the corresponding dependencies in `pom.xml`.

Not a "must", but would be nice.

toKrause (dismissed):

PR #2189, opened 2024-06-20
MCR-3132 mycore-iview2: deprecations and refactoring

yagee-de:
mark MCR*Servlet as @Deprecated, use IIIF instead

[Link to jira](https://mycore.atlassian.net/browse/MCR-3132).


rsteph-de (approved):
looks good for me

PR #2201, opened 2024-07-10
MCR-3140 Provide means to delete old entries from job queue

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3140).


rsteph-de (approved):
looks good for me

PR #2202, opened 2024-07-12
MCR-3141 Property-defined patterns for xpath-mapping

erodde:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3141).

### DIFF
--- a/mycore-mods/src/main/java/org/mycore/mods/classification/MCRClassificationMappingEventHandler.java
+++ b/mycore-mods/src/main/java/org/mycore/mods/classification/MCRClassificationMappingEventHandler.java
@@ -84,7 +84,7 @@ public class MCRClassificationMappingEventHandler extends MCREventHandlerBase {
     private static final String X_PATH_MAPPING_CLASSIFICATIONS
         = MCRConfiguration2.getString("MCR.Category.XPathMapping.ClassIDs").orElse("");
 
-    private static final Map<String, String> mappingPatterns = MCRConfiguration2.getSubPropertiesMap(
+    private static final Map<String, String> MAPPING_PATTERNS = MCRConfiguration2.getSubPropertiesMap(
         "MCR.Category.XPathMapping.Pattern.");
 
     private static final Logger LOGGER = LogManager.getLogger(MCRClassificationMappingEventHandler.class);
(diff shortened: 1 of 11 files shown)


erodde:
@fluetze: I still need to implement the corresponding mechanism for the mods-independent x-path-mapping. But could you have a look at this and tell me if this is what you had in mind?

mycore-mods/src/main/java/org/mycore/mods/classification/MCRClassificationMappingEventHandler.java
```diff
@@ -151,6 +157,45 @@ private static String getXPathMappingGenerator(MCRCategoryID target) {
...
+     * (when there are no values, the "=" is optional)<p>
+     * Ex.:<p>
+     * <b>Input XPath:</b> {pattern:Genre=article} and not(mods:relatedItem[@type='host'])<p>
+     * <b>Property:</b> MCR.Category.XPathMapping.Pattern.Genre=mods:genre[substring-after(@valueURI,'#')='{0}']<p>
+     * <b>Substituted XPath:</b> mods:genre[substring-after(@valueURI,'#')='article']
+     * and not(mods:relatedItem[@type='host'])
+     * @param xPath the XPath containing a pattern to substitute
+     * @return the resolved xPath
+     */
+    private static String replacePattern(String xPath) {
```
rsteph-de, line 175: 1) With an improved Regex (more groups) you could further reduce the String operations:
https://regex101.com/r/7H2fcJ/3

2) You should support multiple occurrences: `while (matcher.finde())` should do the work

3) You may use the string positions provided by the regex matcher:
 `xpath = xpath.substring(0, matcher.start()) + placeholderText + xpath.substring(matcher.end())`

rsteph-de:
All in all an interesting feature ;-)
I'm still trying to figure out, if there are other use cases than the x-mapping in classifications.
Maybe we should also get feedback from other developers on our next Skype conference.
Would you prepare  some usage examples on a slide or add them to the confluence page.

mycore-mods/src/test/resources/MCRClassificationMappingEventHandlerTest/orcidWorkType.xml
fluetze, line 1: Could use multiple patterns here: another one for the host genre.
mods:relatedItem[@type='host']/mods:genre[substring-after(@valueURI,'#')='...']

fluetze (approved):
Ok, for me, but you might adapt the pattern syntax proposed in the last video conf.

toKrause (approved):

PR #2206, opened 2024-07-16
MCR-3146 Guard aginst unmodifiable resource URL list

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3146).


rsteph-de (approved):
good

PR #2209, opened 2024-07-23
MCR-3149 replace ckeditor with tinymce

Mewel:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3149).

### DIFF
--- a/mycore-wcms2/GruntFile.js
+++ b/mycore-wcms2/GruntFile.js
@@ -28,14 +28,6 @@ module.exports = function (grunt) {
                     files: [
                         {
                             expand: true,
-                            cwd: './node_modules/ckeditor',
-                            dest: './target/classes/META-INF/resources/modules/wcms2/node_modules/ckeditor',
-                            src: [
-                                '**'
-                            ]
-                        },
-                        {
-                            expand: true,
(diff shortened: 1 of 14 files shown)


mycore-wcms2/GruntFile.js
```diff
@@ -28,8 +28,16 @@ module.exports = function (grunt) {
...
-                            dest: './target/classes/META-INF/resources/modules/wcms2/node_modules/ckeditor',
+                            cwd: './node_modules/tinymce',
+                            dest: './target/classes/META-INF/resources/modules/wcms2/node_modules/tinymce',
+                            src: [
+                                '**'
+                            ]
+                        },
+                        {
+                            expand: true,
+                            cwd: './node_modules/tinymce-i18n/langs7',
```
sebhofmann, line 39: This seems to be wrong `/langs7',`

sebhofmann (changes requested):

mycore-wcms2/GruntFile.js
Mewel: This is actually correct. It refers to the tinyMCE version 7.

sebhofmann (approved):
I tested it and it works.

PR #2210, opened 2024-07-25
MCR-3150 Inconsistent/unideal selection of job statuses

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3150).


sebhofmann:
I think the code is fine since, the job will be reset to new, because the status is not max_tries yet.

toKrause:
Okay, that makes sense. At least in a development environment, I should probably set `MCR.QueuedJob.MaxTry=1` in order to be able to retry immediately.

toKrause:
One more adjacent question:

`MCRJobStaticContentGenerator` queries `List.of(MCRJobStatus.NEW, MCRJobStatus.ERROR, MCRJobStatus.PROCESSING)` in order to check, if an equal job already exists, but `MCRMODSMetadataShareAgent` only queries `List.of(MCRJobStatus.NEW, MCRJobStatus.ERROR)` for the same purpose. Shouldn't the letter also query `MCRJobStatus.PROCESSING`?

sebhofmann:
Yes they should be the same. I was not sure which one is better. If a Job is already processing it could also be finished with most of the work and the actual new changes are not done. So safest way would be to add the job even if there is already a processing one.

toKrause:
I've reused this PR for an actually useful change.

sebhofmann (approved):

PR #2226, opened 2024-08-14
MCR-3164 add custom permission resolver to rest auth filter

golsch:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3164).

### DIFF
--- a/mycore-restapi/src/main/java/org/mycore/restapi/v2/MCRRestAuthorizationFilter.java
+++ b/mycore-restapi/src/main/java/org/mycore/restapi/v2/MCRRestAuthorizationFilter.java
@@ -141,26 +141,10 @@ public class MCRRestAuthorizationFilter implements ContainerRequestFilter {
         if (HttpMethod.OPTIONS.equals(method)) {
             return;
         }
-        final String permission = Optional
-            .ofNullable(resourceInfo.getResourceMethod().getAnnotation(MCRRestRequiredPermission.class)).map(a -> {
-                final String requiredPermission = a.value();
-                final Class<? extends MCRRestRequiredPermissionResolver> requiredPermissionResolver = a.resolver();
-                if (requiredPermission.isEmpty()
-                    && MCRRestRequiredPermissionDefaultResolver.class.equals(requiredPermissionResolver)) {
-                    LogManager.getLogger("explicit value and resolver is not allowed");
-                    throw new InternalServerErrorException();
-                } else if (!requiredPermission.isEmpty()) {
-                    return requiredPermission;
-                } else {
-                    try {
-                        return requiredPermissionResolver.getDeclaredConstructor().newInstance()
-                            .resolvePermission(requestContext);
-                    } catch (Exception e) {
-                        LogManager.getLogger().error(e);
-                        throw new InternalServerErrorException();
-                    }
-                }
-            }).orElseGet(() -> getPermissionFromHttpMethod(method));
+        final String permission
+            = Optional.ofNullable(resourceInfo.getResourceMethod().getAnnotation(MCRRestRequiredPermission.class))
+                .map(a -> resolveRequiredPermission(a, requestContext))
+                .orElseGet(() -> getPermissionFromHttpMethod(method));
         Optional.ofNullable(resourceInfo.getResourceClass().getAnnotation(Path.class))
             .map(Path::value)
             .ifPresent(path -> {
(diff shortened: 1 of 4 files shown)


mycore-restapi/src/main/java/org/mycore/restapi/v2/MCRRestAuthorizationFilter.java
```diff
@@ -138,9 +141,26 @@ public void filter(ContainerRequestContext requestContext) {
...
+                    throw new InternalServerErrorException();
+                } else if (!requiredPermission.isEmpty()) {
+                    return requiredPermission;
+                } else {
+                    try {
+                        return requiredPermissionResolver.getDeclaredConstructor().newInstance()
+                            .resolvePermission(requestContext);
+                    } catch (Exception e) {
+                        LogManager.getLogger().error(e);
+                        throw new InternalServerErrorException();
```
toKrause: Should be `new InternalServerErrorException(e);` so the original exception won't get lost where it is eventually logged/evaluated.

toKrause (changes requested):

mycore-restapi/src/main/java/org/mycore/restapi/v2/MCRRestAuthorizationFilter.java
Possommi: 159 `LogManager.getLogger().error(e);` should be removed since the exception is thrown

mycore-restapi/src/main/java/org/mycore/restapi/v2/annotation/MCRRestRequiredPermission.java
```diff
@@ -38,5 +41,13 @@
...
-    String value();
+    String value() default "";
+
+    /**
+     * Resolves the required permission.
+     *
+     * @return the permission value
+     */
+    Class<? extends MCRRestRequiredPermissionResolver> resolver()
+        default MCRRestRequiredPermissionDefaultResolver.class;
```
yagee-de, line 52: This annotation is on a given endpoint, how does it have to resolve a permission (a.k.a. privilege) from the request? Can you elaborate a bit more on the use case?
For example if I write on a PUT endpoint @MCRRequiredPermission("myWritePerm") that's one permission I would want there. Where comes additional usage from the dynamic part in here?

mycore-restapi/src/main/java/org/mycore/restapi/v2/access/MCRRestRequiredPermissionResolver.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+public interface MCRRestRequiredPermissionResolver {
+
+    /**
+     * Determines the required permission for the given {@link ContainerRequestContext}.
+     *
+     * @param requestContext the request context
+     * @return the required permission
+     */
+    String resolvePermission(ContainerRequestContext requestContext);
```
yagee-de, line 34: should add `throws WebApplicationException` as user provided value should be validated

mycore-restapi/src/main/java/org/mycore/restapi/v2/access/MCRRestRequiredPermissionDefaultResolver.java
```diff
@@ -0,0 +1,30 @@
...
+
+package org.mycore.restapi.v2.access;
+
+import jakarta.ws.rs.container.ContainerRequestContext;
+
+public class MCRRestRequiredPermissionDefaultResolver implements MCRRestRequiredPermissionResolver {
+
+    @Override
+    public String resolvePermission(ContainerRequestContext requestContext) {
+        return "";
```
yagee-de, line 27: should always throw a InternalServerErrorException and the class should therefor be named MCRRestFailingPermissionResolver or similar

mycore-restapi/src/main/java/org/mycore/restapi/v2/MCRRestAuthorizationFilter.java
```diff
@@ -157,6 +161,26 @@ public void filter(ContainerRequestContext requestContext) {
...
+        } else if (!requiredPermission.isEmpty()) {
+            return requiredPermission;
+        } else {
+            try {
+                return requiredPermissionResolver.getDeclaredConstructor().newInstance()
+                    .resolvePermission(requestContext);
+            } catch (Exception e) {
+                throw new InternalServerErrorException(e);
+            }
+        }
```
yagee-de, line 181: I think this could be simplified and have those cases reduced. See my other remarks

yagee-de (changes requested):
I currently do not see the ContainerRequestContext to permission String use case yet.

mycore-restapi/src/main/java/org/mycore/restapi/v2/annotation/MCRRestRequiredPermission.java
golsch: One case would be if the permission has to be derived from the properties of the resource itself. 
Assume there are read and write keys that are to be managed via one endpoint. For updating, different permissions are required.  If I want to update an access key, I do not know directly whether it is a read or write key and must first determine the corresponding permission.

yagee-de: I see. Than I would suggest a quite different approach. A resolver should decide if the request is allowed or not. It should get the ResourceInfo and the ContainerRequestContext object and than make a final decision. A default implementation could than implement the current behavior of using the HTTP method and path parameters. This should come with a new Annotation then.

golsch: I had already thought about this.  However, it should always be ensured that the basic check resp. `#checkRestAPIAccess` is executed, which is why I opted for this approach. I had also considered refactoring `#checkRestAPIAccess` and then providing an abstract resolver, but I found the current approach  easier to use. Going further, I have considered a similar approach in combination with https://github.com/MyCoRe-Org/mycore/pull/2257.

golsch:
Permissions should be set statically via the annotation and, if necessary, checked in more detail in a strategy. 
Example for read/write access keys, `MCRRestRequiredPermission("mange-access-keys")`. The final authorization with regard to read or write keys should then be checked in a strategy, see #2257.

PR #2229, opened 2024-08-22
MCR-3167 Fix MCR<->ORCID transformation for OCLC, PMID and PMC

chemicus:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3167).


golsch (dismissed):
Looks good to me, the identifiers are compliant with the [standard](https://www.loc.gov/standards/sourcelist/standard-identifier.html) we use.

kkrebs (changes requested):
we should'nt change IDs here, this needs migration steps. Please add these changes to main instead.

kkrebs (approved):

PR #2230, opened 2024-08-22
MCR-3168 Add ORCID work identifiers Handle and K10Plus-PPN and and External-id-url where missing

chemicus:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3168).


golsch (approved):
Looks good to me, the identifiers are compliant with the [standard](https://www.loc.gov/standards/sourcelist/standard-identifier.html) we use.

kkrebs (approved):

PR #2231, opened 2024-08-23
MCR-3169 Allow to suppress instantiation of disabled components

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3169).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/common/MCRUserInformationResolver.java
+++ b/mycore-base/src/main/java/org/mycore/common/MCRUserInformationResolver.java
@@ -18,14 +18,17 @@
 
 package org.mycore.common;
 
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.mycore.common.config.MCRConfiguration2;
+import org.mycore.common.config.annotation.MCRConfigurationProxy;
+import org.mycore.common.config.annotation.MCRInstanceMap;
+import org.mycore.common.config.annotation.MCRSentinel;
+
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.function.Supplier;
-
-import org.mycore.common.config.MCRConfiguration2;
-import org.mycore.common.config.annotation.MCRConfigurationProxy;
-import org.mycore.common.config.annotation.MCRInstanceMap;
 
 /**
  * A {@link MCRUserInformationResolver} can be used to obtain {@link MCRUserInformation}, without knowledge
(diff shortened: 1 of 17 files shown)


sebhofmann (changes requested):
I think the default case should always be true. The MCRSentinel-Interface has a weird name and its only value is to change default behaviour, which in this case is a bad thing. The property Name should always be the same "Enabled" and the default value should always be "true". It should be hardcoded.
Changing this for a specific Implementation would require more Documentation and the User can never be sure if anything is enabled or disabled without reading the documentation. 
The class should be removed and the behaviour should be hardcoded into all related functions. It should also work for general instantiation. From a user perspective it makes no sense to make only MCRInstanceList or MCRInstanceMap entries disabled and not all Instances.
If a instance is disabled, then:
```
Optional<S> getInstanceOf  -> should return an empty optional
S getInstanceOfOrThrow -> should throw an exception
Optional<S> getSingleInstanceOf -> should return an empty optional
S getSingleInstanceOfOrThrow -> should throw an exception 
getInstantiatablePropertyKeys -> should not be contained in the Stream
getInstances -> should not be contained in the map 
```

toKrause:
> The property Name should always be the same "Enabled" and the default value should always be "true". It should be hardcoded.


> The class should be removed and the behaviour should be hardcoded into all related functions. It should also work for general instantiation.

I considered that and in the end decided to not go that route. Here' my thinking:

- The `Enabled` property of this PR is a new feature and introducing it as a general feature might break backwards compatibility.  There might already be `MCRProperty` annotations using the name `Enabled` (for example `MCRCronJob`)  where the semantics of the property most likely will be deviating from what this PR tries to achieve (as it is the case with `MCRCronJob` where the enabled state is combined with the context in order to calculate weather a cron job should be executed).
- Adding the new feature as a general feature would set e precedent that some property names have a special meaning in the instantiating and configuration process. We already have the "reserved keywords" `Class`, `class` (and the absent property name). Adding `Enabled` to the list of property names that have special meaning and that affect the instantiation and configuration process would invite the next person that inevitable adds another feature to the instantiation and configuration process to add yet another keyword (where the first point would also need to be considered and might even be less clear cut than the word `Enabled`).  
If there was a clear namespacing between property names that affect the instantiation and configuration process and actual configuration properties (for example, a leading underscore to differentiate between the former and the latter, i.e. `_Enabled`) than this wouldn't be a problem. But I wouldn't like to introduce such a concept now, seeing that we already have `Class` ans `class` that don't follow such a rule.
- There might be domains where names other than `Enabled` (such as `Ignored`) might feel more natural and, consequently, another default value might be more appropriate.
- I'm not sure that the feature makes sense  in every domain. For example, in `MCRPIAndPredicate` and `MCRInstanceList`, is it really a useful feature to be able to enable or disable individual clauses (and is `Enabled` really a good name for that)?

So, to summarize: 

1. More thought would have to be put in, in order to figure how this feature would interact with existing use cases. If this were a clean-sheet design, than all of that wouldn't be a problem. But as it is, I prefer an approach where the new feature can be included on a case-by-case basis. 
2. As the saying goes, "one man constant is another mans variable". Some flexibility might be useful.

Having said all of this, I like your idea that there is a general approach that works for nested instances (i.e. ones included with one of the annotations) as well as top-level instances (i.e. ones instantiated through the methods in `MCRConfiguration2`), but I think that the configurability and unobtrusiveness, that the proposed implementation offers, is (sadly) necessary, in order to work well within the given code base.

But that configurability and unobtrusiveness need to be anchored somewhere and the best (least worse) place is the annotations for nested instances.

toKrause:
Regarding the name: something being called a "sentinel" is an non-uncommon terminology to denote something that "indicates that further processing should be interrupted/aborted", see for example  [sentinel value](https://en.wikipedia.org/wiki/Sentinel_value), [sentinel node](https://en.wikipedia.org/wiki/Sentinel_node).

sebhofmann (approved):

PR #2236, opened 2024-09-02
MCR 3171 fix pmd issues in mycore pi

Antoniafriedrich:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3171).

### DIFF
--- a/mycore-pi/src/test/java/org/mycore/pi/MCRPIUtils.java
+++ b/mycore-pi/src/test/java/org/mycore/pi/MCRPIUtils.java
@@ -74,7 +74,8 @@ public class MCRPIUtils {
         try {
             return new URI(url).toURL();
         } catch (MalformedURLException | URISyntaxException e) {
-            LOGGER.error("Malformed URL: {}", url,e);
+            String message = "Malformed URL: " + url;
+            LOGGER.error(message, e);
         }
 
         return null;
--- a/mycore-pi/src/test/java/org/mycore/pi/MCRPIUtils.java
+++ b/mycore-pi/src/test/java/org/mycore/pi/MCRPIUtils.java
@@ -74,8 +74,7 @@ public class MCRPIUtils {
         try {
             return new URI(url).toURL();
         } catch (MalformedURLException | URISyntaxException e) {
-            LOGGER.error(e.getMessage(), e);
-            LOGGER.error("Malformed URL: {}", url);
+            LOGGER.error("Malformed URL: {}", url,e);
         }
 
         return null;
--- a/mycore-pi/src/test/java/org/mycore/pi/MCRPIUtils.java
+++ b/mycore-pi/src/test/java/org/mycore/pi/MCRPIUtils.java
@@ -86,6 +86,6 @@ public class MCRPIUtils {
     }
 
     public static MCRURNJsonBundle getBundle(MCRPIRegistrationInfo urnInfo) {
-        return MCRURNJsonBundle.instance(urnInfo, MCRPIUtils.getUrl(urnInfo));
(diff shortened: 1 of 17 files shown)


mycore-pi/src/test/java/org/mycore/pi/MCRPIUtils.java
```diff
@@ -74,7 +74,7 @@ public static URL getUrl(MCRPIRegistrationInfo info) {
         try {
             return new URI(url).toURL();
         } catch (MalformedURLException | URISyntaxException e) {
-            e.printStackTrace();
+            LOGGER.error(e.getMessage(), e);
             LOGGER.error("Malformed URL: {}", url);
```
toKrause: Logging twice is not necessary (and possibly confusing when reading the log), maybe simplify: sth. like `LOGGER.error("Malformed URL: " + url, e);`

mycore-pi/src/test/java/org/mycore/pi/MCRPIUtils.java
```diff
@@ -74,8 +74,7 @@ public static URL getUrl(MCRPIRegistrationInfo info) {
         try {
             return new URI(url).toURL();
         } catch (MalformedURLException | URISyntaxException e) {
-            LOGGER.error(e.getMessage(), e);
-            LOGGER.error("Malformed URL: {}", url);
+            LOGGER.error("Malformed URL: {}", url,e);
```
toKrause: The documentation of the logging method you used, `Logger#error(String message, Object p0, Object p0)`, states:

```
/**
 * Logs a message with parameters at error level.
 *
 * @param message the message to log; the format depends on the message factory.
 * @param p0 parameter to the message.
 * @param p1 parameter to the message.
 */
```
The third parameter is intended to be used as the second replacement for a placeholder in the message. Your message has only one placeholder.

When running your code, it turns out that the third parameter is indeed logged as an exception, but this relies on [undocumented behavior of a specific message implementation in log4j](https://github.com/apache/logging-log4j2/blob/5a5d3aefdc75045bb66f55a16c40a9a07a463738/log4j-api/src/main/java/org/apache/logging/log4j/message/ReusableParameterizedMessage.java#L140) (which is, if I counted correctly, just one of 24 implementations that are part of log4j core library alone). In other words: This will most likely break for users with a differing logging configuration.

We should stick to the documented interface contract of [`Logger`](https://github.com/apache/logging-log4j2/blob/rel/2.19.0/log4j-api/src/main/java/org/apache/logging/log4j/Logger.java), i.e. call, as previously suggested,  `LOGGER.error("Malformed URL: " + url, e);`

toKrause (changes requested):

mycore-pi/src/test/java/org/mycore/pi/MCRPIUtils.java
Antoniafriedrich: Just so I understand you right, you want me to remove the {}, correct?

toKrause: I recommend to use the [2-parameter-version of the `error`-method](https://github.com/apache/logging-log4j2/blob/5a5d3aefdc75045bb66f55a16c40a9a07a463738/log4j-api/src/main/java/org/apache/logging/log4j/Logger.java#L893) that actually uses the provided exception an exception, i.e. `LOGGER.error("Malformed URL: " + url, e);`.

toKrause: Just removing the `{}` would still use the same 3-parameter-version of the `error`-method. Additionally, the first two parameters need to be concatenated to form a single first parameter for the 2-parameter version.

toKrause (approved):

PR #2238, opened 2024-09-03
MCR 3174 fix pmd issues in mycore sword

Antoniafriedrich:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3174).

### DIFF
--- a/mycore-sword/src/main/java/org/mycore/sword/MCRSwordUtil.java
+++ b/mycore-sword/src/main/java/org/mycore/sword/MCRSwordUtil.java
@@ -230,18 +230,18 @@ public class MCRSwordUtil {
 
         final Path zipTempFile = Files.createTempFile("swordv2_", fileName);
         MessageDigest md5Digest = null;
-
+        InputStream digestedInputStream=inputStream;
         if (checkMd5 != null) {
             try {
                 md5Digest = MessageDigest.getInstance("MD5");
-                inputStream = new DigestInputStream(inputStream, md5Digest);
+                digestedInputStream = new DigestInputStream(inputStream, md5Digest);
             } catch (NoSuchAlgorithmException e) {
                 MCRTransactionHelper.beginTransaction();
                 throw new MCRConfigurationException("No MD5 available!", e);
             }
         }
 
-        Files.copy(inputStream, zipTempFile, StandardCopyOption.REPLACE_EXISTING);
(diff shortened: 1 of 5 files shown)


mycore-sword/src/main/java/org/mycore/sword/MCRSwordUtil.java
```diff
@@ -572,10 +576,6 @@ static void addEditMediaLinks(String collection, DepositReceipt depositReceipt,
         }
     }
 
-    public interface MCRFileValidator {
```
toKrause, line 575: Why has this public interface  been removed while fixing `AvoidReassigningParameters`?

mycore-sword/src/main/java/org/mycore/sword/MCRSwordUtil.java
```diff
@@ -178,7 +176,6 @@ public static MediaResource getZippedDerivateMediaResource(String object) {
     }
 
     private static void addDirectoryToZip(ZipArchiveOutputStream zipOutputStream, Path directory) {
-        MCRSession currentSession = MCRSessionMgr.getCurrentSession();
```
toKrause, line 181: `MCRSessionMgr#getCurrentSession()` has a potential side effect. Not sure if removing the method call is okay here.

mycore-sword/src/main/java/org/mycore/sword/MCRSwordUtil.java
Antoniafriedrich: Just checked it. It seems to be in line 376 now. I am not sure why it is marked as deleted here but it is still there.

mycore-sword/src/main/java/org/mycore/sword/MCRSwordUtil.java
Antoniafriedrich: The variable was targeted by UnusedLocalVariable PMD rule as well as greyed out by IntelliJ but if it has negative side effects I can leave it in and surpress the warning. Would that be better?

toKrause: Or just call `MCRSessionMgr.getCurrentSession();` and ignore the return value.

Not sure if there are side effect, but the call could be intentional. That's why I just commented and didn't request a change. Someone with more knowledge should make that decision.

sebhofmann (approved):

PR #2255, opened 2024-09-11
MCR 3184 fix pmd issues in mycore user 2

Antoniafriedrich:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3184).

### DIFF
--- a/mycore-user2/src/main/java/org/mycore/user2/MCRRealmFactory.java
+++ b/mycore-user2/src/main/java/org/mycore/user2/MCRRealmFactory.java
@@ -158,9 +158,6 @@ public class MCRRealmFactory {
             }
         }
         realmsDocument = root.getDocument();
-        MCRRealmFactory.realmsMap = realmsMap;
-        MCRRealmFactory.realmsList = realmsList;
-        MCRRealmFactory.attributeMapper = attributeMapper;
     }
 
     private static Document getRealms() throws JDOMException, TransformerException, IOException {
--- a/mycore-user2/src/main/java/org/mycore/user2/MCRRealmFactory.java
+++ b/mycore-user2/src/main/java/org/mycore/user2/MCRRealmFactory.java
@@ -157,7 +157,7 @@ public class MCRRealmFactory {
                 localRealm = realm;
             }
         }
-        MCRRealmFactory.realmsDocument = root.getDocument();
+        realmsDocument = root.getDocument();
         MCRRealmFactory.realmsMap = realmsMap;
         MCRRealmFactory.realmsList = realmsList;
         MCRRealmFactory.attributeMapper = attributeMapper;
--- a/mycore-user2/src/main/java/org/mycore/user2/MCRRealmFactory.java
+++ b/mycore-user2/src/main/java/org/mycore/user2/MCRRealmFactory.java
@@ -158,6 +158,9 @@ public class MCRRealmFactory {
             }
         }
         realmsDocument = root.getDocument();
+        MCRRealmFactory.realmsMap = realmsMap;
+        MCRRealmFactory.realmsList = realmsList;
+        MCRRealmFactory.attributeMapper = attributeMapper;
     }
 
     private static Document getRealms() throws JDOMException, TransformerException, IOException {
--- a/mycore-user2/src/main/java/org/mycore/user2/MCRUser.java
+++ b/mycore-user2/src/main/java/org/mycore/user2/MCRUser.java
@@ -546,10 +546,7 @@ public class MCRUser implements MCRUserInformation, Cloneable, Serializable {
     @Override
     public boolean isUserInRole(final String role) {
         boolean directMember = getSystemRoleIDs().contains(role) || getExternalRoleIDs().contains(role);
-        if (directMember) {
-            return true;
-        }
-        return MCRRoleManager.isAssignedToRole(this, role);
+        return directMember||MCRRoleManager.isAssignedToRole(this, role);
     }
 
     @ElementCollection(fetch = FetchType.EAGER)
--- a/mycore-user2/src/main/java/org/mycore/user2/MCRUser.java
+++ b/mycore-user2/src/main/java/org/mycore/user2/MCRUser.java
@@ -730,6 +730,7 @@ public class MCRUser implements MCRUserInformation, Cloneable, Serializable {
 
     @Transient
     @XmlElement(name = "owner")
+    @SuppressWarnings("PMD.UnusedPrivateMethod")
     private UserIdentifier getOwnerId() {
         if (owner == null) {
             return null;
--- a/mycore-user2/src/main/java/org/mycore/user2/MCRUser.java
+++ b/mycore-user2/src/main/java/org/mycore/user2/MCRUser.java
@@ -552,13 +552,6 @@ public class MCRUser implements MCRUserInformation, Cloneable, Serializable {
         return MCRRoleManager.isAssignedToRole(this, role);
     }
 
-    /**
-     * @param attributes the attributes to set
-     */
-    public void setAttributes(SortedSet<MCRUserAttribute> attributes) {
-        this.attributes = attributes;
-    }
-
     @ElementCollection(fetch = FetchType.EAGER)
     @CollectionTable(name = "MCRUserAttr",
         joinColumns = @JoinColumn(name = "id"),
@@ -569,6 +562,13 @@ public class MCRUser implements MCRUserInformation, Cloneable, Serializable {
     @XmlElement(name = "attribute")
     public SortedSet<MCRUserAttribute> getAttributes() {
         return this.attributes;
+    }
+
+    /**
+     * @param attributes the attributes to set
+     */
+    public void setAttributes(SortedSet<MCRUserAttribute> attributes) {
+        this.attributes = attributes;
     }
 
     /**
@@ -725,7 +725,7 @@ public class MCRUser implements MCRUserInformation, Cloneable, Serializable {
         if (anyMatch.isPresent()) {
             MCRUserAttribute attr = anyMatch.get();
             attr.setValue(value);
-            getAttributes().removeIf(a -> a.getName().equals(name) && a != attr);
+            getAttributes().removeIf(a -> a.getName().equals(name) && !(a.equals(attr)));
         } else {
             getAttributes().add(new MCRUserAttribute(name, value));
         }
--- a/mycore-user2/src/main/java/org/mycore/user2/MCRUser.java
+++ b/mycore-user2/src/main/java/org/mycore/user2/MCRUser.java
@@ -546,7 +546,7 @@ public class MCRUser implements MCRUserInformation, Cloneable, Serializable {
     @Override
     public boolean isUserInRole(final String role) {
         boolean directMember = getSystemRoleIDs().contains(role) || getExternalRoleIDs().contains(role);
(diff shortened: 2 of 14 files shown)


mycore-user2/src/main/java/org/mycore/user2/MCRUser.java
```diff
@@ -725,14 +722,15 @@ public void setUserAttribute(String name, String value) {
         if (anyMatch.isPresent()) {
             MCRUserAttribute attr = anyMatch.get();
             attr.setValue(value);
-            getAttributes().removeIf(a -> a.getName().equals(name) && a != attr);
+            getAttributes().removeIf(a -> a.getName().equals(name) && !(a.equals(attr)));
```
sebhofmann: This would remove all attributes which do not have the same name and value and previously it would remove all attributes which are not the same object. I don't know if this produces any negative side effects.

mycore-user2/src/main/java/org/mycore/user2/MCRUserAttributeMapper.java
```diff
@@ -113,88 +113,106 @@ public boolean mapAttributes(final Object object, final Map<String, ?> attribute
...
+
+            final String name = attrAnno.name().isEmpty() ? getAttriutebName(annotated) : attrAnno.name();
+            final List<Attribute> attribs = attributeMapping.get(name);
 
-            if (annotated instanceof Field field) {
-                attrAnno = field.getAnnotation(MCRUserAttribute.class);
-            } else if (annotated instanceof Method method) {
-                attrAnno = method.getAnnotation(MCRUserAttribute.class);
+            if (attributes == null) {
+                return changed;
```
sebhofmann: This needs to be continue or the other annotation are skipped.

mycore-user2/src/main/java/org/mycore/user2/MCRUserAttributeMapper.java
```diff
@@ -113,88 +113,106 @@ public boolean mapAttributes(final Object object, final Map<String, ?> attribute
...
-                    }
+            for (Attribute attribute : attribs) {
+                if (!attributes.containsKey(attribute.mapping)) {
+                    continue;
+                }
+
+                Object value = attributes.get(attribute.mapping);
+                if (value == null) {
+                    LOGGER.warn("Could not apply mapping for {}", attribute.mapping);
+                    continue;
```
sebhofmann: The Code before would not continue, instead it would use the MCRUserAttributeConverter with an null value and abort later.
Since all the MCRUserAttributeConverter in mycore check for a null value, the behavior is probably the same.

mycore-user2/src/main/java/org/mycore/user2/MCRUserServlet.java
```diff
@@ -237,116 +241,163 @@ private void saveCurrentUser(HttpServletRequest req, HttpServletResponse res) th
      */
     private void saveUser(HttpServletRequest req, HttpServletResponse res) throws Exception {
         MCRUser currentUser = MCRUserManager.getCurrentUser();
-        if (!checkUserIsNotNull(res, currentUser, null)) {
+        Document doc = (Document) (req.getAttribute("MCRXEditorSubmission"));
+        Element u = doc.getRootElement();
+        MCRUser user = getUserOrCreateNew(res, u, currentUser);
+
+        if (!checkUserIsNotNull(res, currentUser, null) || !checkPermissions(res) || user == null) {
```
sebhofmann: checkUserIsNotNull was the first check before and now the stuff in getUserOrCreateNew can be called with a null user.

sebhofmann: The new user is created before checking the rights. You should insert the old method again and rework it again.

mycore-user2/src/main/java/org/mycore/user2/login/MCRLDAPClient.java
```diff
@@ -213,30 +228,10 @@ public boolean updateUserProperties(MCRUser user) throws NamingException {
...
-        name = name.replaceAll("\\s+", " ").trim();
-        if (name.contains(",")) {
-            return name;
-        }
-        int pos = name.lastIndexOf(' ');
-        if (pos == -1) {
-            return name;
-        }
-        return name.substring(pos + 1) + ", " + name.substring(0, pos);
+        return (name.replaceAll("\\s+", " ").trim()).contains(",") ? name : name.lastIndexOf(' ') == -1 ? name
```
sebhofmann: This is not better readable.

sebhofmann (changes requested):

mycore-user2/src/main/java/org/mycore/user2/MCRUserAttributeMapper.java
Antoniafriedrich: This is correct. Like in the original (line128-191) the rest of the code should be skipped if `attributes==null`

sebhofmann: The Code is wrapped in a for loop. (line 115 old code, line 114 new code) If attributes == null then you skip all the remaining loops, but the old code continues with the other annotations.

Antoniafriedrich: On a second look you are right! Thank you. I corrected the problem.

sebhofmann:
if (!attributes.containsKey(attribute.mapping)) {
                            return changed;
                        }
 in MCRUserAttributeMapper.java seems to still be wrong.

mycore-user2/src/main/java/org/mycore/user2/MCRRealmFactory.java
```diff
@@ -157,7 +157,7 @@ private static void loadRealms() {
                 localRealm = realm;
             }
         }
-        MCRRealmFactory.realmsDocument = root.getDocument();
+        realmsDocument = root.getDocument();
         MCRRealmFactory.realmsMap = realmsMap;
```
rsteph-de, line 161: also remove MCRRealmFactory. for the following next static properties

mycore-user2/src/main/java/org/mycore/user2/MCRUserAttributeMapper.java
```diff
@@ -111,88 +111,103 @@ public static MCRUserAttributeMapper instance(Element attributeMapping) {
...
-
         for (Object annotated : getAnnotated(object)) {
-            MCRUserAttribute attrAnno = null;
-
-            if (annotated instanceof Field field) {
-                attrAnno = field.getAnnotation(MCRUserAttribute.class);
-            } else if (annotated instanceof Method method) {
-                attrAnno = method.getAnnotation(MCRUserAttribute.class);
-            }
+            MCRUserAttribute attrAnno = getMCRUserAttribute(annotated);
```
rsteph-de: method name: use `get*` only for real getters, suggestion `retrieve*` or `calc*`

mycore-user2/src/main/java/org/mycore/user2/MCRUserServlet.java
```diff
@@ -268,25 +324,20 @@ private void saveUser(HttpServletRequest req, HttpServletResponse res) throws Ex
...
-        Attribute hintAttr = hintPath.evaluateFirst(u);
-        String hint = hintAttr == null ? null : hintAttr.getValue();
-        if ((hint != null) && (hint.trim().length() == 0)) {
-            hint = null;
+            }
         }
-        user.setHint(hint);
-
-        updateBasicUserInfo(u, user);
+        return null;
```
sebhofmann: There needs to be a `return user;` after line 334, otherwise it will return null for a existing user and before line 334 there needs to be a `return null;`.

mycore-user2/src/main/java/org/mycore/user2/MCRUserServlet.java
```diff
@@ -268,25 +324,20 @@ private void saveUser(HttpServletRequest req, HttpServletResponse res) throws Ex
...
+    }
 
+    private boolean updateUserRoles(HttpServletResponse res, boolean hasAdminPermission, Element u, MCRUser user,
```
sebhofmann: This method name is confusing because it does something else and there is already a method with the same name.

mycore-user2/src/main/java/org/mycore/user2/MCRUserServlet.java
```diff
@@ -74,10 +75,96 @@ public class MCRUserServlet extends MCRServlet {
...
+        // Check for required fields is done in the editor form itself, not here
+        if ((userName == null) || (realmID == null)) {
+            return true;
+        }
+
+        // In all other cases, combination of userName and realm must not exist
+        return !MCRUserManager.exists(userName, realmID);
+    }
+
+    private static void updateUserOwner(String userName, String realmID, MCRUser user) {
```
sebhofmann: This method name is confusing because it updates the actual user and has nothing todo with the owner.

rsteph-de (changes requested):

mycore-user2/src/main/java/org/mycore/user2/MCRRealmFactory.java
Antoniafriedrich: This one was not marked by pmd and removing it breaks the build.

mycore-user2/src/main/java/org/mycore/user2/MCRUserAttributeMapper.java
```diff
@@ -111,88 +111,103 @@ public static MCRUserAttributeMapper instance(Element attributeMapping) {
...
+                                    "A not nullable attribute \"" + name + "\" was null.");
                             }
 
-                            if (convCls != null) {
-                                MCRUserAttributeConverter converter = convCls.getDeclaredConstructor().newInstance();
-                                LOGGER.debug("convert value \"{}\" with \"{}\"", value, converter.getClass().getName());
-                                value = converter.convert(value,
-                                    attribute.separator != null ? attribute.separator : attrAnno.separator(),
-                                    attribute.getValueMap());
+                            if (updateFieldOrMethod(object, annotated, attribute, value)) {
```
rsteph-de, line 138: `changed = updateFieldOrMethod(...)`

or maybe you can remove the `changed` variable and use `return updateFieldOrMethod(..)` here

and `return false` at the end.

rsteph-de (changes requested):

mycore-user2/src/main/java/org/mycore/user2/MCRUserAttributeMapper.java
sebhofmann: changed = changed || updateFieldOrMethod(...) would work, but the other two suggestions not, because this is inside two loops.

sebhofmann: But i think this is fine now

sebhofmann (dismissed):
We can merge that, but i would prefer if we reset the MCRUserServlet and make a exception for it.

rsteph-de:
I agree. We are not sure that the modifications in MCRUserServlet do not break current behaviour. 
Therefore we should look into these changes seperately after the release.

Antoniafriedrich:
From the last discussion it is not obvious to me, which changes are wanted.
I assume:
`  changed = changed || updateFieldOrMethod(object, attrAnno, attribute, value);`

Anything else I might have missed?

rsteph-de (approved):
looks good for me now

PR #2264, opened 2024-09-17
MCR-3195 New Command: export solr configset {0} to folder {1}

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3195).

### DIFF
--- a/mycore-solr/src/main/java/org/mycore/solr/commands/MCRSolrCoreAdminCommands.java
+++ b/mycore-solr/src/main/java/org/mycore/solr/commands/MCRSolrCoreAdminCommands.java
@@ -206,6 +206,11 @@ public class MCRSolrCoreAdminCommands {
             return;
         }
 
+        if(!Files.isDirectory(exportFolder)){
+            LOGGER.error("Folder {} is not a directory", exportFolder);
+            return;
+        }
+
         Map<String, MCRSolrConfigSetProvider> configSets = MCRSolrConfigSetHelper.getLocalConfigSets();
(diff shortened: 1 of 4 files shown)


mycore-solr/src/main/java/org/mycore/solr/commands/MCRSolrCoreAdminCommands.java
```diff
@@ -188,4 +194,33 @@ public static void reloadSolrConfiguration(String configType, String coreID) {
...
         MCRSolrConfigReloader.processConfigFiles(configType, coreID);
     }
+
+    @MCRCommand(
+            syntax = "export solr configset {0} to folder {1}",
+            help = "exports a solr configset to a folder",
+            order = 80)
+    public static void exportSolrConfigSet(String configSet, String folder) {
+        Path exportFolder = Paths.get(folder);
+        if(Files.notExists(exportFolder)) {
```
toKrause, line 204: Could also check if `exportFolder` is indeed a folder and actually writable

rsteph-de: isFolder() seems a good idea,
on the other side this command is for admins that usually understand the exception that is thrown if the folder is not writabe.

sebhofmann: Added isDirectory check. Checking if a file is writable is only possible if the file already exists.

toKrause (approved):

PR #2267, opened 2024-09-18
MCR-3198 event handler for merging duplicate categories

erodde:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3198).


mycore-base/src/main/java/org/mycore/common/xml/MCRURIResolver.java
```diff
@@ -1026,7 +1027,7 @@ public Source resolve(String href, String base) {
                 }
             } catch (Exception ex) {
                 LOGGER.info("MCRNotNullResolver caught exception: {}", ex.getLocalizedMessage());
-                LOGGER.debug(ex.getStackTrace());
+                LOGGER.debug(Arrays.toString(ex.getStackTrace()));
```
erodde: I noticed this while debugging UBO, just sneaked it in for better debug logs.

erodde:
In the video conference yesterday the question came up whether to include the new code in the MCRClassificationMappingEventHandler or leave it in its own EventHandler.

mycore-base/src/main/java/org/mycore/common/xml/MCRURIResolver.java
rsteph-de: usually we log the whole exception object
`LOGGER.debug(ex)` 
This would probably duplicate some information, because part of it is was already logged as "INFO".

rsteph-de (dismissed):

rsteph-de (dismissed):
perfect!

mycore-mods/src/test/java/org/mycore/mods/merger/MCRCategoryMergeEventHandlerTest.java
```diff
@@ -0,0 +1,81 @@
...
+    @Test
+    public void testHandleObjectCreatedMultipleGenres() throws IOException, JDOMException, URISyntaxException {
+
+        MCRSessionMgr.getCurrentSession();
+        MCRTransactionHelper.isTransactionActive();
+        ClassLoader classLoader = getClass().getClassLoader();
+        SAXBuilder saxBuilder = new SAXBuilder();
+
+        MCRCategory category = MCRXMLTransformer
+            .getCategory(saxBuilder.build(classLoader.getResourceAsStream(TEST_DIRECTORY + "genre.xml")));
```
toKrause: Should rather use `MCRResourceHelper#getResourceAsStream(String)` instead of `ClassLoader#getResourceAsStream(String)` (two occurances).

toKrause (changes requested):

toKrause (dismissed):
I have created further test cases, not all of which work and for which I am not necessarily sure what the expected result actually is. The description of the issue doesn't provide enough information. You can merge [this PR](https://github.com/MyCoRe-Org/mycore/pull/2295) into this PR to obtain these test cases.

But as it is, I think too many cases are not (correctly) taken into account. Maybe the issue (and then the event handler implementation) needs to be reduced in scope in order to to get rid of these problems.

This probably needs to be discussed again in the next web conference.

erodde:
After our conversation on the 3rd of December I implemented the discussed changes. The Handlers are now ready for review.

mycore-mods/src/main/java/org/mycore/mods/merger/MCRRedundantModsGenreEventHandler.java
```diff
@@ -0,0 +1,82 @@
...
+            Objects.equals(el1.getAttributeValue("authority"), el2.getAttributeValue("authority"));
+        final String displayLabel1 = el1.getAttributeValue("displayLabel");
+        final String displayLabel2 = el2.getAttributeValue("displayLabel");
+
+        final String type1 = el1.getAttributeValue("type");
+        final String type2 = el2.getAttributeValue("type");
+
+        final String authorityName = el1.getAttributeValue("authorityURI") != null
+                                     ? el1.getAttributeValue("authorityURI")
+                                     : el1.getAttributeValue("authority");
```
golsch: Suggestion: refactor to `#getAutorityName(el)` and may move to parent class.

mycore-mods/src/main/java/org/mycore/mods/merger/MCRRedundantModsGenreEventHandler.java
```diff
@@ -0,0 +1,82 @@
...
+
+        final MCRCategoryID cat1 = MCRClassMapper.getCategoryID(el1);
+        final MCRCategoryID cat2 = MCRClassMapper.getCategoryID(el2);
+
+        final String classificationName1 = cat1 != null ? cat1.toString() : "unknown";
+        final String classificationName2 = cat2 != null ? cat2.toString() : "unknown";
```
golsch: Suggestion: refactor to `getClassificationName` and move to parent class.
```
protected String getClassificationName(Element el) {
  return Optional.ofNullable(MCRClassMapper.getCategoryID(el)).map(MCRCategoryID::toString).orElse(UNKNOWN)
}
```

mycore-mods/src/main/java/org/mycore/mods/merger/MCRRedundantModsGenreEventHandler.java
```diff
@@ -0,0 +1,82 @@
...
+
+        if (hasSameAuthority && !Objects.equals(displayLabel1, displayLabel2)) {
+
+            String logMessage = new MessageFormat("""
+            There are inconsistencies found between the classifications {0} and {1}. They have the same authority "{2}",
+            but {0} has the displayLabel "{3}" and {1} has the displayLabel "{4}".""", Locale.ROOT)
+                .format(new Object[] {classificationName1, classificationName2, authorityName,
+                    displayLabel1, displayLabel2});
+
+            LOGGER.warn(logMessage);
```
golsch: log4j supports formatted outputs.

mycore-mods/src/main/java/org/mycore/mods/merger/MCRRedundantModsGenreEventHandler.java
```diff
@@ -0,0 +1,82 @@
...
+        if (hasSameAuthority && !Objects.equals(type1,type2)) {
+
+            String logMessage = new MessageFormat("""
+            There are inconsistencies found between the classifications {0} and {1}. They have the same authority "{2}",
+            but {0} has the type "{3}" and {1} has the type "{4}".""", Locale.ROOT)
+                .format(new Object[] {classificationName1, classificationName2, authorityName, type1, type2});
+
+            LOGGER.warn(logMessage);
+            return false;
+        }
```
golsch: Suggestion: the readability could be improved here.

```
if (hasSameAutority) {
  if(!checkDisplayLabelConsistence(el1, el2) {
    return false;
  }
  ...
}
```

mycore-mods/src/main/java/org/mycore/mods/merger/MCRAbstractRedundantModsEventHandler.java
```diff
@@ -0,0 +1,139 @@
...
+import org.mycore.mods.MCRMODSWrapper;
+import org.mycore.mods.classification.MCRClassMapper;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Checks for and removes redundant classifications in Mods-Documents. If a classification category and
+ * the classification's child category are both present in the document, the parent classification will
+ * be removed. The processed document will be finally be sorted using {@link MCRMODSSorter}.<br><br>
```
golsch: Every handler would end up sorting again, which is not necessary. I would keep the operation more atomic, possibly in a separate event handler.

mycore-mods/src/main/java/org/mycore/mods/merger/MCRRedundantModsClassificationEventHandler.java
```diff
@@ -0,0 +1,66 @@
...
+        final String displayLabel2 = el2.getAttributeValue("displayLabel");
+
+        if (hasSameAuthority && !Objects.equals(displayLabel1, displayLabel2)) {
+            final String authorityName = el1.getAttributeValue("authorityURI") != null
+                                         ? el1.getAttributeValue("authorityURI")
+                                         : el1.getAttributeValue("authority");
+            final MCRCategoryID cat1 = MCRClassMapper.getCategoryID(el1);
+            final MCRCategoryID cat2 = MCRClassMapper.getCategoryID(el2);
+            final String classificationName1 = cat1 != null ? cat1.toString() : "unknown";
+            final String classificationName2 = cat2 != null ? cat2.toString() : "unknown";
```
golsch: see `MCRRedunantModsGenreEventHandler`

mycore-mods/src/main/java/org/mycore/mods/merger/MCRRedundantModsClassificationEventHandler.java
```diff
@@ -0,0 +1,66 @@
...
+
+            String logMessage = new MessageFormat("""
+                There are inconsistencies found between the classifications {0} and {1}.
+                They have the same authority "{2}", but {0} has the displayLabel "{3}" and {1}
+                has the displayLabel "{4}".""", Locale.ROOT).format(new Object[] {classificationName1,
+                classificationName2, authorityName, displayLabel1, displayLabel2});
+
+            LOGGER.warn(logMessage);
```
golsch: see `MCRRedunantModsGenreEventHandler`

mycore-mods/src/main/java/org/mycore/mods/merger/MCRAbstractRedundantModsEventHandler.java
```diff
@@ -0,0 +1,139 @@
...
+     */
+    protected void dropRedundantCategories(List<Element> elements) {
+        for (int i = 0; i < elements.size(); i++) {
+            for (int j = i + 1; j < elements.size(); j++) {
+
+                Element element1 = elements.get(i);
+                Element element2 = elements.get(j);
+                Element parentElement = MCRCategoryMerger.getElementWithParentCategory(element1, element2);
+                if (parentElement != null && isConsistent(element1, element2)) {
+                        parentElement.detach();
```
golsch, line 150: fix indentation

mycore-mods/src/main/java/org/mycore/mods/merger/MCRAbstractRedundantModsEventHandler.java
```diff
@@ -0,0 +1,139 @@
...
+            .toList();
+        dropRedundantCategories(supportedElements);
+
+        List<Element> relatedItems = getAllRelatedItems(mods);
+        for (Element relatedItem : relatedItems) {
+            if (relatedItem.getAttribute("href", MCRConstants.XLINK_NAMESPACE) == null) {
+                dropRedundantCategories(getAllDescendants(relatedItem));
+            }
+        }
+        MCRMODSSorter.sort(mods);
```
golsch: see javadoc

mycore-mods/src/main/java/org/mycore/mods/merger/MCRRedundantModsClassificationEventHandler.java
```diff
@@ -0,0 +1,66 @@
...
+        final boolean hasSameAuthority = Objects.equals(el1.getAttributeValue("authorityURI"),
+            el2.getAttributeValue("authorityURI")) &&
+            Objects.equals(el1.getAttributeValue("authority"), el2.getAttributeValue("authority"));
+        final String displayLabel1 = el1.getAttributeValue("displayLabel");
+        final String displayLabel2 = el2.getAttributeValue("displayLabel");
+
+        if (hasSameAuthority && !Objects.equals(displayLabel1, displayLabel2)) {
+            final String authorityName = el1.getAttributeValue("authorityURI") != null
+                                         ? el1.getAttributeValue("authorityURI")
+                                         : el1.getAttributeValue("authority");
```
golsch: see `MCRRedunantModsGenreEventHandler`

mycore-mods/src/main/java/org/mycore/mods/merger/MCRRedundantModsClassificationEventHandler.java
```diff
@@ -0,0 +1,66 @@
...
+     * the displayLabels differ from each other.
+     * @param el1 the first element to be compared
+     * @param el2 the first element to be compared
+     * @return false if inconsistent
+     */
+    @Override
+    protected boolean isConsistent(Element el1, Element el2) {
+        final boolean hasSameAuthority = Objects.equals(el1.getAttributeValue("authorityURI"),
+            el2.getAttributeValue("authorityURI")) &&
+            Objects.equals(el1.getAttributeValue("authority"), el2.getAttributeValue("authority"));
```
golsch: see `MCRRedunantModsGenreEventHandler`

golsch (changes requested):

mycore-mods/src/main/java/org/mycore/mods/merger/MCRAbstractRedundantModsEventHandler.java
erodde: Do you mean creating yet another EventHandler that does nothing but sorting the document?
Edit: I removed the sorting from the EventHandler. Implement sorting-EventHandler in another ticket?

golsch: In principle, merging, sorting and so on are post processing for me. I'm not sure whether it generally makes sense to always write a new event handler for this or whether it would make more sense to write an event handler for which you can configure all post process actions.

erodde: I'll make a note to discuss this at the next developer meeting. But I don't see this in the scope of this ticket, so I'll resolve the conversation.

toKrause (changes requested):
The problem with deeply nested related items should be fixed.

The remaining notes are only suggestions for a potentially improved design and small stuff.

mycore-mods/src/main/java/org/mycore/mods/merger/MCRAbstractRedundantModsEventHandler.java
```diff
@@ -0,0 +1,171 @@
...
+import java.util.Objects;
+import java.util.Optional;
+
+/**
+ * Checks for and removes redundant classifications in Mods-Documents. If a classification category and
+ * the classification's child category are both present in the document, the parent classification will
+ * be removed. <br><br>
+ * This abstract class can be extended by overriding the following abstract methods:
+ * <ol>
+ *     <li>{@link MCRAbstractRedundantModsEventHandler#isConsistent(Element, Element)}: custom checks for
```
toKrause, line 43: The concept of "consistency" as used here confused me for a bit.

From the point of view, where only the currently supported MODS elements (`mods:genre`, `mods:classification`) are considered, this is reasonable, because, in context of a MyCoRe application, we always expect those elements to have identical types and display labels; i.e. we expect consistency across all occurrences. Framing the concept of "checking, whether two elements are in a way related to each other and should therefore be considered for redundancy merging" as  consistency check is therefore okay for now.

However, I suggest changing the naming here to allow for future use cases where elements may actually be in different "worlds"; for example `mods:classification` elements with differing `generator` attributes.

Suggestions: `areRelated`, `areInSameScope`

mycore-mods/src/main/java/org/mycore/mods/merger/MCRAbstractRedundantModsEventHandler.java
```diff
@@ -0,0 +1,171 @@
...
+
+        Element mods = mcrmodsWrapper.getMODS();
+        List<Element> supportedElements = getAllDescendants(mods).stream()
+            .filter(element -> element.getName().equals(getClassificationElementName()))
+            .filter(element -> MCRClassMapper.getCategoryID(element) != null).toList();
+        dropRedundantCategories(supportedElements);
+
+        List<Element> relatedItems = getAllRelatedItems(mods);
+        for (Element relatedItem : relatedItems) {
+            if (relatedItem.getAttribute("href", MCRConstants.XLINK_NAMESPACE) == null) {
```
toKrause, line 98: Related items may be nested more than one deep (even though it will be very unlikely to ever encounter such a MODS document in the real world). The current implementation doesn't work for example for:

```xml
<mods:mods xmlns:mods="http://www.loc.gov/mods/v3" xmlns:xlink="http://www.w3.org/1999/xlink">
  <mods:relatedItem type="intern">
    <mods:relatedItem type="intern">
      <mods:genre type="intern" authorityURI="http://www.mycore.org/classifications/mir_genres" valueURI="http://www.mycore.org/classifications/mir_genres#x"/>
      <mods:genre type="intern" authorityURI="http://www.mycore.org/classifications/mir_genres" valueURI="http://www.mycore.org/classifications/mir_genres#x-1"/>
      <mods:genre type="intern" authorityURI="http://www.mycore.org/classifications/mir_genres" valueURI="http://www.mycore.org/classifications/mir_genres#x-1-1"/>
      <mods:genre type="intern" authorityURI="http://www.mycore.org/classifications/mir_genres" valueURI="http://www.mycore.org/classifications/mir_genres#y"/>
    </mods:relatedItem>
  </mods:relatedItem>
</mods:mods>
```
Also, the additional filters (correct element name, element represents a classification category) should also applied for the descendants of the related item (which will mostly be a performance hit, not a functional problem, but the current implementation would for example remove a `mods:name` that has the same authority and a more general category then a `mods:genre`, just to give a ridiculous example).

The filtering can also be done while compiling the list of descendants. No need to filter the list after compiling all non-related-item-descendents.

Suggestion:
``` java
protected void mergeCategories(MCRObject obj) {
    MCRMODSWrapper mcrmodsWrapper = new MCRMODSWrapper(obj);
    if (mcrmodsWrapper.getMODS() == null) {
        return;
    }
    logger.info("merge redundant " + getClassificationElementName() + " categories for {}", obj.getId());
    Element mods = mcrmodsWrapper.getMODS();
    mergeCategories(mods);
}

private void mergeCategories(Element modsOrRelatedItem) {
    List<Element> supportedElements = getAllSupportedDescendants(modsOrRelatedItem);
    dropRedundantCategories(supportedElements);

    List<Element> relatedItems = getAllRelatedItems(modsOrRelatedItem);
    for (Element relatedItem : relatedItems) {
        if (relatedItem.getAttribute("href", MCRConstants.XLINK_NAMESPACE) == null) {
            mergeCategories(relatedItem);
        }
    }
}

protected List<Element> getAllSupportedDescendants(Element modsOrRelatedItem) {
    List<Element> descendants = new ArrayList<>();
    for (Element child : modsOrRelatedItem.getChildren()) {
        if (child.getName().equals("relatedItem")) {
            continue;
        }
        if (child.getName().equals(getClassificationElementName()) && MCRClassMapper.getCategoryID(child) != null) {
            descendants.add(child);
        }
        descendants.addAll(getAllSupportedDescendants(child));
    }
    return descendants;
}
```

mycore-mods/src/main/java/org/mycore/mods/merger/MCRAbstractRedundantModsEventHandler.java
```diff
@@ -0,0 +1,171 @@
...
+     * @param elements a list of classification elements that are all compared to each other in pairs
+     */
+    protected void dropRedundantCategories(List<Element> elements) {
+        for (int i = 0; i < elements.size(); i++) {
+            for (int j = i + 1; j < elements.size(); j++) {
+
+                Element element1 = elements.get(i);
+                Element element2 = elements.get(j);
+                Element parentElement = MCRCategoryMerger.getElementWithParentCategory(element1, element2);
+                if (parentElement != null && isConsistent(element1, element2)) {
```
toKrause: If a parent element can be determined by `MCRCategoryMerger.getElementWithParentCategory` it can be asserted, that the "authority" of the two elements representing a classification category must be the same, otherwise, the two categories would not belong to the same classification.

This knowledge could be used to simplify the implementations of `isConsistent` in the two concrete classes inheriting from `MCRAbstractRedundantModsEventHandler` and make it easier to reason about these.

Suggestion:

```java
Element parentElement = MCRCategoryMerger.getElementWithParentCategory(element1, element2);
if (parentElement != null) {
    assert Objects.equals(getAuthority(element1), getAuthority(element2));
    if (isConsistent(element1, element2)) {
        parentElement.detach();
   }
}

File: mycore-mods/src/main/java/org/mycore/mods/merger/MCRRedundantModsClassificationEventHandler.java [hunk 17]
```diff
@@ -0,0 +1,65 @@
...
+    /**
+     * Returns false if the authorities of the two classification elements are the same, but
+     * the displayLabels differ from each other.
+     * @param el1 the first element to be compared
+     * @param el2 the first element to be compared
+     * @return false if inconsistent
+     */
+    @Override
+    protected boolean isConsistent(Element el1, Element el2) {
+        return !hasSameAuthority(el1, el2) || checkDisplayLabelConsistence(el1, el2);
```
toKrause: Could just be `checkDisplayLabelConsistence(el1, el2);`.

mycore-mods/src/main/java/org/mycore/mods/merger/MCRRedundantModsClassificationEventHandler.java
```diff
@@ -0,0 +1,65 @@
...
+     * Checks if both elements have the same displayLabel. Logs a warning if not.
+     * @param el1 first element to check
+     * @param el2 second element to check
+     * @return true, if both elements have the same displayLabel (or both have none)
+     */
+    private boolean checkDisplayLabelConsistence(Element el1, Element el2) {
+        final String displayLabel1 = el1.getAttributeValue("displayLabel");
+        final String displayLabel2 = el2.getAttributeValue("displayLabel");
+
+        final String classificationName1 = getClassificationName(el1);
```
toKrause: Values `classificationName1` and `classificationName2` only need to be retrieved inside of the if statement (in this method and similarly in two others)

mycore-mods/src/main/java/org/mycore/mods/merger/MCRRedundantModsGenreEventHandler.java
```diff
@@ -0,0 +1,91 @@
...
+     * Returns false if the authorities of the two genre elements are the same, but
+     * the displayLabels or the types differ from each other.
+     * @param el1 the first element to be compared
+     * @param el2 the first element to be compared
+     * @return false if inconsistent
+     */
+    @Override
+    protected boolean isConsistent(Element el1, Element el2) {
+        return !hasSameAuthority(el1, el2) ||
+            (checkDisplayLabelConsistence(el1, el2) && checkTypeConsistence(el1, el2));
```
toKrause: Could just be `checkDisplayLabelConsistence(el1, el2) && checkTypeConsistence(el1, el2)`.

toKrause (changes requested):
The problem with deeply nested related items should be fixed.

The remaining notes are only suggestions for a potentially improved design and small stuff.

mycore-mods/src/main/java/org/mycore/mods/merger/MCRRedundantModsGenreEventHandler.java
```diff
@@ -0,0 +1,91 @@
...
+     * Checks if both elements have the same type. Logs a warning if not.
+     * @param el1 first element to check
+     * @param el2 second element to check
+     * @return true, if both elements have the same type (or both have none)
+     */
+    private boolean checkTypeConsistence(Element el1, Element el2) {
+        final String type1 = el1.getAttributeValue("type");
+        final String type2 = el2.getAttributeValue("type");
+
+        final String classificationName1 = getClassificationName(el1);
```
toKrause: The values `classificationName1` and `classificationName2` are only needed inside of the if statement (in this and two other, similarly constructed methods).

toKrause (changes requested):

mycore-mods/src/main/java/org/mycore/mods/merger/MCRAbstractRedundantModsEventHandler.java
erodde: I actually feel like "isConsistent" is the vaguer/more abstract term and would like to stick with that for now.

mycore-mods/src/main/java/org/mycore/mods/merger/MCRRedundantModsClassificationEventHandler.java
erodde: why do you think that the authority check is unnecessary?

mycore-mods/src/main/java/org/mycore/mods/merger/MCRRedundantModsClassificationEventHandler.java
erodde: good catch :)

mycore-mods/src/main/java/org/mycore/mods/merger/MCRRedundantModsClassificationEventHandler.java
toKrause: See https://github.com/MyCoRe-Org/mycore/pull/2267#discussion_r1952663017

mycore-mods/src/main/java/org/mycore/mods/merger/MCRAbstractRedundantModsEventHandler.java
erodde: Okay I see, I'll do that.

toKrause (dismissed):

mycore-mods/src/main/java/org/mycore/mods/merger/MCRAbstractRedundantModsEventHandler.java
```diff
@@ -0,0 +1,172 @@
+package org.mycore.mods.merger;
```
golsch, line 19: add license info

mycore-mods/src/main/java/org/mycore/mods/merger/MCRAbstractRedundantModsEventHandler.java
```diff
@@ -0,0 +1,172 @@
...
+ * <ol>
+ *     <li>{@link MCRAbstractRedundantModsEventHandler#isConsistent(Element, Element)}: custom checks for
+ *     the two classifications' consistency.</li>
+ *     <li>{@link MCRAbstractRedundantModsEventHandler#getClassificationElementName()}: the name of
+ *     the mods-element that the EventHandler is checking duplicates for.</li>
+ * </ol>
+ */
+public abstract class MCRAbstractRedundantModsEventHandler extends MCREventHandlerBase {
+
+    private final Logger logger = LogManager.getLogger(getClass());
```
golsch, line 51: remove `getClass()`

mycore-mods/src/main/java/org/mycore/mods/merger/MCRRedundantModsClassificationEventHandler.java
```diff
@@ -0,0 +1,64 @@
+package org.mycore.mods.merger;
```
golsch, line 19: add license info

mycore-mods/src/main/java/org/mycore/mods/merger/MCRRedundantModsGenreEventHandler.java
```diff
@@ -0,0 +1,89 @@
+package org.mycore.mods.merger;
```
golsch, line 19: add license info

mycore-mods/src/test/java/org/mycore/mods/merger/MCRRedundantModsClassificationEventHandlerTest.java
```diff
@@ -0,0 +1,147 @@
+package org.mycore.mods.merger;
```
golsch, line 19: add license info

golsch (changes requested):

mycore-mods/src/test/java/org/mycore/mods/merger/MCRRedundantModsGenreEventHandlerTest.java
```diff
@@ -0,0 +1,184 @@
...
+        modsWrapper.setMODS(modsDocument.getRootElement().detach());
+        modsWrapper.setID("junit", 1);
+
+        MCRRedundantModsGenreEventHandler mergeEventHandler = new MCRRedundantModsGenreEventHandler();
+        mergeEventHandler.handleObjectCreated(null, object);
+
+        LOGGER.info(new MCRJDOMContent(modsWrapper.getMODS()).asString());
+
+        return modsWrapper.getMODS();
+
```
golsch, line 55: remove line

golsch (changes requested):

mycore-mods/src/main/java/org/mycore/mods/merger/MCRAbstractRedundantModsEventHandler.java
toKrause: This is an abstract class, afaik the `getClass()` is necessary to initialize the logger with the actual class (that extends from this abstract class). This is also the reason, why this logger is not static. Without this, interpreting logging messages can be relay challenging, because from the log messages, it is not discernable which handler is actually executing.

golsch: my fault..

erodde:
Sorry for comment deletion, that was an IntelliJ add-in I didn't know yet :)

golsch (approved):

kkrebs (approved):

toKrause (approved):

PR #2270, opened 2024-09-20
MCR-3199 MODS SOLR name bugfix

bt1cu:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3199). fix with for-each loop

### DIFF
--- a/mycore-mods/src/main/resources/xslt/solr/indexing/mods-solr.xsl
+++ b/mycore-mods/src/main/resources/xslt/solr/indexing/mods-solr.xsl
@@ -278,7 +278,12 @@
   <xsl:template name="printModsName">
     <xsl:choose>
       <xsl:when test="mods:displayForm">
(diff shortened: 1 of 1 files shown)


mycore-mods/src/main/resources/xslt/solr/indexing/mods-solr.xsl
```diff
@@ -278,7 +278,12 @@
   <xsl:template name="printModsName">
     <xsl:choose>
       <xsl:when test="mods:displayForm">
-        <xsl:value-of select="fn:normalize-unicode(mods:displayForm)" />
+        <xsl:for-each select="mods:displayForm" >
```
rsteph-de: This can be solved shorter with a String-function:
`<xsl:value-of select="fn:string-join(fn:normalize-unicode(mods:displayForm), ' ')" />`

toKrause (approved):

PR #2274, opened 2024-09-24
MCR-3204 MCRObjectService.setFromXML() is broken

rsteph-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3204).


kkrebs (approved):

yagee-de (changes requested):
Looks good to me, but ticket should be MCR-2997

PR #2284, opened 2024-10-11
MCR-3209 improved orcid2 endpoints

golsch:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3209).


chemicus (approved):
Some public methods have been deleted, which might require code migration in other applications, which need to be mentioned in migration instructions!

PR #2291, opened 2024-10-22
MCR-3210 improve transaction system

Mewel:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3210).

### MCRTransactionManager
**Centralized Transaction Management:** The MCRTransactionManager ensures that all transaction-related operations—beginning, committing, and rolling back—are managed centrally. This eliminates scattered transaction management logic and consolidates error handling, which improves the overall reliability of the system.

**Error Handling and Rollback Mechanism:** If any transaction fails during the commit or rollback process, MCRTransactionManager attempts to roll back any previously started transactions and suppresses any rollback exceptions. This helps prevent partial commits and leaves the system in a consistent state.

**Transaction Prioritization:** The manager allows setting a commit priority for each transaction, ensuring that more critical transactions are committed before others, thus enabling finer control over the transaction lifecycle.

### DIFF
--- a/mycore-base/src/main/java/org/mycore/backend/jpa/MCREntityTransaction.java
+++ b/mycore-base/src/main/java/org/mycore/backend/jpa/MCREntityTransaction.java
@@ -58,4 +58,10 @@ public class MCREntityTransaction implements MCRPersistenceTransaction {
     public boolean isActive() {
         return MCREntityManagerProvider.getCurrentEntityManager().getTransaction().isActive();
     }
+
+    @Override
+    public int getPriority() {
+        return 4000;
+    }
+
 }
--- a/mycore-base/src/main/java/org/mycore/backend/jpa/MCREntityTransaction.java
+++ b/mycore-base/src/main/java/org/mycore/backend/jpa/MCREntityTransaction.java
@@ -45,21 +45,6 @@ public class MCREntityTransaction implements MCRPersistenceTransaction {
     }
 
     @Override
-    public boolean getRollbackOnly() {
-        return MCREntityManagerProvider.getCurrentEntityManager().getTransaction().getRollbackOnly();
-    }
-
-    @Override
-    public void setRollbackOnly() throws IllegalStateException {
-        MCREntityManagerProvider.getCurrentEntityManager().getTransaction().setRollbackOnly();
-    }
-
-    @Override
-    public boolean isActive() {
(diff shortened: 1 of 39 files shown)


mycore-base/src/main/java/org/mycore/common/MCRTransactionManager.java
```diff
@@ -0,0 +1,673 @@
...
+     * @return the currently active {@link TransactionLoader}.
+     */
+    public static TransactionLoader getTransactionLoader() {
+        return transactionLoader;
+    }
+
+    /**
+     * Begins all available {@link MCRPersistenceTransaction} instances that are active.
+     *
+     * <p>It throws an exception if there are already active transactions in the current thread.</p>
```
sebhofmann: This is already stated in the `@throws ` part.

sebhofmann (changes requested):

mycore-base/src/main/java/org/mycore/common/MCRSession.java
```diff
@@ -407,7 +407,7 @@ public long getCreateTime() {
      */
     @Deprecated
```
rsteph-de, line 408: Deprecated in 2021.06 - time to delete?

yagee-de: Not in this PR. Let's do this together and remove deprecated methods all together in main AFTER we forked release branch

mycore-base/src/main/java/org/mycore/backend/jpa/MCREntityTransaction.java
```diff
@@ -45,17 +45,8 @@ public void rollback() {
     }
 
     @Override
-    public boolean getRollbackOnly() {
-        return MCREntityManagerProvider.getCurrentEntityManager().getTransaction().getRollbackOnly();
```
yagee-de, line 49: A library may support transaction that we want to delegate. If an error occurs the transaction of the library may signal via getRollbackOnly() that a commit will fail anyway. So why do we drop getRollbackOnly()?

yagee-de: It's likely that this is only needed, if JPA transaction handling is done via other libraries/container. So we could test the code as suggested.

yagee-de (approved):

sebhofmann (approved):

PR #2295, opened 2024-10-24
MCR-3198 add additional test cases

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3198).


erodde:
Use cases are not only genres, but for example also destatis classifications, which are not on the root level. Further requirements must be explicitly specified.

PR #2299, opened 2024-10-25
MCR-3214 Add handlings for non validatable pdf documents

Antoniafriedrich:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3214).


mycore-validation/src/main/java/org/mycore/validation/pdfa/MCRPDFAFunctions.java
```diff
@@ -82,7 +81,8 @@ public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) throws IO
                     try {
                         results.put(dir.relativize(file).toString(), PDF_A_VALIDATOR.validate(file));
                     } catch (MCRPDFAValidationException e) {
-                        throw new MCRException(e);
+                        results.put(dir.relativize(file).toString(), new ValidationErrorResult());
+
```
yagee-de: Details (cause, stacktrace) of Exception are lost here. Better way is to put the Exception into the map and serialize at least the cause into the Document. You could use a record to old ValidationResult or Exception and use it as a map value.

yagee-de (changes requested):
ValidationErrorResult seems not to fit the purpose

mycore-validation/src/main/java/org/mycore/validation/pdfa/MCRPDFAFunctions.java
```diff
@@ -125,11 +130,77 @@ private static Document createXML(String objectId, Map<String, ValidationResult>
...
+            exceptionElement.appendChild(stackTraceElement);
+
+            exceptionElementList.add(exceptionElement);
+
+            exceptionsQueue.addAll(List.of(currentException.getSuppressed()));
+            if (currentException.getCause() != null &&
+                !currentException.getCause().equals(currentException)) {
+                exceptionsQueue.add(currentException.getCause());
+            }
+        }
```
yagee-de: This should be a hierarchy of elements not a list. We need to know what are causes and what a suppressed exception of which exception

yagee-de (changes requested):

mycore-validation/src/main/java/org/mycore/validation/pdfa/MCRPDFAFunctions.java
```diff
@@ -125,11 +130,99 @@ private static Document createXML(String objectId, Map<String, ValidationResult>
...
+        } else {
+            Element fileElement = document.createElement("file");
+            fileElement.setAttribute("name", fileName);
+            derivateElement.appendChild(fileElement);
+            fileElement.setAttribute("flavour", "Validation Error");
+            List<Element> exceptions = createExceptionElements(document, resultEntry.exception());
+            Element exceptionElements = document.createElement("exceptions");
+            for (Element exception : exceptions) {
+                exceptionElements.appendChild(exception);
+            }
```
yagee-de: You hand over an Exception and receive a List of Exception Elements? I think only one exception could be thrown finally and caught. So no need for "<exceptions>" or a List return type.

mycore-validation/src/main/java/org/mycore/validation/pdfa/MCRPDFAFunctions.java
```diff
@@ -125,11 +130,99 @@ private static Document createXML(String objectId, Map<String, ValidationResult>
...
+
+    /**
+     * Recursive helper method to build the exception tree.
+     */
+    private static List<Element> buildExceptionTree(Document document, Throwable currentException,
+        Queue<Throwable> exceptionsQueue) {
+        List<Element> exceptionElementList = new ArrayList<>();
+        if (currentException == null || exceptionsQueue.contains(currentException)) {
+            return exceptionElementList;
+        }
```
yagee-de: When you have a tree, how can an exception be in there twice?

yagee-de (changes requested):

yagee-de (approved):

PR #2315, opened 2024-11-03
MCR-3219 remove check if directory exists for property MCR.basedir

rsteph-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3219).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/common/config/MCRConfigurationInputStream.java
+++ b/mycore-base/src/main/java/org/mycore/common/config/MCRConfigurationInputStream.java
@@ -99,9 +99,7 @@ public class MCRConfigurationInputStream extends InputStream {
             LogManager.getLogger().info("Current configuration directory: {}",
                 configurationDirectory.getAbsolutePath());
             // set MCR.basedir, is normally overwritten later
-            if (configurationDirectory.isDirectory()) {
-                initStream = getBaseDirInputStream(configurationDirectory);
-            }
+            initStream = getBaseDirInputStream(configurationDirectory);
         }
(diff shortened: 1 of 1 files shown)


mycore-base/src/main/java/org/mycore/common/config/MCRConfigurationInputStream.java
```diff
@@ -99,9 +99,7 @@ public static MCRConfigurationInputStream getMyCoRePropertiesInstance() throws I
             LogManager.getLogger().info("Current configuration directory: {}",
                 configurationDirectory.getAbsolutePath());
             // set MCR.basedir, is normally overwritten later
-            if (configurationDirectory.isDirectory()) {
-                initStream = getBaseDirInputStream(configurationDirectory);
-            }
+            initStream = getBaseDirInputStream(configurationDirectory);
```
toKrause, line 106: Maybe log a warning about the missing base directory? Could help in cases  other then the first startup where there actually is a problem because of a missing or inaccessible directory.

yagee-de: I think the log statement above is sufficient. If an exception occurs later, that should contains all the details, e.g. a specific subdirectory that is not accessible.

yagee-de (approved):

mycore-base/src/main/java/org/mycore/common/config/MCRConfigurationInputStream.java
toKrause: Except that is is just an info and doesn't point to a problem.

toKrause (approved):

PR #2330, opened 2024-11-20
MCR-3303 Digest header is non-standard and deprecated

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3303).

### DIFF
--- a/mycore-restapi/src/main/java/org/mycore/restapi/v2/MCRRestDerivateContents.java
+++ b/mycore-restapi/src/main/java/org/mycore/restapi/v2/MCRRestDerivateContents.java
@@ -252,7 +252,7 @@ public class MCRRestDerivateContents {
      * @see <a href="https://tools.ietf.org/html/rfc3230">RFC 3230</a>
      * @see <a href="https://tools.ietf.org/html/rfc5843">RFC 45843</a>
      */
-    private static String getDigestHeader(MCRDigest digest) {
+    private static String getReprDigestHeaderValue(MCRDigest digest) {
         final String algorithm = digest.getAlgorithm().toLowerCase();
         final String encodedValue = Base64.getEncoder().encodeToString(digest.toBytes());
         return algorithm + "=:" + encodedValue + ":";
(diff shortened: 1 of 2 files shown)


mycore-restapi/src/main/java/org/mycore/restapi/v2/MCRRestDerivateContents.java
```diff
@@ -313,7 +313,7 @@ public Response getFileOrDirectoryMetadata() {
             .lastModified(Date.from(fileAttributes.lastModifiedTime().toInstant()))
             .header(HttpHeaders.CONTENT_LENGTH, fileAttributes.size())
             .tag(getETag(fileAttributes))
-            .header("Digest", getDigestHeader(fileAttributes.digest().toHexString()))
+            .header("Content-Digest", getDigestHeader(fileAttributes.digest().toHexString()))
```
yagee-de: we could use fileAttributes.digest() here directly as a parameter to get the header value. This will support directly SHA-512 in the future, when we move to OCFL.

the format of the value will change to:
Before: MD5={checksum}
After: md5=:{checksum}:
So small letters for the algorithm and the actual value is wrapped in colons (:)
See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Repr-Digest

yagee-de (changes requested):

mycore-restapi/src/main/java/org/mycore/restapi/v2/MCRRestDerivateContents.java
toKrause: Sorry, confused myself her for a second about which header to use and the formatting. Should be correct now.

Not sure what you mean by "use `fileAttributes.digest()` here directly", since that method returns a `MCRDigest` object which is certainly not directly usable as a header value.

toKrause: I think you mean that we should account for the fact, that `MCRFileAttributes#digest()`will not necessarily return a MD5 digest, correct?

yagee-de (approved):
Looks good to me

PR #2332, opened 2024-11-22
MCR-3304 neo4j raises exception if the server url is not defined

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3304).

### DIFF
--- a/mycore-neo4j/src/main/java/org/mycore/mcr/neo4j/index/MCRNeo4JIndexEventHandler.java
+++ b/mycore-neo4j/src/main/java/org/mycore/mcr/neo4j/index/MCRNeo4JIndexEventHandler.java
@@ -62,7 +62,7 @@ public class MCRNeo4JIndexEventHandler extends MCREventHandlerBase {
 
     static {
         if (!ENABLED) {
-            LOGGER.info("Neo4J Indexing is disabled");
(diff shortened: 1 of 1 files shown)


mycore-neo4j/src/main/java/org/mycore/mcr/neo4j/index/MCRNeo4JIndexEventHandler.java
```diff
@@ -53,46 +53,50 @@ public class MCRNeo4JIndexEventHandler extends MCREventHandlerBase {
...
+    private static final boolean ENABLED = MCRConfiguration2
+        .getString(MCRNeo4JConstants.DEFAULT_NEO4J_SERVER_URL).isPresent();
+
     private static final DelayQueue<MCRDelayedRunnable> NEO4J_TASK_QUEUE = new DelayQueue<>();
 
     private static final ScheduledExecutorService NEO4J_TASK_EXECUTOR = Executors.newSingleThreadScheduledExecutor();
 
     static {
+        if (!ENABLED) {
+            LOGGER.info("Neo4J Indexing is disabled");
```
michael-becker: maybe add: "because the property MCRNeo4JConstants.DEFAULT_NEO4J_SERVER_URL ist not set" to specify the reason for disabling neo4j

michael-becker (changes requested):

golsch:
Why not simply make the event handler optional? This way, you are also open to adaptations if necessary.

kkrebs (approved):

PR #2349, opened 2024-12-20
MCR-3227 enable PMD rule GuardLogStatement

yagee-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3227).


mycore-base/src/main/java/org/mycore/access/MCRAccessBaseImpl.java
```diff
@@ -35,13 +35,13 @@
...
- * 
+ *
  * It will simply allow everything and will do nothing on persistent operations.
  * Feel free to extend this class if your implementation can only support parts
  * of the Interface definition.
- * 
+ *
  * @author Jens Kupferschmidt
- * 
+ *
```
Mewel: You have a blank line at the end of each javadoc comment. Looks strange.

Also in between the javadoc you added some. Intellij doesn't like it: "Blank line will be ignored -> Insert &lt;p&gt;."

mycore-base/src/main/java/org/mycore/common/MCRMailer.java
```diff
@@ -302,8 +302,8 @@ public static void send(Element email, Boolean allowException) throws Exception
      * Sends email. When sending email fails (for example, outgoing mail server
      * is not responding), sending will be retried after five minutes. This is
      * done up to 10 times.
-     * 
-     * 
+     *
```
Mewel, line 305: One line break should be enough

Mewel (changes requested):

mycore-base/src/main/java/org/mycore/access/MCRAccessBaseImpl.java
yagee-de: Comes from IntelliJ. It automatically removes blanks at end of line on save. Do not know how to deactivate it, but those "space" are not needed. That's why I didn't put much effort in it.

Mewel (approved):

PR #2351, opened 2025-01-03
MCR-3202 vite processing ui

Mewel:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3202).


sebhofmann (approved):
Works for me. But when we upgrade to bootstrap 5, then we should remove the custom css and replace it with bootstrap classes.

PR #2353, opened 2025-01-10
MCR-3229 enable PMD rule LooseCoupling

yagee-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3229).

This pull request focuses on refactoring the codebase to use more generic collection interfaces (`List` and `Map`) instead of specific implementations (`ArrayList` and `HashMap`). This change improves code flexibility and readability.

Also exceptions to the PMD rule where configured, were specific subclasses provide additional functionality beside their collection functionality.

:warning: **Changes API** in _2024.06.x_ due to nature of the PMD rule, e.g. if a method return type was `ArrayList` it will be `List` after the merge, so minor changes to the code will be necessary.


mycore-base/src/main/java/org/mycore/common/xml/MCRURIResolver.java
```diff
@@ -196,14 +195,15 @@ public static synchronized void init(ServletContext ctx) {
...
+    public static Map<String, String> getParameterMap(String key) {
         String[] param;
         StringTokenizer tok = new StringTokenizer(key, "&");
-        Hashtable<String, String> params = new Hashtable<>();
+        Map<String, String> params = new HashMap<>();
 
         while (tok.hasMoreTokens()) {
             param = tok.nextToken().split("=");
-            params.put(param[0], param.length >= 2 ? param[1] : "");
+            params.put(URLDecoder.decode(param[0], StandardCharsets.UTF_8),
```
Mewel, line 205: Should this be in this commit? Looks like a behavior change.

mycore-base/src/test/java/org/mycore/common/events/MCRShutdownHandlerTest.java
```diff
@@ -143,7 +143,7 @@ public String toString() {
      * It only used by {@link #mark} and {@link #resetCloseables()}.
      */
     private record MCRShutdownHandlerState(boolean shuttingDown,
-        ConcurrentSkipListSet<MCRShutdownHandler.Closeable> requests) {
+        java.util.NavigableSet<MCRShutdownHandler.Closeable> requests) {
```
Mewel: why use package?

mycore-iview2/src/main/java/org/mycore/iview2/services/MCRTilingQueue.java
```diff
@@ -43,6 +43,7 @@
 import jakarta.persistence.TypedQuery;
 
 public class MCRTilingQueue extends AbstractQueue<MCRTileJob> implements Closeable {
+    @SuppressWarnings("PMD.LooseCoupling")
```
Mewel, line 46: looks like a lazy implemented singleton pattern
maybe we should change this to avoid those suppressed warnings

Mewel (changes requested):

mycore-base/src/main/java/org/mycore/common/xml/MCRURIResolver.java
yagee-de: Allows to remove a redundant method with the same signature in the same file. It's used in "mcrobject:" for versions of an object and in xslStyle: and xslFormat: where you can define XSLT parameters. So the risk of this to accidentally result in a different behavior is minimal.

mycore-iview2/src/main/java/org/mycore/iview2/services/MCRTilingQueue.java
yagee-de: The problem is, that MCRTilingQueue implements a Collection interface therefor the SuppressWarnings is necessary here

Mewel (approved):

PR #2354, opened 2025-01-13
MCR-3231 enable PMD rule MissingOverride

yagee-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3131).

This pull request includes changes to add `@Override` annotations to methods. These changes ensure that methods correctly override methods from their respective interfaces or superclasses, improving code clarity and maintainability.


yagee-de:
In theory this could be merged without review, as only `@Override` is added where a method is overridden. PMD ensures that no method is skipped and compiler ensures that every annotated method actually overrides a method in the super class.

PR #2361, opened 2025-01-15
MCR-3291 PMD rule RedundantFieldInitializer

Mewel:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3291).


golsch (approved):
Looks good. If there is no rule for this, perhaps we should think about standardizing the initialization of the singletons, including the variables and names. For me, the names should not be in `UPPER_CASE`.

Mewel:
> Looks good. If there is no rule for this, perhaps we should think about standardizing the initialization of the singletons, including the variables and names. For me, the names should not be in `UPPER_CASE`.

True, not sure if there is a PMD rule for that.

PR #2370, opened 2025-01-21
MCR-3256 Enable PMD rule CognitiveComplexity for level 30

Mewel:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3256).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/access/facts/MCRFactsAccessSystem.java
+++ b/mycore-base/src/main/java/org/mycore/access/facts/MCRFactsAccessSystem.java
@@ -27,7 +27,6 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.TimeUnit;
 
-import jakarta.inject.Singleton;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.jdom2.Element;
@@ -53,6 +52,8 @@ import org.mycore.datamodel.classifications2.MCRCategoryDAOFactory;
 import org.mycore.datamodel.classifications2.MCRCategoryID;
 import org.mycore.datamodel.metadata.MCRMetadataManager;
 import org.mycore.datamodel.metadata.MCRObjectID;
+
+import jakarta.inject.Singleton;
 
 /**
  * base class for XML fact based access system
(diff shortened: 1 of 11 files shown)


mycore-base/src/main/java/org/mycore/access/facts/MCRFactsAccessSystem.java
```diff
@@ -205,11 +183,39 @@ public boolean checkPermission(String checkID, String permission, List<MCRFact>
...
+                    facts.add(new MCRObjectIDFact("objid", checkID, mcrId));
+                }
+            } else {
+                LOGGER.debug(() -> "There is no object or derivate with id " + mcrId + " in metadata store");
+            }
+        } else {
+            target = "unknown";
+        }
+        return target;
+    }
```
golsch: may return target directly, refactor facts part into seperate method `#computeFacts(checkID, MCRFactsHolder facts)` and call it from `#checkPermission`.

mycore-mets/src/main/java/org/mycore/mets/tools/MCRMetsSave.java
```diff
@@ -949,4 +776,211 @@ public static String getFileId(MCRPath path) {
...
+            // modify logical struct Map
+            String logID = linkToRemove.getFrom();
+            // the deleted file was not directly assigned to a structure
+            if (logicalStructMap.getDivContainer().getId().equals(logID)) {
+                return;
+            }
+            LogicalDiv logicalDiv = logicalStructMap.getDivContainer().getLogicalSubDiv(logID);
+            if (logicalDiv == null) {
+                LOGGER.error("Could not find 'LogicalDiv' with id '{}'.", () -> logID);
+                return;
```
golsch, line 949:
```
if (logicalDiv == null) {
    LOGGER.error("Could not find {} with id {}",
        () -> LogicalDiv.class.getSimpleName(), () -> logID);
    LOGGER.error("Mets document remains unchanged");
    return mets;
}
```
I'm not that familiar with Mets, maybe it's an optimization, but the behavior doesn't seem to be equivalent.

golsch (changes requested):

mycore-mets/src/main/java/org/mycore/mets/tools/MCRMetsSave.java
Mewel: Your right, it's not. In my opinion it was a bug before which probably never occurred. So I changed it to this version.

golsch (approved):

yagee-de:
@Mewel please fix import order

PR #2381, opened 2025-01-22
MCR-3233 Enable PMD rule bestpractices/ReplaceHashtableWithMap

rsteph-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3233).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/common/MCRSession.java
+++ b/mycore-base/src/main/java/org/mycore/common/MCRSession.java
@@ -27,7 +27,6 @@ import java.net.URI;
 import java.net.UnknownHostException;
 import java.util.ArrayDeque;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Locale;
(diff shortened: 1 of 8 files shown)


rsteph-de:
nothing more to do or say - ready to merge

yagee-de (changes requested):
Hashtable cannot be replaced by HashMap if used by mutliple threads. Please use ConcurrentHashMap if unsure and HashMap only if you are absolutely sure that missing synchronization is not an issue.

mycore-base/src/main/java/org/mycore/common/MCRSession.java
```diff
@@ -69,7 +69,7 @@ sealed public class MCRSession implements Cloneable permits MCRScopedSession {
     private static final URI DEFAULT_URI = URI.create("");
 
     /** A map storing arbitrary session data * */
-    private Map<Object, Object> map = new Hashtable<>();
+    private Map<Object, Object> map = new HashMap<>();
```
yagee-de: Pretty sure that this need proper handling for multiple threads. Use ConcurrentHashMao here

rsteph-de: You are right ... I changed possible candidates for multithreaded access with ConcurrentHashMap
(MCRSession and MCRAccessControlSystem)

I'm not sure about the "data"-Map in MCREvent. That should not be an issue?

The remaining maps are only used within methods and should be safe.

yagee-de (approved):

yagee-de (approved):

PR #2387, opened 2025-01-24
MCR-3322 Expiration ignored during partial update of access keys when deleted

golsch:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3322).


rsteph-de (approved):
changed like previously discussed in web conference

PR #2390, opened 2025-01-27
MCR-3235 Enable PMD rule SystemPrintln

Mewel:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3235).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/common/config/MCRConfigurationDirSetup.java
+++ b/mycore-base/src/main/java/org/mycore/common/config/MCRConfigurationDirSetup.java
@@ -36,7 +36,6 @@ import java.util.stream.Stream;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.core.LoggerContext;
-import org.apache.logging.log4j.status.StatusLogger;
 import org.apache.logging.log4j.web.Log4jServletContainerInitializer;
 import org.mycore.common.MCRClassTools;
 import org.mycore.common.MCRSessionMgr;
@@ -54,9 +53,8 @@ import jakarta.servlet.ServletException;
  * @author Thomas Scheffler (yagee)
  * @since 2013.12
  */
+@SuppressWarnings("PMD.SystemPrintln")
 public class MCRConfigurationDirSetup implements AutoExecutable {
-
-    private static final StatusLogger LOGGER = StatusLogger.getLogger();
 
     public static void loadExternalLibs() {
         File resourceDir = MCRConfigurationDir.getConfigFile("resources");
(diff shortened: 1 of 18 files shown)


Mewel:
I'm unsure of how to handle our CLI classes. We have a great mix there between System.out.* and LOGGER. There is also some cases where we read from the stdin. Any suggestions?

toKrause (approved):

golsch:
> I'm unsure of how to handle our CLI classes. We have a great mix there between System.out.* and LOGGER. There is also some cases where we read from the stdin. Any suggestions?

I think we should always output CLI messages directly via System.*.println / std if they are useful for the user... Also, its analogous to stdin.

mycore-base/src/main/java/org/mycore/common/config/MCRConfigurationDirSetup.java
```diff
@@ -54,10 +53,9 @@
  * @author Thomas Scheffler (yagee)
  * @since 2013.12
  */
+@SuppressWarnings("PMD.SystemPrintln")
 public class MCRConfigurationDirSetup implements AutoExecutable {
 
-    private static final StatusLogger LOGGER = StatusLogger.getLogger();
```
toKrause, line 60: Why not use the `StatusLogger` here (as in other classes that work in a pre-configured state of MyCoRe)? That seemed to be the better solution and is working fine.

toKrause (changes requested):

mycore-base/src/main/java/org/mycore/common/config/MCRConfigurationDirSetup.java
Mewel: I just tested it and it doesn't work.
That's why I replaced it.

toKrause: It works perfectly fine here, just tested it. So there might be something else going on in your setup, that prevents the log messages from being shown?

If this is a general problem, then all usages of `StatusLogger` should probably be replaced.

Mewel: Hm strange, I tested with an Ubuntu 24.04 and MIR 2024.06.01.
Apparently it also didn't work for @yagee-de back then, that's probably why the StatusLogger was just used once in the entire class before.

Mewel:
I used the normal Logger in MCRComponent cause it works. I changed the StatusLogger in MCRResourceResolver to a normal Logger and used System.println for the factory.
The logs work now as intended.

Two cents from chatgpt:

`StatusLogger` is an internal Log4j component used to log status messages about Log4j's own initialization and configuration. It is **not** intended for application logging.

## When to Use `StatusLogger`
- **Debugging Log4j configuration issues**: If Log4j is not working as expected, you can enable `StatusLogger` to see internal status messages.
- **Checking Log4j initialization problems**: If your logs are missing or Log4j isn't behaving correctly, `StatusLogger` might help diagnose the issue.

toKrause (approved):

PR #2392, opened 2025-01-28
MCR-3255 Enable PMD rule ClassWithOnlyPrivateConstructorsShouldBeFinal

Mewel:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3255).


golsch:
I find the rule useful for factories, instance holders and helper classes. I'm not sure about exceptions and classes such as clients that could potentially be extended. Of course, it is actually useful in combination with the singleton pattern, but this approach should generally be questioned.

Mewel:
> I find the rule useful for factories, instance holders and helper classes. I'm not sure about exceptions and classes such as clients that could potentially be extended. Of course, it is actually useful in combination with the singleton pattern, but this approach should generally be questioned.

I agree. The only exception this changes is MCRAccessException. Which Client classes would you exclude from this rule?

toKrause (approved):

PR #2397, opened 2025-01-31
MCR-3242 Fix restapi enum query params

golsch:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3242).

The REST API no longer boots up because the values do not match the enum...


golsch:
Converters would probably be the more elegant solution, but I would find this too overkill for this particular use case. Perhaps a fallback could be considered if the params are not suitable.

sebhofmann (approved):

yagee-de (changes requested):
Changes REST-API. API should be stable

yagee-de:
There are many more enums that need to be fixed. I will open another PR for all enums.

yagee-de:
fix is part of https://github.com/MyCoRe-Org/mycore/pull/2398

PR #2404, opened 2025-02-10
MCR-3277 Enable PMD rule SingletonClassReturningNewInstance

Mewel:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3277).

My naming convention for this PR:
createInstance() -> if a new instance of this class is created without any parameter
createInstanceOf(param...) -> if a new instance of this class is create with parameters
getInstanceOf(param) -> used if an instance is created or get from a cache

### DIFF
--- a/mycore-base/src/main/java/org/mycore/common/config/MCRConfigurableInstanceHelper.java
+++ b/mycore-base/src/main/java/org/mycore/common/config/MCRConfigurableInstanceHelper.java
@@ -146,9 +146,9 @@ class MCRConfigurableInstanceHelper {
         MCRConfigurationProxy productAnnotation = targetClass.getDeclaredAnnotation(MCRConfigurationProxy.class);
         if (productAnnotation != null) {
             Class<Supplier<T>> proxyClass = (Class<Supplier<T>>) productAnnotation.proxyClass();
-            return getInfo(proxyClass).getInstance(configuration).get();
+            return getInfo(proxyClass).getInstanceOf(configuration).get();
         } else {
-            return getInfo(targetClass).getInstance(configuration);
+            return getInfo(targetClass).getInstanceOf(configuration);
         }
     }
 
@@ -378,7 +378,7 @@ class MCRConfigurableInstanceHelper {
 
         }
 
-        public T getInstance(MCRInstanceConfiguration configuration) {
+        public T getInstanceOf(MCRInstanceConfiguration configuration) {
             T instance = factory.get();
             for (Injector<T, ?> injector : injectors) {
                 injector.inject(instance, configuration);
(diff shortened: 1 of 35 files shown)


toKrause (commented):
The intent of rule "SingletonClassReturningNewInstance" is sensible and it is certainly a good thing, that we decided to include the rule. However, the [implementation](https://github.com/pmd/pmd/blob/main/pmd-java/src/main/java/net/sourceforge/pmd/lang/java/rule/errorprone/SingletonClassReturningNewInstanceRule.java) of this rule is a bit lackluster. It assumes, that every class providing a method named `getInstance` is intended to be a  singleton. I'm not sure, that this implementation is sensible. This is also at odds with long-standing MyCoRe design principle to call static factory methods `getInstance`. This conflict leads to the huge amount of modified files modified in this PR.

Since this PR will modify a lot of existing static factory methods, we could use this opportunity to formulate a proper MyCoRe design guideline for factory methods. 
I think we should discuss this in today's web conference, as this is a larger-scope change to MyCoRe and the MyCoRe design principles. 



A few notes:
- I'm not sure that callers generally should or aught to know, whether they will always receive a newly instantiated or a potentially reused instance, i.e. if the differentiation between `createInstanceOf` and `getInstanceOf` is useful. A class providing a `createInstanceOf` method can't easily be changed to cache the provided iinstances, without changing the API of this class.
- If there are cases where it is necessary that the returned instance is a fresh instance, such methods are often called `newInstance` in Java.
- Currently, there are a lot of classes that use a `instance()` method, some of which are true singletons, some are just singletons in order to provide a shared instance (for convenience or performance, i.e. `MCRLDAPClient`, `MCRObjectCacheFactory`, `MCRLayoutService`, `MCRSolrPathDocumentFactory`, ...)
 
Suggestion for a design guideline for static factory methods:

- `getInstance()`: accessor method for true singleton classes
- `getSharedInstance()`: accessor method for classes that aren't singleton, but provide a shared instance anyways (for convenience or performance, i.e. `MCRResourceResolver`)
- `createInstance()` or `createInstance(...)`: static factory methods where the contract necessarily guarantees that a freshly created instance is returned
- `getOrCreateInstance()` or `getOrCreateInstance(...)`: static factory methods that may or may not return a freshly created instance
- `createInstanceFromConfiguration()`: static factory methods that uses the current configuration instead of provided parameters (i.e. `MCRJobQueueCleaner`)
- `getInstance(...)`, `getSharedInstance(...)` or `createInstanceFromConfiguration()`: never

mycore-base/src/main/java/org/mycore/common/config/MCRConfigurableInstanceHelper.java
```diff
@@ -378,7 +378,7 @@ private Supplier<T> createFactoryMethodFactory(Method method) {
 
         }
 
-        public T getInstance(MCRInstanceConfiguration configuration) {
+        public T getInstanceOf(MCRInstanceConfiguration configuration) {
```
toKrause, line 381: The change here is not necessary (this is a method of an inner helper class) or should be coordinated with the methods in line 125 and 145.

toKrause:
This PR has become obsolete by MCR-3348.

PR #2410, opened 2025-02-13
MCR-3269 enable PMD rule MissingSerialVersionUID

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3269).

Also added missing `@Serial` annotation to each `serialVersionUID` and unified the values to `1L`, since serialization isn't used in MyCoRe (in a way hat needs class versioning), only a hand full of classes had actual values and the values were never updated when implementations changed.


sebhofmann (approved):

yagee-de:
Target branch should be 2024.06.x for this ticket

PR #2424, opened 2025-02-20
MCR-3339 Ensure usage of MCRResourceResolver with custom PMD rule

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3339).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/tools/MyCoReWebPageProvider.java
+++ b/mycore-base/src/main/java/org/mycore/tools/MyCoReWebPageProvider.java
@@ -37,6 +37,7 @@ import org.jdom2.Namespace;
 import org.jdom2.Text;
 import org.jdom2.input.SAXBuilder;
 import org.mycore.frontend.servlets.MCRServlet;
+import org.mycore.resource.MCRResourceHelper;
 import org.xml.sax.InputSource;
 
 /**
@@ -56,6 +57,7 @@ import org.xml.sax.InputSource;
  * @author shermann
  * @author Matthias Eichner
  */
+@SuppressWarnings("PMD.MCR.ResourceResolver")
 public class MyCoReWebPageProvider {
 
     /** German language key */
@@ -112,7 +114,7 @@ public class MyCoReWebPageProvider {
         SAXBuilder saxBuilder = new SAXBuilder();
         saxBuilder.setEntityResolver((publicId, systemId) -> {
             String resource = systemId.substring(systemId.lastIndexOf('/'));
-            InputStream is = getClass().getResourceAsStream(resource);
+            InputStream is = MCRResourceHelper.getResourceAsStream(resource);
             if (is == null) {
                 throw new IOException(new FileNotFoundException("Unable to locate resource " + resource));
             }
--- a/mycore-webtools/src/main/java/org/mycore/webtools/pdf/MCRXHTML2PDFTransformer.java
+++ b/mycore-webtools/src/main/java/org/mycore/webtools/pdf/MCRXHTML2PDFTransformer.java
@@ -23,12 +23,12 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.text.Normalizer;
 
-import org.mycore.common.MCRClassTools;
 import org.mycore.common.config.MCRConfiguration2;
 import org.mycore.common.config.MCRConfigurationException;
 import org.mycore.common.content.MCRByteContent;
 import org.mycore.common.content.MCRContent;
 import org.mycore.common.content.transformer.MCRContentTransformer;
+import org.mycore.resource.MCRResourceHelper;
 
 import com.itextpdf.html2pdf.ConverterProperties;
 import com.itextpdf.html2pdf.HtmlConverter;
@@ -67,12 +67,11 @@ public class MCRXHTML2PDFTransformer extends MCRContentTransformer {
     protected synchronized FontProvider lazyInitFontProvider() {
         if (fontProvider == null) {
             fontProvider = new FontProvider();
-            final ClassLoader cl = MCRClassTools.getClassLoader();
             MCRConfiguration2.getString("MCR.ContentTransformer." + id + ".FontResources")
                 .stream()
                 .flatMap(MCRConfiguration2::splitValue)
                 .forEach((path) -> {
-                    try (InputStream is = cl.getResource(path).openStream()) {
+                    try (InputStream is = MCRResourceHelper.getResourceAsStream(path)) {
                         fontProvider.addFont(is.readAllBytes());
                     } catch (IOException e) {
                         throw new MCRConfigurationException("Error while loading configured fonts!", e);
--- a/mycore-webtools/src/main/java/org/mycore/webtools/vue/MCRVueRootServlet.java
+++ b/mycore-webtools/src/main/java/org/mycore/webtools/vue/MCRVueRootServlet.java
@@ -46,6 +46,7 @@ import org.mycore.common.content.MCRJDOMContent;
 import org.mycore.common.content.MCRURLContent;
 import org.mycore.frontend.MCRFrontendUtil;
 import org.mycore.frontend.servlets.MCRContentServlet;
+import org.mycore.resource.MCRResourceHelper;
 import org.mycore.tools.MyCoReWebPageProvider;
 import org.xml.sax.SAXException;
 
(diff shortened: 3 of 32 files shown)


yagee-de:
please resolve conflict

toKrause:
rebased

mycore-webtools/src/main/java/org/mycore/webtools/vue/MCRVueRootServlet.java
```diff
@@ -132,26 +133,30 @@ public MCRContent getContent(HttpServletRequest req, HttpServletResponse resp) t
...
-                 * redirect to this /404 route the get the actual 404 Code.
-                 * see also https://www.youtube.com/watch?v=vjj8B4sq0UI&t=1815s
-                 * */
-                resp.setStatus(HttpServletResponse.SC_NOT_FOUND);
-            }
-            try {
-                return getLayoutService().getTransformedContent(req, resp, new MCRJDOMContent(mycoreWebpage));
-            } catch (TransformerException | SAXException e) {
-                throw new IOException(e);
+        if (!pathInfo.endsWith("/") && !pathInfo.endsWith(indexHtmlPage)) {
```
golsch: `pathInfo` is not null-safe. Previously, a null `pathInfo` apparently led to a null `resource`, so there was probably no (wanted?) problem. `/`as `pathInfo` is problematic too and leads to a null `indexResource`.
@sebhofmann, what is your opinion?

golsch (changes requested):

mycore-webtools/src/main/java/org/mycore/webtools/vue/MCRVueRootServlet.java
toKrause: You're right. Previously, the `pathInfo` was used without null-check in line 135 and 137, so I assumed, wrongly, that it can't be `null` here. But the concatenation in line 135 leads, in almost all cases, to a resource path that won't match anything and therefore resolves to a `null` resource URL. This safety feature seems to be more of a coincidence then a design choice.

mycore-webtools/src/main/java/org/mycore/webtools/vue/MCRVueRootServlet.java
```diff
@@ -132,26 +133,30 @@ public MCRContent getContent(HttpServletRequest req, HttpServletResponse resp) t
...
-                 * see also https://www.youtube.com/watch?v=vjj8B4sq0UI&t=1815s
-                 * */
-                resp.setStatus(HttpServletResponse.SC_NOT_FOUND);
-            }
-            try {
-                return getLayoutService().getTransformedContent(req, resp, new MCRJDOMContent(mycoreWebpage));
-            } catch (TransformerException | SAXException e) {
-                throw new IOException(e);
+        if (pathInfo != null && !pathInfo.endsWith("/") && !pathInfo.endsWith(indexHtmlPage)) {
+            URL resource = MCRResourceHelper.getResourceUrl(req.getServletPath() + pathInfo);
```
golsch: Unfortunately, this does not work. The code does not seem to be equivalent, `resource` is `null`.

mycore-webtools/src/main/java/org/mycore/webtools/vue/MCRVueRootServlet.java
```diff
@@ -132,26 +133,30 @@ public MCRContent getContent(HttpServletRequest req, HttpServletResponse resp) t
...
+            if (resource != null) {
+                return new MCRURLContent(resource);
             }
         }
+
+        URL indexResource = MCRResourceHelper.getResourceUrl(indexHtmlPath);
```
golsch: See above, `indexResource` is `null`.

golsch (changes requested):

mycore-webtools/src/main/java/org/mycore/webtools/vue/MCRVueRootServlet.java
toKrause: I'll look into it. Can you tell me the easiest way to test this?

golsch: I test in MIR with the job queue manager via `/jobqueue`. I already suspected the paths with the starting `/`, but that doesn't solve the problem either.

toKrause: Simple mistake. `ServletContext` only serves web resources, so we need to resolve a web resource (i.e. a resource with a path starting with `META-INF/resources`) using `MCRResourceHelper#getWebResourceUrl`, not `MCRResourceHelper.getResourceUrl`. Thx for spotting the issue.

golsch (approved):
Vue part looks good for me.

mycore-base/src/main/java/org/mycore/tools/MyCoReWebPageProvider.java
```diff
@@ -112,7 +114,7 @@ public Element addSection(String title, String xmlAsString, String lang) throws
         SAXBuilder saxBuilder = new SAXBuilder();
         saxBuilder.setEntityResolver((publicId, systemId) -> {
             String resource = systemId.substring(systemId.lastIndexOf('/'));
-            InputStream is = getClass().getResourceAsStream(resource);
+            InputStream is = MCRResourceHelper.getResourceAsStream(resource);
```
rsteph-de, line 117: what about the leading "/" , in code above it was removed

mycore-webtools/src/main/java/org/mycore/webtools/pdf/MCRXHTML2PDFTransformer.java
```diff
@@ -67,12 +67,11 @@ public MCRContent transform(MCRContent source) throws IOException {
     protected synchronized FontProvider lazyInitFontProvider() {
         if (fontProvider == null) {
             fontProvider = new FontProvider();
-            final ClassLoader cl = MCRClassTools.getClassLoader();
             MCRConfiguration2.getString("MCR.ContentTransformer." + id + ".FontResources")
                 .stream()
                 .flatMap(MCRConfiguration2::splitValue)
                 .forEach((path) -> {
-                    try (InputStream is = cl.getResource(path).openStream()) {
+                    try (InputStream is = MCRResourceHelper.getResourceUrl(path).openStream()) {
```
rsteph-de: why not getResourceAsStream()?

rsteph-de (changes requested):
2 little questions/remarks

mycore-base/src/main/java/org/mycore/tools/MyCoReWebPageProvider.java
toKrause: `ResourceResolver` treats all paths as absolute paths and adds or removes leading slashes as required, so there is no need to add or remove them manually before resolving a resource. Here, the code can remain simpler by not doing additional indexing math.

There should be no code left that explicitly adds or removes leading slashes before resolving resources.

mycore-webtools/src/main/java/org/mycore/webtools/pdf/MCRXHTML2PDFTransformer.java
toKrause: Also a good point. I've also double-checked all other usages of `MCRREsourceHelper` for similar usages, but found none.

mycore-base/src/main/java/org/mycore/tools/MyCoReWebPageProvider.java
rsteph-de: -> `systemId.substring(systemId.lastIndexOf('/')+1);`

toKrause: That would indeed remove the leading slash, but makes absolutely no difference API-wise (technically, `MCRResourcePath` will add it back, so keeping it is ever so slightly more efficient 🙂)

toKrause: Or in other words: `systemId.substring(systemId.lastIndexOf('/'));` is simpler than `systemId.substring(systemId.lastIndexOf('/')+1);` and therefore preferable.

rsteph-de (approved):
looks good for me now

PR #2431, opened 2025-03-04
MCR-3345 improve MCRVueRootServlet

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3345).

### DIFF
--- a/mycore-webtools/src/main/java/org/mycore/webtools/vue/MCRVueRootServlet.java
+++ b/mycore-webtools/src/main/java/org/mycore/webtools/vue/MCRVueRootServlet.java
@@ -24,15 +24,23 @@ import java.io.Serial;
 import java.io.StringReader;
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
+import java.util.AbstractMap;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
+import java.util.Optional;
 import java.util.Properties;
+import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import javax.xml.transform.TransformerException;
 
+import jakarta.servlet.ServletConfig;
+import jakarta.servlet.ServletException;
 import org.jdom2.Content;
 import org.jdom2.Element;
 import org.jdom2.JDOMException;
@@ -41,11 +49,14 @@ import org.jsoup.Jsoup;
 import org.jsoup.nodes.Document;
 import org.jsoup.nodes.Entities;
 import org.mycore.common.MCRException;
+import org.mycore.common.MCRSessionMgr;
+import org.mycore.common.config.MCRConfiguration2;
 import org.mycore.common.content.MCRContent;
 import org.mycore.common.content.MCRJDOMContent;
 import org.mycore.common.content.MCRURLContent;
 import org.mycore.frontend.MCRFrontendUtil;
 import org.mycore.frontend.servlets.MCRContentServlet;
+import org.mycore.services.i18n.MCRTranslation;
 import org.mycore.tools.MyCoReWebPageProvider;
 import org.xml.sax.SAXException;
 
@@ -60,6 +71,14 @@ import jakarta.servlet.http.HttpServletResponse;
  * &lt;servlet&gt;
  *   &lt;servlet-name&gt;MCRVueRootServlet&lt;/servlet-name&gt;
  *   &lt;servlet-class&gt;org.mycore.webtools.vue.MCRVueRootServlet&lt;/servlet-class&gt;
+ *   &lt;init-param&gt;
+ *   &lt;param-name&gt;heading&lt;/param-name&gt;
+ *   &lt;param-value&gt;webtools.texteditor.heading&lt;/param-value&gt;
+ *   &lt;/init-param&gt;
+ *   &lt;init-param&gt;
+ *   &lt;param-name&gt;properties&lt;/param-name&gt;
+ *   &lt;param-value&gt;MCR.Webtools.Special.Property&lt;/param-value&gt;
+ *   &lt;/init-param&gt;
  * &lt;/servlet&gt;
  * &lt;servlet-mapping&gt;
  *   &lt;servlet-name&gt;MCRVueRootServlet&lt;/servlet-name&gt;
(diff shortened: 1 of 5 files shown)


mycore-webtools/src/main/java/org/mycore/webtools/vue/MCRVueRootServlet.java
```diff
@@ -127,6 +146,43 @@ public class MCRVueRootServlet extends MCRContentServlet {
...
+    private static final Set<Integer> PASSTHROUGH_HTTP_STATUS_CODE_LIST = Set.of(
+        HttpServletResponse.SC_NOT_FOUND,
+        HttpServletResponse.SC_UNAUTHORIZED,
+        HttpServletResponse.SC_FORBIDDEN
+    );
+
+    @Override
+    public void init(ServletConfig config) throws ServletException {
+        super.init(config);
+        headingI18nKey = config.getInitParameter(HEADING_INIT_PARAM);
```
golsch, line 168: Perhaps we should enable a fallback here or at least display a warning if the value is not set.
This is because there may be a null pointer exception when rendering because at least the existence of the translation is checked.

mycore-webtools/src/main/java/org/mycore/webtools/vue/MCRVueRootServlet.java
```diff
@@ -184,7 +245,17 @@ protected org.jdom2.Document getIndexDocument(URL indexResource, String absolute
...
+
+        MCRConfiguration2.getOrThrow(VUE_PASSTHROUGH_PROPERTY_NAME, MCRConfiguration2::splitValue)
+                .map(k -> new AbstractMap.SimpleImmutableEntry<>(k, MCRConfiguration2.getString(k)) {})
+                .filter(e -> e.getValue().isPresent())
+                .forEach(e -> properties.setProperty(e.getKey(), e.getValue().get()));
+
+        properties.putAll(servletInitProperties);
+
         properties.setProperty("webApplicationBaseURL", MCRFrontendUtil.getBaseURL());
+        properties.setProperty("CurrentLang", MCRSessionMgr.getCurrentSession().getCurrentLanguage());
```
golsch: `CurrentLang` is inconsistent with `webApplicationBaseURL`, may change to `currentLang`.

mycore-webtools/src/main/java/org/mycore/webtools/vue/MCRVueRootServlet.java
```diff
@@ -127,6 +146,43 @@ public class MCRVueRootServlet extends MCRContentServlet {
...
+    protected static final String VUE_PASSTHROUGH_PROPERTY_NAME = "MCR.Vue.Properties";
+
+    protected static final String HEADING_INIT_PARAM = "heading";
+    protected static final String PROPERTIES_INIT_PARAM = "properties";
+
+    protected String headingI18nKey;
+    protected Set<String> servletInitPropertiesKeys;
+    private Map<String, String> servletInitProperties;
+
+    private static final Set<Integer> PASSTHROUGH_HTTP_STATUS_CODE_LIST = Set.of(
```
golsch, line 159: My naive thought is that maybe 400 (or 500) also has to be added or not? Of course it should not occur because of client-side validation, but only if implemented.

mycore-webtools/src/main/java/org/mycore/webtools/vue/MCRVueRootServlet.java
```diff
@@ -184,7 +245,17 @@ protected org.jdom2.Document getIndexDocument(URL indexResource, String absolute
      */
     protected Properties getProperties() {
         Properties properties = new Properties();
+
+        MCRConfiguration2.getOrThrow(VUE_PASSTHROUGH_PROPERTY_NAME, MCRConfiguration2::splitValue)
+                .map(k -> new AbstractMap.SimpleImmutableEntry<>(k, MCRConfiguration2.getString(k)) {})
```
golsch: `... (k, MCRConfiguration2.getString(k)) {}` - I don't know the `{}` notation or rather my eclipse cries.

mycore-webtools/src/main/java/org/mycore/webtools/vue/MCRVueRootServlet.java
```diff
@@ -184,7 +245,17 @@ protected org.jdom2.Document getIndexDocument(URL indexResource, String absolute
...
+                .filter(e -> e.getValue().isPresent())
```
golsch: It might be useful to include empty properties to make it more explicit and consistent.

golsch (changes requested):

mycore-webtools/src/main/java/org/mycore/webtools/vue/MCRVueRootServlet.java
sebhofmann: I tried to use the same variable name used in org.mycore.common.xsl.MCRParameterCollector#setUnmodifyableParameters (used for all xsl stylesheets)

mycore-webtools/src/main/java/org/mycore/webtools/vue/MCRVueRootServlet.java
sebhofmann: I would only passthrough stuff, which is relevant to crawlers.

mycore-webtools/src/main/java/org/mycore/webtools/vue/MCRVueRootServlet.java
golsch: Ok, but in the Javascript world it is rather camelCase or `currentLang`.

mycore-webtools/src/main/java/org/mycore/webtools/vue/MCRVueRootServlet.java
sebhofmann: ill add a null check before the translation

mycore-webtools/src/main/java/org/mycore/webtools/vue/MCRVueRootServlet.java
```diff
@@ -127,6 +146,43 @@ public class MCRVueRootServlet extends MCRContentServlet {
...
+
+        updateInitProperties();
+
+        MCRConfiguration2.addPropertyChangeEventLister(servletInitPropertiesKeys::contains, this::changeProperty);
+    }
+
+    protected void updateInitProperties() {
+        servletInitProperties = new HashMap<>();
+        servletInitPropertiesKeys.forEach(key -> {
+            MCRConfiguration2.getString(key).ifPresent(value -> servletInitProperties.put(key, value));
```
golsch: Unfortunately this does not work, `MCRConfiguration2#getString()` returns an empty Optional if the value is empty resp. this should be analogous to the `MCR.Vue.Properties` approach.

mycore-webtools/src/main/java/org/mycore/webtools/vue/MCRVueRootServlet.java
```diff
@@ -184,7 +245,16 @@ protected org.jdom2.Document getIndexDocument(URL indexResource, String absolute
      */
     protected Properties getProperties() {
         Properties properties = new Properties();
+
+        MCRConfiguration2.getOrThrow(VUE_PASSTHROUGH_PROPERTY_NAME, MCRConfiguration2::splitValue)
+                .map(k -> new AbstractMap.SimpleImmutableEntry<>(k, MCRConfiguration2.getString(k)))
+                .forEach(e -> properties.setProperty(e.getKey(), e.getValue().orElse(null)));
```
golsch: `null` as a value is problematic and leads to a NPE.

mycore-webtools/src/main/java/org/mycore/webtools/vue/MCRVueRootServlet.java
```diff
@@ -237,8 +311,15 @@ protected Element buildPropertiesScript() {
         propertiesScript.setAttribute("type", "text/javascript");
         StringBuilder propertiesJson = new StringBuilder("var mycore = mycore || {};");
         propertiesJson.append(System.lineSeparator());
-        getProperties().forEach((key, value) -> propertiesJson.append("mycore.").append(key).append("=\"").append(value)
-            .append("\";").append(System.lineSeparator()));
+        getProperties().forEach((key, value) -> {
+            propertiesJson.append("mycore[\"").append(key)
+                    .append("\"]").append('=');
+            if(value == null) {
+                propertiesJson.append("null;");
```
golsch: In the Javascript context, `null` is of course correct. I'm not sure whether an empty string would make sense at this point.

golsch (changes requested):

mycore-webtools/src/main/java/org/mycore/webtools/vue/MCRVueRootServlet.java
sebhofmann: So maybe filter for not empty value here and remove null code in 317-319, then it should be the same behaviour as in updateInitProperties?

golsch: I would be in favor of passing the properties one-to-one at the level and using empty strings instead of null. Then it is up to the application to consider empty strings as null.

mycore.properties
PropertyA=test
PropertyB=

leads to
mycore['PropertyA']='test'
mycore['PropertyB']=''

mycore-webtools/src/main/java/org/mycore/webtools/vue/MCRVueRootServlet.java
```diff
@@ -127,6 +146,54 @@ public class MCRVueRootServlet extends MCRContentServlet {
...
+
+    /**
+     * Resolves the property value from the configuration. If the property is not found, an exception is thrown.
+     * If the property is found, but the value is empty, then the value is still returned.
+     * {@link MCRConfiguration2#getString(String)} would return an empty optional in this case.
+     * @param key the property key
+     * @return the property value, even if it is an empty string
+     */
+    private static String resolveProperty(String key) {
+        return MCRConfiguration2.getOrThrow(key, o -> o);
```
golsch: An empty value leads to an exception, possibly `MCRConfiguration2#getPropertiesMap` should be used.

golsch (changes requested):

golsch (approved):

PR #2433, opened 2025-03-06
MCR-3347 fix xml to json serialization

Mewel:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3347).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/common/content/transformer/MCRXsonifyTransformer.java
+++ b/mycore-base/src/main/java/org/mycore/common/content/transformer/MCRXsonifyTransformer.java
@@ -20,7 +20,9 @@ import static org.mycore.xsonify.serialize.SerializerSettings.PrefixHandling;
 import static org.mycore.xsonify.serialize.SerializerSettings.XsAnyNamespaceStrategy;
 
 import java.io.IOException;
+import java.util.List;
 import java.util.Locale;
+import java.util.Map;
 
 import javax.xml.parsers.ParserConfigurationException;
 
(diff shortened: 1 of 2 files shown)


mycore-base/src/main/java/org/mycore/common/content/transformer/MCRXsonifyTransformer.java
```diff
@@ -1,9 +1,29 @@
 package org.mycore.common.content.transformer;
 
-import com.fasterxml.jackson.databind.JsonNode;
-import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.databind.node.ObjectNode;
-import jakarta.inject.Singleton;
+import static org.mycore.xsonify.serialize.SerializerSettings.AdditionalNamespaceDeclarationStrategy;
```
toKrause, line 3: Could you undo the import statement reordering?

rsteph-de: "My eclipse" (configured like documented on MyCoRe website)
 sorts all static imports before "normal" imports, so the change is O.K. for me.

Mewel: Same here, I also use the eclipse code formatter for intellij. And now it seems correct.

Another unrelated question @toKrause 
I still have this disabled code in the class

```
    /**
     * Map of namespaces. This setting is required for the json2xml serialisation process if the
     * {@link SerializerSettings#namespaceDeclaration()} is set to <b>OMIT</b>.
     * TODO: Map property doesn't seem to work right now
     */
    //@MCRProperty(name = "Namespaces", required = false)
    //public Map<String, String> namespaces;

```

Should this work now?

toKrause: A feature like that has not yet been added. I've just made [MCR-3362](https://github.com/MyCoRe-Org/mycore/pull/2451) that would allow for this. Code here would be

``` java
@MCRProperty(name = "Namespaces.*", required = false)
public Map<String, String> namespaces;
```

and contain keys with the `Namespaces.` prefix already removed.

Mewel: Ok wow cool, das ging schnell. Dank dir :)

toKrause (approved):

toKrause (approved):

PR #2434, opened 2025-03-07
MCR-3348 Implement static factory method naming convention

toKrause:

### DIFF
--- a/mycore-base/src/main/java/org/mycore/datamodel/classifications2/MCRCategoryID.java
+++ b/mycore-base/src/main/java/org/mycore/datamodel/classifications2/MCRCategoryID.java
@@ -73,17 +73,37 @@ public class MCRCategoryID implements Serializable {
     /**
      * @param rootID
      *            aka Classification ID
+     */
+    public MCRCategoryID(String rootID) {
+        this(rootID, "");
+    }
+
+    /**
+     * @param rootID
+     *            aka Classification ID
      * @param id
      *            aka Category ID
      */
     public MCRCategoryID(String rootID, String id) {
-        this();
+        super();
         setId(id);
         setRootID(rootID);
     }
 
+    /**
+     * @deprecated Use {@link #MCRCategoryID(String)}} instead
+     */
+    @Deprecated
     public static MCRCategoryID rootID(String rootID) {
-        return new MCRCategoryID(rootID, "");
+        return new MCRCategoryID(rootID);
+    }
+
+    /**
+     * @deprecated Use {@link #ofString(String)} instead
+     */
+    @Deprecated
+    public static MCRCategoryID fromString(String categoryId) {
+        return ofString(categoryId);
     }
 
     /**
@@ -92,11 +112,11 @@ public class MCRCategoryID implements Serializable {
      * @throws IllegalArgumentException if the given categoryId is invalid
      */
     @JsonCreator(mode = JsonCreator.Mode.DELEGATING)
-    public static MCRCategoryID fromString(String categoryId) {
+    public static MCRCategoryID ofString(String categoryId) {
         String[] parts = categoryId.split(":");
         try {
             return switch (parts.length) {
-                case 1 -> rootID(parts[0]);
+                case 1 -> new MCRCategoryID(parts[0]);
                 case 2 -> new MCRCategoryID(parts[0], parts[1]);
                 default -> throw new IllegalArgumentException("CategoryId is ambiguous: " + categoryId);
             };
@@ -196,7 +216,7 @@ public class MCRCategoryID implements Serializable {
      */
     public static boolean isValid(String id) {
         try {
-            fromString(id);
+            ofString(id);
             return true;
         } catch (Exception e) {
             return false;
--- a/mycore-base/src/main/java/org/mycore/frontend/jersey/filter/access/MCRResourceAccessCheckerFactory.java
+++ b/mycore-base/src/main/java/org/mycore/frontend/jersey/filter/access/MCRResourceAccessCheckerFactory.java
@@ -26,21 +26,26 @@ import java.util.concurrent.ConcurrentHashMap;
  *
  */
 public class MCRResourceAccessCheckerFactory {
-    private static Map<Class<? extends MCRResourceAccessChecker>, MCRResourceAccessChecker> implMap;
-    static {
-        //static block as code style requires 120 character line width
-        implMap = new ConcurrentHashMap<>();
-    }
 
+    private static final Map<Class<? extends MCRResourceAccessChecker>, MCRResourceAccessChecker> IMPLEMENTATION_MAP
+        = new ConcurrentHashMap<>();
+
+    @Deprecated
+    @SuppressWarnings("PMD.SingletonClassReturningNewInstance")
     public static <T extends MCRResourceAccessChecker> T getInstance(Class<T> clazz)
         throws ReflectiveOperationException {
+        return obtainInstance(clazz);
+    }
+
+    public static <T extends MCRResourceAccessChecker> T obtainInstance(Class<T> clazz)
+        throws ReflectiveOperationException {
         @SuppressWarnings("unchecked")
-        T accessChecker = (T) implMap.get(clazz);
+        T accessChecker = (T) IMPLEMENTATION_MAP.get(clazz);
         if (accessChecker != null) {
             return accessChecker;
         }
         accessChecker = clazz.getDeclaredConstructor().newInstance();
-        implMap.put(clazz, accessChecker);
+        IMPLEMENTATION_MAP.put(clazz, accessChecker);
         return accessChecker;
     }
 
--- a/mycore-fo/src/main/java/org/mycore/component/fo/common/fo/MCRFoFormatterFOP.java
+++ b/mycore-fo/src/main/java/org/mycore/component/fo/common/fo/MCRFoFormatterFOP.java
@@ -58,6 +58,7 @@ import org.mycore.common.MCRCoreVersion;
 import org.mycore.common.config.MCRConfiguration2;
 import org.mycore.common.content.MCRContent;
 import org.mycore.common.content.MCRSourceContent;
+import org.mycore.common.events.MCRServletContextHolder;
 import org.mycore.common.xml.MCRURIResolver;
 import org.mycore.common.xsl.MCRErrorListener;
 import org.mycore.resource.MCRResourceHelper;
@@ -77,7 +78,7 @@ public class MCRFoFormatterFOP implements MCRFoFormatterInterface {
     final ResourceResolver resolver = new ResourceResolver() {
         @Override
         public OutputStream getOutputStream(URI uri) throws IOException {
-            URL url = MCRURIResolver.getServletContext().getResource(uri.toString());
+            URL url = MCRServletContextHolder.getInstance().get().get().getResource(uri.toString());
             return url.openConnection().getOutputStream();
         }
 
@@ -85,7 +86,7 @@ public class MCRFoFormatterFOP implements MCRFoFormatterInterface {
         public Resource getResource(URI uri) throws IOException {
             MCRContent content;
             try {
-                content = MCRSourceContent.getInstance(uri.toString());
+                content = MCRSourceContent.createInstance(uri.toString());
                 return new Resource(uri.getScheme(), content.getInputStream());
             } catch (TransformerException e) {
                 LOGGER.error("Error while resolving uri: {}", uri);
@@ -148,8 +149,8 @@ public class MCRFoFormatterFOP implements MCRFoFormatterInterface {
             .getString("MCR.LayoutService.FoFormatter.transformerFactoryImpl")
             .map(impl -> TransformerFactory.newInstance(impl, MCRClassTools.getClassLoader()))
             .orElseGet(TransformerFactory::newInstance);
-        transformerFactory.setURIResolver(MCRURIResolver.instance());
-        transformerFactory.setErrorListener(MCRErrorListener.getInstance());
+        transformerFactory.setURIResolver(MCRURIResolver.obtainInstance());
+        transformerFactory.setErrorListener(new MCRErrorListener());
         return transformerFactory;
     }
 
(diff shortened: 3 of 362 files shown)


mycore-base/src/main/java/org/mycore/datamodel/classifications2/MCRCategoryID.java
```diff
@@ -70,20 +70,40 @@ private MCRCategoryID() {
...
      */
     public MCRCategoryID(String rootID, String id) {
-        this();
+        super();
         setId(id);
         setRootID(rootID);
     }
 
+    /**
+     * @deprecated Use {@link #MCRCategoryID(String)}} instead
```
rsteph-de, line 94: i would prefer `ofString()` as replacement instead

mycore-base/src/main/java/org/mycore/frontend/jersey/filter/access/MCRResourceAccessCheckerFactory.java
```diff
@@ -26,21 +26,26 @@
  *
  */
 public class MCRResourceAccessCheckerFactory {
-    private static Map<Class<? extends MCRResourceAccessChecker>, MCRResourceAccessChecker> implMap;
-    static {
-        //static block as code style requires 120 character line width
-        implMap = new ConcurrentHashMap<>();
-    }
 
+    private static final Map<Class<? extends MCRResourceAccessChecker>, MCRResourceAccessChecker> IMPLEMENTATIN_MAP
```
rsteph-de: `IMPLEMENTATIN_MAP` - Ich kaufe ein "O" ;-)

mycore-fo/src/main/java/org/mycore/component/fo/common/fo/MCRFoFormatterFOP.java
```diff
@@ -77,15 +78,15 @@ public class MCRFoFormatterFOP implements MCRFoFormatterInterface {
     final ResourceResolver resolver = new ResourceResolver() {
         @Override
         public OutputStream getOutputStream(URI uri) throws IOException {
-            URL url = MCRURIResolver.getServletContext().getResource(uri.toString());
+            URL url = MCRServletContextHolder.getInstance().get().get().getResource(uri.toString());
```
rsteph-de, line 81: `get().get() `looks strange, but due to Optional probably O.K.
Maybe we should implement a `getServletContext() / setServletContext()` to get/set the object directly (can be done in the follow up PR)

mycore-iview2/src/main/java/org/mycore/iview2/events/MCRIView2TilingThreadStarter.java
```diff
@@ -30,9 +30,12 @@
  * @author Thomas Scheffler (yagee)
  *
  */
-public class MCRIView2TilingThreadStarter implements MCRStartupHandler.AutoExecutable {
+public final class MCRIView2TilingThreadStarter implements MCRStartupHandler.AutoExecutable {
+
     private static final Logger LOGGER = LogManager.getLogger(MCRIView2TilingThreadStarter.class);
 
+    private static boolean started;
```
rsteph-de, line 37: holding "state" in a class variable? - maybe refactor to singleton later or rewrite as "normal" class?

mycore-iview2/src/main/java/org/mycore/iview2/frontend/MCRIView2Commands.java
```diff
@@ -358,8 +354,8 @@ private static void startMasterTilingThread() {
     @MCRCommand(syntax = "delete all tiles", help = "removes all tiles of all derivates", order = 80)
     public static void deleteAllTiles() throws IOException {
         Path storeDir = MCRIView2Tools.getTileDir();
-        Files.walkFileTree(storeDir, MCRRecursiveDeleter.instance());
-        MCRTilingQueue.getInstance().clear();
+        Files.walkFileTree(storeDir, new MCRRecursiveDeleter());
+        MCRTilingQueue.getSingletonInstance().clear();
```
rsteph-de: why is this method explicitely  named `getSingletonInstance()`?

mycore-jobqueue/src/main/java/org/mycore/services/queuedjob/MCRJobThreadStarter.java
```diff
@@ -104,14 +104,12 @@ public class MCRJobThreadStarter implements Runnable, Closeable {
...
 
-        MCRProcessableRegistry registry = MCRProcessableRegistry.getSingleInstance();
         processableCollection = new MCRProcessableDefaultCollection(getName());
         processableCollection.setProperty("activated", activated);
         processableCollection.setProperty("maxTryCount", maxTryCount);
         processableCollection.setProperty("maxJobThreadCount", maxJobThreadCount);
         processableCollection.setProperty("timeTillReset", timeTillReset.toString());
-
-        registry.register(processableCollection);
+        MCRProcessableManager.getRegistry().register(processableCollection);
```
rsteph-de: convert ProcessableManager to singleton?

mycore-mods/src/main/java/org/mycore/mods/classification/MCRTypeOfResource.java
```diff
@@ -39,33 +39,49 @@ class MCRTypeOfResource extends MCRAuthorityInfo {
...
+    @Deprecated
+    public static MCRTypeOfResource getAuthorityInfo(Element modsElement) {
+        return parseXML(modsElement);
+    }
+
     /**
      * If the given element is mods:typeOfResource, returns the MCRTypeOfResource mapping.
      */
-    public static MCRTypeOfResource getAuthorityInfo(org.jdom2.Element modsElement) {
+    public static MCRTypeOfResource parseXML(Element modsElement) {
```
rsteph-de, line 59: the switch of the default import from org.jdom2.Element to w3c.dom.Element is confusing. (The class supports both Elmement objects)

mycore-mods/src/main/java/org/mycore/mods/enrichment/MCRDataSourceFactory.java
```diff
@@ -32,42 +32,39 @@
  * Per data source, for each identifier type, there must be a pattern configured
  * that defines the URI to get the data for this type of identifier
  *
+ * @author Frank Lützenkirchen
  * @see MCRIdentifierResolver
- * 
- * As a global parameter, or per data source, it can be configured whether 
+ * <p>
```
rsteph-de: Is this intended - where is the closing  `<p>`?

mycore-ocfl/src/main/java/org/mycore/ocfl/commands/MCROCFLCommands.java
```diff
@@ -162,14 +162,14 @@ public static List<String> updateOCFLClassifications() {
...
-        final MCRCategoryID rootID = MCRCategoryID.rootID(classId);
+        final MCRCategoryID rootID = new MCRCategoryID(classId);
         MCROCFLPersistenceTransaction.addClassficationEvent(rootID, MCRAbstractMetadataVersion.UPDATED);
     }
 
     @MCRCommand(syntax = "delete ocfl classification {0}",
         help = "Delete classification {0} in the OCFL Store")
     public static void deleteOCFLClassification(String classId) {
-        final MCRCategoryID rootID = MCRCategoryID.rootID(classId);
+        final MCRCategoryID rootID = new MCRCategoryID(classId);
```
rsteph-de, line 172: Seeing the mixture of `new MCRCategoryID(classID)`for root IDs and `MCRCategoryID.ofString()`for "normal IDs
... I start to like the previous `rootID()` which makes it clear that we create a root ID.
Maybe we should make the 1arg constructor private and use a new `MCRCategoryID.rootIdFromString()` instead

mycore-solr/src/main/java/org/mycore/solr/index/file/MCRSolrPathDocumentFactory.java
```diff
@@ -43,14 +43,26 @@
  */
 public class MCRSolrPathDocumentFactory {
 
+    private static final Logger LOGGER = LogManager.getLogger(MCRSolrPathDocumentFactory.class);
```
rsteph-de: wrong logger instantiation - the LOGGER PR was merged 2 hours ago.

rsteph-de (changes requested):
good PR - some minor issues and ideas for follow up cleanups
Please change as little as possible - use a follow up PR for larger modifications

mycore-fo/src/main/java/org/mycore/component/fo/common/fo/MCRFoFormatterFOP.java
toKrause: Yes, didn't like that, but this implementation didn't take into account that the `ServletContext`, previously obtained from `MCRURIREsolver` might be `null`, should probably be fixed, but I didn't want to do that as part of this PR, so I'll leave it for now as is.

mycore-iview2/src/main/java/org/mycore/iview2/events/MCRIView2TilingThreadStarter.java
toKrause: `MCRIView2TilingThreadStarter` is a bit cursed. It currently needs to be a instantiable class, because it is a `MCRStartupHandler.AutoExecutable`, but also should be a singleton to ensure that only one tiling thread is managed. I have an idea on how to improve this, but will do that in the follow-up PR.

mycore-mods/src/main/java/org/mycore/mods/classification/MCRTypeOfResource.java
toKrause: I switched the order, (W3C first, then JDom2) so that this class has the same internal structure as otherwise extremely similar classes`MCRAuthorityAndCode` and `MCRAuthorityWithURI`.

mycore-ocfl/src/main/java/org/mycore/ocfl/commands/MCROCFLCommands.java
toKrause: Previously we had a mixture of `new MCRCategoryId(String, String)` to create a non-root ID and `MCRCategoryID.rootID(String)` to create a root ID as well as `MCRCategoryID.fromString(String)` to parse a string representation of an ID.

Now we have `new MCRCategoryId(String, String)` to create a non-root ID and `new MCRCategoryID(String)` to create a root ID as well as `MCRCategoryID.ofString(String)` to parse a string representation of an ID. This should be less muddled than before. 

---
Maybe sth. like `MCRCategoryId.childId(String, String)` to create a non-root ID and `MCRCategoryID.rootId(String)` to create a root ID as well as `MCRCategoryID.ofString(String)` to parse a string representation of an ID. But I don't know if there is a commonly accepted term for non-root IDs in the world of MyCoRe and I dont' feel like introducing one as part of this PR.

mycore-mods/src/main/java/org/mycore/mods/enrichment/MCRDataSourceFactory.java
toKrause: Changes to JacaDoc were unintentional. Will undo. That being said, closing `<p>`-tags are optional in HTML and especially discouraged in the 3.2 version of HTML used by JavaDoc.

mycore-jobqueue/src/main/java/org/mycore/services/queuedjob/MCRJobThreadStarter.java
toKrause: I was unsure if `MCRProcessableManager` and `MCRTikaManager` should be singletons or if they should just be utility classes that provide some static methods. Changed to singletons now, because this will simplify future extending.

rsteph-de (approved):
everything looks good for me,
some problems will be handled seperately

sebhofmann:
This broke the Test-Code in mycore-pi org/mycore/pi/MCRPIManagerTest.java:168
```
    @Before
    public void resetManagerInstance() {
        try {
            Field instance = MCRPIManager.class.getDeclaredField("instance");
            instance.setAccessible(true);
            instance.set(null, null);
        } catch (NoSuchFieldException | IllegalAccessException e) {
            LOGGER.error(e.getMessage(), e);
        }
    }
```

toKrause:
> This broke the Test-Code in mycore-pi org/mycore/pi/MCRPIManagerTest.java:168
> 
> ```
>     @Before
>     public void resetManagerInstance() {
>         try {
>             Field instance = MCRPIManager.class.getDeclaredField("instance");
>             instance.setAccessible(true);
>             instance.set(null, null);
>         } catch (NoSuchFieldException | IllegalAccessException e) {
>             LOGGER.error(e.getMessage(), e);
>         }
>     }
> ```

Tests still passed, so must be fine :) 

This code is very ugly. It prevents the singleton instance of `MCRPIManager` from being held in an [initialization-on-demand holder](https://en.wikipedia.org/wiki/Initialization-on-demand_holder_idiom) as well as from being instantiated as a constant (i.e. `private static final MCRPIManager INSTANCE = new MCRPIManager()`). This test code imposes restrictions on the internal structure of `MCRPIManager`. That shouldn't be the case.

Furthermore, I don't see why resetting the manager instance is done at all. If I remove `resetManagerInstance`, tests seem to work okay.

sebhofmann:
The Tests dont run locally on my mac and on my other linux test device. I used git bisect to find the commit which caused the tests to fail and it found the commit generated by this pr. 
On my pc the method `org.mycore.pi.urn.rest.MCRURNGranularRESTServiceTest#fullRegister` fails and i could not figure out why. The tests fails somwhere in `org.mycore.pi.urn.rest.MCRDerivateURNUtils#getURL` and it says the derivate does not exists. 
I ran the tests with the ide were the method `org.mycore.pi.MCRPIManagerTest#testRegistrationService` fails and i found this in the logs:
```
ERROR MCRPIManagerTest.resetManagerInstance: instance java.lang.NoSuchFieldException: instance
	at java.base/java.lang.Class.getDeclaredField(Class.java:2782)
	at org.mycore.pi.MCRPIManagerTest.resetManagerInstance(MCRPIManagerTest.java:169)
```

Weirdly i can not reproduce the broken Tests now.

PR #2438, opened 2025-03-11
MCR-3351 Remove Xalan extensions from XEditor

fluetze:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3351).


mycore-xeditor/src/main/java/org/mycore/frontend/xeditor/includes/MCRElementCache.java
```diff
@@ -0,0 +1,77 @@
...
+import org.apache.logging.log4j.Logger;
+import org.jdom2.Element;
+
+/** 
+ * Caches preloaded and imported XML elements by @id during transformation.
+ * The cache size is unlimited, but the cache's lifetime is that of the single transformation. 
+ * 
+ * @author Frank L\U00FCtzenkirchen
+ */
+public class MCRElementCache {
```
rsteph-de: should / can this use /extend MCRCache?

fluetze: I'd rather rename that. It is a sort of cache, but it caches every element with an @id, not the most recently used etc.
Maybe ElementLookupTable is better? Using MCRCache would be just overhead.

fluetze: "There are only two hard things in Computer Science: cache invalidation and naming things."
-- Phil Karlton

PR #2440, opened 2025-03-12
MCR-3349 add support for JUnit 5.x

yagee-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3349).

Adds support for JUnit with `@MyCoReTest` annotation and JUnit extensions `MCRJPAExtension` and `MCRMetadataExtension`. For reference a few JUnit 4 test are already migrated to Junit 5 using annotations instead of inheritance from `MCRTestCase`.

Documentation is handled by https://github.com/MyCoRe-Org/mycore-website/pull/90

### DIFF
--- a/mycore-base/src/test/java/org/mycore/frontend/jersey/resources/MCRJerseyJPATest.java
+++ b/mycore-base/src/test/java/org/mycore/frontend/jersey/resources/MCRJerseyJPATest.java
@@ -18,33 +18,54 @@
 
 package org.mycore.frontend.jersey.resources;
 
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.fail;
+import static org.mycore.test.MCRJPATestHelper.executeUpdate;
+import static org.mycore.test.MCRJPATestHelper.getTableName;
+import static org.mycore.test.MCRJPATestHelper.printTable;
+import static org.mycore.test.MCRJPATestHelper.queryTable;
+import static org.mycore.test.MCRJPATestHelper.quoteSchemaIdentifierIfNeeded;
+
 import java.sql.SQLException;
-import java.util.Map;
+import java.util.Locale;
 import java.util.Set;
 
-import org.junit.After;
-import org.junit.Assert;
-import org.junit.Before;
-import org.junit.Test;
+import org.apache.logging.log4j.LogManager;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
 import org.mycore.access.MCRAccessBaseImpl;
 import org.mycore.access.MCRAccessException;
-import org.mycore.common.MCRStoreTestCase;
+import org.mycore.common.MCRTestConfiguration;
+import org.mycore.common.MCRTestProperty;
 import org.mycore.datamodel.metadata.MCRMetadataManager;
 import org.mycore.datamodel.metadata.MCRObject;
 import org.mycore.datamodel.metadata.MCRObjectID;
 import org.mycore.frontend.jersey.filter.MCRDBTransactionFilter;
 import org.mycore.frontend.jersey.filter.MCRSessionHookFilter;
+import org.mycore.test.MCRJPAExtension;
+import org.mycore.test.MCRMetadataExtension;
+import org.mycore.test.MyCoReTest;
 
 import jakarta.ws.rs.POST;
 import jakarta.ws.rs.Path;
 import jakarta.ws.rs.PathParam;
 import jakarta.ws.rs.core.Response;
 
-public class MCRJerseyJPATest extends MCRStoreTestCase {
+@MyCoReTest
+@ExtendWith(MCRJPAExtension.class)
+@ExtendWith(MCRMetadataExtension.class)
+@MCRTestConfiguration(
+    properties = {
+        @MCRTestProperty(key = "MCR.Access.Class", classNameOf = MCRAccessBaseImpl.class),
+        @MCRTestProperty(key = "MCR.Metadata.Type.object", string = "true")
+    })
+public class MCRJerseyJPATest {
 
     private MCRJerseyTestFeature jersey;
 
-    @Before
+    @BeforeEach
     public void setUpJersey() throws Exception {
         jersey = new MCRJerseyTestFeature();
         jersey.setUp(Set.of(
@@ -54,17 +75,9 @@ public class MCRJerseyJPATest extends MCRStoreTestCase {
             MCRJerseyExceptionMapper.class));
     }
 
-    @After
+    @AfterEach
     public void tearDownJersey() throws Exception {
         jersey.tearDown();
-    }
-
-    @Override
-    protected Map<String, String> getTestProperties() {
-        Map<String, String> testProperties = super.getTestProperties();
-        testProperties.put("MCR.Access.Class", MCRAccessBaseImpl.class.getName());
-        testProperties.put("MCR.Metadata.Type.object", "true");
-        return testProperties;
     }
 
     @Test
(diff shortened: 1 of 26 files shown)


mycore-base/src/test/java/org/mycore/frontend/jersey/resources/MCRJerseyJPATest.java
```diff
@@ -108,7 +119,7 @@ public void create(@PathParam("id") String id) throws MCRAccessException {
         @POST
         public void breakIt(@PathParam("id") String id) {
             String tableName = getTableName("MCRObject");
-            executeUpdate("UPDATE " + tableName + " SET CREATEDATE=null WHERE ID='" + id + "'");
+            executeUpdate("UPDATE " + tableName + " SET \"createdate\"=null WHERE \"id\"='" + id + "'");
```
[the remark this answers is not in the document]
yagee-de: good point: removed sql injection in junit tests

mycore-base/src/test/java/org/mycore/test/MCRMetadataExtension.java
```diff
@@ -0,0 +1,103 @@
...
+import org.junit.jupiter.api.extension.AfterAllCallback;
+import org.junit.jupiter.api.extension.AfterEachCallback;
+import org.junit.jupiter.api.extension.BeforeAllCallback;
+import org.junit.jupiter.api.extension.BeforeEachCallback;
+import org.junit.jupiter.api.extension.Extension;
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.mycore.common.config.MCRConfiguration2;
+import org.mycore.datamodel.common.MCRXMLMetadataManager;
+
+public class MCRMetadataExtension implements Extension, BeforeAllCallback, BeforeEachCallback,
```
toKrause, line 47: This extension could use JavaDoc with a short description. The other extensions have such a descripyion ant it might be rather helpful.

toKrause (approved):

mycore-base/src/test/java/org/mycore/test/tests/MCRTestExtensionTest.java
```diff
@@ -0,0 +1,99 @@
...
+import org.junit.jupiter.api.Test;
+import org.mycore.common.MCRTestConfiguration;
+import org.mycore.common.MCRTestProperty;
+import org.mycore.common.config.MCRConfiguration2;
+import org.mycore.common.config.MCRConfigurationBase;
+import org.mycore.common.config.MCRConfigurationException;
+import org.mycore.test.MyCoReTest;
+
+@MyCoReTest
+public class MCRTestExtensionTest {
```
toKrause, line 32: The following test fails:

```java
package org.mycore.test.tests;

import org.junit.jupiter.api.Assertions;
import org.junit.jupiter.api.Nested;
import org.junit.jupiter.api.Test;
import org.mycore.common.MCRTestConfiguration;
import org.mycore.common.MCRTestProperty;
import org.mycore.common.config.MCRConfiguration2;
import org.mycore.test.MyCoReTest;

@MyCoReTest
public class MCRTestExtensionTest {

    @Nested
    @MCRTestConfiguration(
        properties = {
            @MCRTestProperty(key = "xxx", string = "XXX")
        })
    class MCRTestExtensionTest1 {

        @Test
        public void methodInMCRTestExtensionTest1() {
            Assertions.assertEquals("XXX", MCRConfiguration2.getStringOrThrow("xxx"));
        }

    }

    @Nested
    @MCRTestConfiguration(
        properties = {
            @MCRTestProperty(key = "xxx", string = "not-XXX")
        })
    class MCRTestExtensionTest2 extends MCRTestExtensionTest1 {

        @Test
        public void methodInMCRTestExtensionTest2() {
            Assertions.assertEquals("not-XXX", MCRConfiguration2.getStringOrThrow("xxx"));
        }

    }

}
```

This has nothing to do with nested classes, just with how JUnit 5 handles super classes. I just use nested classes, because that makes  it easier to place everything in one Java file.

When JUnit 5  processes a test class, it doesn't just perform all declared tests, but also all tests in super classes (that are itself test classes).

`MCRTestExtensionTest1` declares  property `xxx` with value `XXX`. When running test method `MCRTestExtensionTest1#methodInMCRTestExtensionTest1`, this checks out.

In this case, `MCRTestExtensionTest2` overwrites  property `xxx` with value `not-XXX`. When running test method `MCRTestExtensionTest2#methodInMCRTestExtensionTest2`, this checks out, but **`MCRTestExtensionTest2#methodInMCRTestExtensionTest1`** fails, because it still expects to see `XXX`, but actually sees the overwritten value `not-XXX`.

toKrause (changes requested):

mycore-base/src/test/java/org/mycore/test/tests/MCRTestExtensionTest.java
yagee-de: This seems not as a bug to me. You can overwrite a test if you want to test something different or not execute a test at all. As you said this is expected behaviour of JUnit 5. Normally you would not rely on a specific value of a property but on a specific feature, e.g. a class implementing an interface and you call this interface methods instead of expecting a specific implementation. So in a subclass you can test for specific implementation tests while also performing general interface tests from the super test class. Or am I missing something here?

toKrause: This is not about overwriting a test method. The example given does not overwrite a test method. The annotation present at the sub class directly influences the super class. This is a highly unusual thing for an annotation to do. I don't think that we should have an overwriting-mechanism for annotations. Annotations should just influence the code they are annotating. This is indeed not a bug, but a confusing design.

Don't just consider the minimal-working-example given above; also consider that the the classes might be in different files and that there may be more than one sub class.

I suggest that a test method should only see the following properties:
- from annotations present at the test method
- from annotations present at the class declaring that test method
- from annotations present at super classes of that class
- from configuration

If you insist, that also annotations present at the sub classes (of the class declaring that test method) should also be visible to a test method, could you at least add a test in `MCRTestExtensionTest` that demonstrates that design choice.

yagee-de: I understand the concern that having a subclass’s annotation affect tests declared in a superclass might seem unusual at first. However, this behavior is actually a natural consequence of how JUnit 5 is designed to work. In JUnit 5, test methods are inherited from superclasses and are executed in the context of the concrete subclass. This means that any class‐level or method-level configuration on the subclass become part of the effective configuration for all tests executed on that instance—even if the test method was originally declared in the superclass. For example a `@BeforeAll` annotated method in a subclass may also change the configuration for the superclass.

In other words, since the test method is executed on an instance of the subclass, it “sees” the annotations applied to that subclass. This is consistent with the object‐oriented principle of polymorphism and is documented in the  [JUnit 5 user guide](https://junit.org/junit5/docs/5.6.1/user-guide/index.html). So if you don't want a specific test method of a superclass not to run in the context of the subclass, you have to hide it with `@Disable`. There would be no use case for test to be executed over and over again in the context of the super class if the context could not be modified by the subclass.

Furthermore, this design enables a very powerful testing pattern: you can define a general set of tests in a superclass and then use subclasses to “customize” the configuration (or even the behaviour) of those tests. This allows you to write one comprehensive suite for an interface or a base class while testing multiple implementations by simply extending that test class. It’s a deliberate mechanism to allow for variation in the test environment across subclasses without duplicating test code.

If you prefer tests that see only the annotations on the declaring class and not those on subclasses, you would need a different design pattern instead of inheritance. However, for our use case—where we want to test different implementations with a common test suite—this inheritance-based configuration is both powerful and appropriate.

If there’s concern about clarity, we could add an explicit test case in MCRTestExtensionTest to document and demonstrate this design choice, as you suggested.

mycore-base/src/test/java/org/mycore/test/MCRMetadataExtension.java
yagee-de: Added Javadoc in https://github.com/MyCoRe-Org/mycore/pull/2440/commits/a33cb180eefb2f6ec8b829a81a8686ba7f8bb2e8 and also removed AfterAllCallback, as it did nothing.

toKrause (approved):

PR #2445, opened 2025-03-14
MCR-3356 Update property name MCR.Persistence.Rule.Store_Class

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3356).


rsteph-de (approved):
OK

PR #2447, opened 2025-03-14
MCR-3358 MCRObjectStaticContentGenerator should not be instantiated via reflection

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3358).


rsteph-de (approved):
OK

PR #2449, opened 2025-03-14
MCR-3360 Remove awkwardly designed methods

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3360).


rsteph-de:
Can you check if it's save to rename this class to MCRMeta**data**HistoryItem.

The current name is confusing to *MCRMetaHistoryDate* which is a MyCoRe datatype and other datatype classes

toKrause:
> Can you check if it's save to rename this class to MCRMeta**data**HistoryItem.
> 
> The current name is confusing to _MCRMetaHistoryDate_ which is a MyCoRe datatype and other datatype classes

`MCRMetaHistoryItem` is a JPA entity class for a table named `MCRMetaHistory`. Could probably be renamed without problem _if the table name stays the same_, but that would only move confusing naming disconnects to another point. So I don't think I want to mess with that.

rsteph-de (approved):

PR #2451, opened 2025-03-17
MCR-3362 Allow string map injection to filter properties by prefix

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3362).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/common/config/MCRConfigurableInstanceHelper.java
+++ b/mycore-base/src/main/java/org/mycore/common/config/MCRConfigurableInstanceHelper.java
@@ -26,6 +26,7 @@ import java.lang.reflect.Modifier;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.EnumSet;
+import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Locale;
@@ -565,8 +566,12 @@ class MCRConfigurableInstanceHelper {
 
             @Override
             public Source<MCRProperty, ?> annotationToSource(MCRProperty annotation) {
-                if (Objects.equals("*", annotation.name())) {
-                    return new AllPropertiesSource(annotation);
+                String annotationName = annotation.name();
+                if (annotationName.equals("*")) {
+                    return new AllPropertiesSource(annotation, "");
+                } else if (annotationName.endsWith(".*")) {
+                    return new AllPropertiesSource(annotation, annotationName
+                        .substring(0, annotationName.length() - 1));
                 } else {
                     return new PropertySource(annotation);
                 }
@@ -744,8 +749,11 @@ class MCRConfigurableInstanceHelper {
 
         protected final MCRProperty annotation;
 
-        private AllPropertiesSource(MCRProperty annotation) {
+        protected final String prefix; 
+        
+        private AllPropertiesSource(MCRProperty annotation, String prefix) {
             this.annotation = annotation;
+            this.prefix = prefix;
         }
 
         @Override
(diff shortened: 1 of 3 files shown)


Mewel (approved):

mycore-base/src/main/java/org/mycore/common/config/MCRConfigurableInstanceHelper.java
```diff
@@ -565,8 +566,12 @@ public Class<MCRProperty> annotationClass() {
 
             @Override
             public Source<MCRProperty, ?> annotationToSource(MCRProperty annotation) {
-                if (Objects.equals("*", annotation.name())) {
-                    return new AllPropertiesSource(annotation);
+                String annotationName = annotation.name();
+                if (annotationName.equals("*")) {
+                    return new AllPropertiesSource(annotation, "");
+                } else if (annotationName.endsWith(".*")) {
```
yagee-de, line 572: We might still have properties like this:
MCR.A=foo
MCR.A.B=bar

So a valid prefix must also end with "\*" only instead of requiring a ".\*"

toKrause: Sorry, but I don't understand what you're trying to communicate. The annotated field here would presumably be one of


1. ```java
    @MCRProperty(name="MCR.A.*", absolute=true)
    public Map<String, String> properties;
    ``` 
    The injected map would contain entry `B` with value `bar`

1. ```java
    @MCRProperty(name="MCR.*", absolute=true)
    public Map<String, String> properties;
    ```
    The injected map would contain entries `A` with value `foo` and `A.B` with value `bar`


Isn't that what you would expect?

---

If you still have concerns, could you please give a full example with properties, annotated field and expected outcome.

yagee-de:
```java
@MCRProperty(name="MCR.A*", absolute=true)
public Map<String, String> properties;
```
Contains all properties prefixed by "MCR.A" so "MCR.A" is included as well as "MCR.A.*"

toKrause: That would then result in a map with:
- key: _empty string_, value: `foo`
- key: `.B`, value: `bar`

And, if we had a property `MCR.AnotherOne=bites the dust`, additionally:
- key: `notherOne`, value: `bites the dust`

Tbh, I don't really see where a feature like this would be sensible and useful (and not better served with multiple annotated fields). Do you have a situation in mind?

Enforcing property-name-segmentation at dots (and signalling that with a required `.*` at the end of the prefix), is, imho, a much better design choice.

yagee-de: I see your point and other configuration systems have strict rules about that. Setting `absolute=true` should always include the prefix. I think it come down to the point: Do we want it either that way or the other. Let's put it on the agenda for the next call.

toKrause: Whether the `key`-attribute of a `@MCRProperty`-annotation is interpreted as an absolute property key (`absolute=true`) or as a relative property key (`absolute=false`, the default) shouldn't have any impact on how the injected map keys prepared. The behavior should always be the same. Keeping the prefix in the injected keys would break established practice and be very confused design choice in relative mode (since tho whole point of annotation based configuration is to abstract away from concrete property keys, especially where nested instances are involved). It would also unnecessarily include redundant information in both modes. The prefix should therefore always be removed.

But yes, lets talk about this in the next web call. It would be very helpful if you could find a real-world example where your proposed change would be useful, b/c I really can't think of any.

yagee-de: > I see your point and other configuration systems have strict rules about that

So we should implement a config checker that WARNS about properties like "MCR.A" when "MCR.A.Foo" is present. So there should be no property x, where name of x is prefix of another property. In future versions of MyCoRe this should be enforced.

toKrause: I don't see that property names being prefixes of other property names are necessarily a problem; or what that has to do with your proposed changed (allowing inter-property-name-segment prefixes for injected property maps).

PR #2453, opened 2025-03-17
MCR-3364 Simplify Solr authentication architecture

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3364).


yagee-de:
This will break API in 2024.06.x and should go into main only, IMHO.

toKrause:
> This will break API in 2024.06.x and should go into main only, IMHO.

My thinking was that the API in question was introduced very recently, so, realistically, there is no one who customized their application here, and the changes are configuration-compatible. Targeting 2024.06.x would create fewer merging conflicts.

But I absolutely see your point here, as I'm usually a proponent of strict compatibility in LTS branches. PR is rebased.

rsteph-de (dismissed):
looks good for me

rsteph-de (approved):
O.K.

PR #2454, opened 2025-03-17
MCR-3365 Simplify and strengthen custom PMD rules for serial version UID

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3365).


rsteph-de (approved):
OK (only Torsten knows what to do here ;-) )

PR #2457, opened 2025-03-17
MCR-3367 Do not require fixed class names to be put in the properties

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3367).

**Relies on [MCR-3352](https://github.com/MyCoRe-Org/mycore/pull/2441).**
**Relies on [MCR-3368](https://github.com/MyCoRe-Org/mycore/pull/2456).**

[MCR-3352]: https://mycore.atlassian.net/browse/MCR-3352?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MCR-3368]: https://mycore.atlassian.net/browse/MCR-3368?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2456, opened 2025-03-17
MCR-3368 Prepare removal of 'legacy' factory methods and prioritize singleton accessor methods

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3368).

**Relies on [MCR-3352](https://github.com/MyCoRe-Org/mycore/pull/2441).**

[MCR-3352]: https://mycore.atlassian.net/browse/MCR-3352?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2441, opened 2025-03-13
MCR-3352 Harmonize configuration based instantiation even more

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3352).

PR #2460, opened 2025-03-17
MCR-3371 Modernize LDAP client into reusable and configurable class

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3371).

**Relies on [MCR-3352](https://github.com/MyCoRe-Org/mycore/pull/2441).**
**Relies on [MCR-3368](https://github.com/MyCoRe-Org/mycore/pull/2456).**
**Relies on [MCR-3367](https://github.com/MyCoRe-Org/mycore/pull/2457).**

[MCR-3352]: https://mycore.atlassian.net/browse/MCR-3352?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MCR-3368]: https://mycore.atlassian.net/browse/MCR-3368?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MCR-3367]: https://mycore.atlassian.net/browse/MCR-3367?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mycore-user2/src/main/java/org/mycore/user2/login/MCRLDAPClient.java
```diff
@@ -258,8 +221,134 @@ private String formatName(String name) {
...
+            Objects.requireNonNull(providerUrl, "Provider URL must not be null");
+            return this;
+        }
+
+        public Map<String, String> toLdapSettings() {
+            validate();
+            Map<String, String> ldapEnv = new HashMap<>();
+            ldapEnv.put(Context.INITIAL_CONTEXT_FACTORY, "com.sun.jndi.ldap.LdapCtxFactory");
+            ldapEnv.put(Context.PROVIDER_URL, providerUrl);
+            ldapEnv.put(Context.SECURITY_AUTHENTICATION, "simple");
```
golsch: Does `SECURITY_AUTHENTICATION` make sense by default if there is no `SECURITY_PRINCIPAL` or `SECURITY_CREDENTIALS`?

mycore-user2/src/main/java/org/mycore/user2/login/MCRLDAPClient.java
```diff
@@ -258,8 +221,134 @@ private String formatName(String name) {
...
     }
 
+    public record ConnectionSettings(
+        String providerUrl,
+        String securityPrincipal,
+        String securityCredentials,
+        Integer readTimeoutMillis) {
+
+        public ConnectionSettings validate() {
+            Objects.requireNonNull(providerUrl, "Provider URL must not be null");
```
golsch, line 321: Perhaps the authenticity should also be conditionally validated.

mycore-user2/src/main/java/org/mycore/user2/login/MCRLDAPClient.java
```diff
@@ -258,8 +221,134 @@ private String formatName(String name) {
...
+                }
+            }
+            ldapEnv.put("com.sun.jndi.ldap.read.timeout",
+                readTimeoutMillis != null ? readTimeoutMillis.toString() : "1000");
+            return ldapEnv;
+        }
+
+    }
+
+    public record Settings(
```
golsch: For me, the name is a bit confusing because there are already the `ConnectionsSettings`.
May move _Base DN_ as "connection entry" to `ConnectionsSettings` and rename `Settings` in `MappingSettings`.

mycore-user2/src/main/java/org/mycore/user2/login/MCRLDAPClient.java
```diff
@@ -91,160 +92,122 @@
  *
  * @author Frank Lützenkirchen
  */
+@SuppressWarnings({"PMD.ReplaceHashtableWithMap"})
+@MCRConfigurationProxy(proxyClass = MCRLDAPClient.Factory.class)
 public final class MCRLDAPClient {
```
golsch, line 98: Is `final` suitable here?

golsch (changes requested):

mycore-user2/src/main/java/org/mycore/user2/login/MCRLDAPClient.java
```diff
@@ -258,8 +221,134 @@ private String formatName(String name) {
...
+
+            }
+
+        }
+
+    }
+
     private static final class LazyInstanceHolder {
-        public static final MCRLDAPClient SINGLETON_INSTANCE = new MCRLDAPClient();
+        public static final MCRLDAPClient SINGLETON_INSTANCE = createInstance();
```
golsch: `SHARED_INSTANCE`?

mycore-user2/src/main/java/org/mycore/user2/login/MCRLDAPClient.java
toKrause: Good point.

In the original implementation, "security principal" and "security credentials" were required.  Connecting to a LDAP server w/o authentication doesn't really make sense or is even possible. I'll change these values back to being required. Then, "security authentication", i.e. the security mechanism to be used, should also be required. But the value should be configurable with `simple` as the default value.

mycore-user2/src/main/java/org/mycore/user2/login/MCRLDAPClient.java
toKrause: Not sure what you mean by "authenticity" here. Could you elaborate?

mycore-user2/src/main/java/org/mycore/user2/login/MCRLDAPClient.java
toKrause: The class has no extension points for a subclass to modify, so I kept if final.

mycore-user2/src/main/java/org/mycore/user2/login/MCRLDAPClient.java
toKrause: The "Base DN" is not used to establish a connection. Maybe `SeachSettings` could be a better name for what is now called `Settings`?

mycore-user2/src/main/java/org/mycore/user2/login/MCRLDAPClient.java
golsch: I meant the part that concerns authentication. But that looks good now after the change.

mycore-user2/src/main/java/org/mycore/user2/login/MCRLDAPClient.java
golsch: Our LDPA can be used without authentication because access is protected via the IP address range. This means that authentication is not necessarily required...

toKrause: Okay. I was unsure if there are reasonable situation where authentication is not required. Do you know, what value you provide for the `SECURITY_AUTHENTICATION` key? Do you not provide one, or do you still set it to `simple` and just don't provide principal and credentials?

golsch: It works without `SECURITY_AUTHENTICATION`. Only `PROVIDER_URL` is sufficient.

mycore-user2/src/main/java/org/mycore/user2/login/MCRLDAPClient.java
```diff
@@ -258,8 +221,164 @@ private String formatName(String name) {
...
+                Objects.requireNonNull(securityCredentials, "Security credentials must not be null");
+            } else if (securityPrincipal != null || securityCredentials != null) {
+                LOGGER.warn("Security principal or credentials given, even though the authentication is 'NONE'");
+            }
+            Objects.requireNonNull(connectTimeoutMillis, "Read timeout must not be null");
+            Objects.requireNonNull(readTimeoutMillis, "Read timeout must not be null");
+            return this;
+        }
+
+        public Map<String, String> toLdapSettings() {
```
golsch, line 327: Only values that are not null should be included in `ldapEnv`. Otherwise line 114 leads to a NPE when creating the `Hashtable` with null values.

golsch (changes requested):

mycore-user2/src/main/java/org/mycore/user2/login/MCRLDAPClient.java
toKrause: Good catch. Will change.

mycore-user2/src/main/java/org/mycore/user2/login/MCRLDAPClient.java
toKrause: I've done some digging. If no `SECURITY_AUTHENTICATION` is set, the default LDAP contxt assumes

- `simple`, if `SECURITY_PRINCIPAL` is set and
- `none` otherwise (as in your case).

Other mechanisms (SASL, for example) are also supported. These may or may not require additional values to be present.

The `ConnectionSetting` is intended to be a nicer API to configure common cases. For specialized cases, `MCRLDAPClient` has a constructor that takes a map of properties. Such a map could be created from sratch or by calling `ConnectionSetting#toLdapSettings` and modifying the result.

For now, I'll change `ConnectionSetting` to explicitly allow for mechanisms `none` and `simple`. That means, that the shared LDAP client, configures by properties, will only be able to be created with those two mechanisms. A completely customizable implementations that retains a nice API is also possible, but not whilst being compatible with the established configuration property names. If such an improvement is required in the future, it should be targeting `main` in a separate PR.

rsteph-de (dismissed):
Looks good for me

rsteph-de (dismissed):
O.K.

golsch (approved):

rsteph-de (approved):

PR #2463, opened 2025-03-18
MCR-3273 Enable PMD rule ProperCloneImplementation and harmonize clone implementations

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3273).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/common/MCRClassTools.java
+++ b/mycore-base/src/main/java/org/mycore/common/MCRClassTools.java
@@ -79,4 +79,30 @@ public class MCRClassTools {
             .orElseGet(MCRClassTools.class::getClassLoader);
     }
 
+    /**
+     * This is a helper function for top most classes in class hierarchies that implement
+     * {@link Cloneable} and expect every subclass to implement a proper {@link Object#clone()}
+     * method, i.e. to never throw a {@link CloneNotSupportedException}.
+     * <p>
+     * It is meant to simplify and harmonize suppression of this exception in the method
+     * signatures of the clone-method of such classes and needs to be called as
+     * <code>MCRClassTools.clone(getClass(), super::clone)</code> exactly.
+     *
+     * @param cloneClass the class to be cloned
+     * @param superClone a method reference to the super classes clone-methode
+     */
+    @SuppressWarnings("unchecked")
+    public static <T> T clone(Class<T> cloneClass, SuperClone<T> superClone) {
+        try {
+            return (T) superClone.getClone();
+        } catch (CloneNotSupportedException e) {
+            throw new MCRException(cloneClass.getName() + " doesn't implement a proper clone-method"
+                + " - this is an implementation mistake", e);
+        }
+    }
+
+    public interface SuperClone<T> {
+        Object getClone() throws CloneNotSupportedException;
+    }
+
 }
--- a/rules.xml
+++ b/rules.xml
@@ -436,4 +436,40 @@
     </properties>
   </rule>
 
+
+  <rule name="MCR.ClassTools.Clone"
+        language="java"
+        message="MCRClassTools#clone should be called as MCRClassTools.clone(getClass(), super::clone) exactly"
+        class="net.sourceforge.pmd.lang.rule.xpath.XPathRule">
+    <description>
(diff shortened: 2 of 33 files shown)


mycore-base/src/main/java/org/mycore/common/MCRClassTools.java
```diff
@@ -79,4 +79,17 @@ public static void updateClassLoader() {
...
 
+    @SuppressWarnings("unchecked")
+    public static <T> T clone(Class<T> cloneClass, SuperClone<T> superClone) {
+        try {
+            return (T) superClone.getClone();
+        } catch (CloneNotSupportedException e) {
+            throw new MCRException(cloneClass.getName() + " is not cloneable", e);
+        }
+    }
+
```
yagee-de: If calling clone() is not supported a CloneNotSupportedException must be thrown not a MCRException. For example if a class implements the clone method but forgets to add "implements Clonable" this will result in a MCRException.

yagee-de: also supplying a SuperClone as a parameter does not enforce that this parameter is always "super::clone"

toKrause: > If calling clone() is not supported a CloneNotSupportedException must be thrown not a MCRException. For example if a class implements the clone method but forgets to add "implements Clonable" this will result in a MCRException.

From the JavaDoc of `CloneNotSupportedException`
> Applications that override the `clone` method can also throw this exception to indicate that an object [whose class implements `Cloneable`] could not or should not be cloned.

Throwing  a `CloneNotSupportedException` is indeed meant to indicate that an object could not be cloned, even thought the class implements `Cloneable`. However, I have the impression that the classes implementing `Cloneable` in MyCoRe also define an implicit contract, that all their sub-classes should have a proper implementation that actually returns a cloned instance (i.e. that they should never use the option to throw a `CloneNotSupportedException`). In case of `MCRMetaInterface`, an interface extending `Cloneable`, such a contract is even made explicit by removing `throws CloneNotSupportedException` from the `clone` methods signature.

This is a helper method to simplify and harmonize implementing classes that define such a contract. It is meant to reduce code duplication.

Examples of former implementations of this contract in `MCRLabel`, `MCRMetaDefault` and `MCRMetaHistoryDate`:

```java
MCRLabel clone = null;
try {
    clone = (MCRLabel) super.clone();
} catch (CloneNotSupportedException ce) {
    // Can not happen
}
return clone;
```

```java
try {
    MCRMetaDefault clone = (MCRMetaDefault) super.clone();
    ...
    return clone;
} catch (CloneNotSupportedException e) {
    //this is impossible!
    return null;
}
```

```java
MCRMetaHistoryDateText clone = null;
try {
    clone = (MCRMetaHistoryDateText) super.clone();
    clone.text = this.text;
    clone.lang = this.lang;
} catch (CloneNotSupportedException e) {
    // this can not happen, as MCRMetaInterface implements Cloneable
}
return clone;
```


The remaining classes modified in this PR mostly had a broken implementation of `clone` that didn't take possible sub-classes into account; the reason for the PMD rule in question.

toKrause: > also supplying a SuperClone as a parameter does not enforce that this parameter is always "super::clone"

It is indeed unusual that a method requires specific constructs as their parameters. Java does not have any language constructs (like templates or macros) that can be used to make such code-snippets reusable. I'll add a custom PMD rule, that will enforce the intended use.

kkrebs (changes requested):

rules.xml
```diff
@@ -436,4 +436,40 @@
...
   </rule>
 
+
+  <rule name="MCR.ClassTools.Clone"
+        language="java"
+        message="MCRClassTools#clone should be called as MCRClassTools.clone(getClass(), super::clone) exactly"
+        class="net.sourceforge.pmd.lang.rule.xpath.XPathRule">
+    <description>
+      MCRClassTools#clone is a helper function for top most classes in class hierarchies that implement Cloneable
+      and expect every subclass to implement a proper clone method, i.e. to never throe a CloneNotSupportedException.
```
yagee-de: throe --> throw

yagee-de (changes requested):

PR #2467, opened 2025-03-21
MCR-3375 Reduce Unnecessary Object Versions

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3375).

requires https://github.com/MyCoRe-Org/datamodel-plugin/pull/9


sebhofmann:
Still need to:
1. ~discuss if MCRObjectNormalizer is a good name
   (possible alternatives are: shrink, reduce, collapse, minimize, codense)~
2. ~discover all places were the expanded metadata is needed~
3. ~discover all eventhandlers which need to handle the new events~
4. ~resolve todos in code~
5. ~write migration command to rename `children`to `childrenOrder`~

mycore-mods/src/main/java/org/mycore/mods/MCRMODSLinkProvider.java
```diff
@@ -0,0 +1,68 @@
...
+        Collection<MCRCategoryID> categoriesOfObject = new HashSet<>(super.getCategoriesOfObject(obj));
+
+        MCRMODSWrapper modsWrapper = new MCRMODSWrapper(obj);
+        categoriesOfObject.addAll(modsWrapper.getMcrCategoryIDs());
+
+        return categoriesOfObject;
+    }
+
+    private static void checkObjectType(MCRObject obj) throws OperationNotSupportedException {
+        if (!obj.getId().getTypeId().equals("mods")) {
```
rsteph-de: This cannot be hardcoded - I have other MCRObject types which contains mods
Please use `MCRMODSWrapper.isSupported(MCRObject obj)` or `MCRMODSWrapper.isSupported(MCRObjectID objId)` instead.

mycore-base/src/main/resources/config/mycore.properties
```diff
@@ -102,6 +102,31 @@ SAXON=net.sf.saxon.TransformerFactoryImpl
 # Developer commands
 # MCR.CLI.Classes.Internal=%MCR.CLI.Classes.Internal%,org.mycore.frontend.cli.MCRDeveloperCommands
 
+##############################################################################
+# ObjectExpander configuration
+##############################################################################
+MCR.ObjectExpander.Impl.Default=org.mycore.common.MCRBasicObjectExpander
```
toKrause: Please use `.Class` suffix for properties containing instantiable class references, since
- support for `.class` and the empty suffix is planned to be removed and
- (given that  `.Class` is the standard, which will be the case in the future) the presence of a `.Class` suffix communicates that a property is a configurable instance and additional configuration properties might be available

Example:
`MCR.ObjectExpander.Impl.Default.Class=org.mycore.common.MCRBasicObjectExpander`

toKrause (dismissed):

mycore-base/src/main/resources/config/mycore.properties
sebhofmann: It seems like the API is broken, because i also need to change my call to `MCR.ObjectExpander.Impl.Default.Class`

sebhofmann: It seems like its only broken with `getSIngleInstance`

toKrause: Automatic suffix detection indeed only works works methods that (indirectly) use `getInstantiatablePropertyKeys`, which evaluates a property name prefix. Other methods need the exact property name. 

This has historic reasons and is one of the reasons why the naming rules should be harmonized.

mycore-base/src/main/java/org/mycore/common/MCRBasicObjectExpander.java
```diff
@@ -51,7 +51,8 @@ public MCRExpandedObject expand(MCRObject mcrObject) {
             newStructure.setParent(mcrObject.getParent());
         }
 
-        for (MCRObjectID mcrObjectID : MCRChildOrderStrategyManager.getChildOrderStrategy().getChildOrder(mcrObject)) {
+        for (MCRObjectID mcrObjectID : MCRChildrenOrderStrategyManager.getChildOrderStrategy()
```
rsteph-de: shouldn't the method be named `getChildrenOrderStrategy`

rsteph-de (dismissed):
looks good for me now

all open discussion has been cleared.

mycore-base/src/main/java/org/mycore/common/MCRBasicObjectExpander.java
rsteph-de: done

rsteph-de (approved):
O.K. 
changed back to return void vor MCRMetadataManager.create /.update

PR #2469, opened 2025-03-24
MCR-3378 Fix MCRPIManagerTest

toKrause:


Mewel:
This does not fix the bug:
```
[ERROR] Tests run: 4, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 1.300 s <<< FAILURE! -- in org.mycore.pi.MCRPIManagerTest
[ERROR] org.mycore.pi.MCRPIManagerTest.testRegistrationService -- Time elapsed: 0.379 s <<< FAILURE!
java.lang.AssertionError: There should be one resolver
```
Looks like it depends on the test execution order.

On Ubuntu 24.04 on console and intellij.

The problem is that the MCRMockResolver is not in the list because it was initialized before.

PR #2470, opened 2025-03-25
MCR-3348 fix MCRPIManagerTest

yagee-de:

### DIFF
--- a/mycore-pi/src/main/java/org/mycore/pi/MCRPIManager.java
+++ b/mycore-pi/src/main/java/org/mycore/pi/MCRPIManager.java
@@ -74,7 +74,13 @@ public final class MCRPIManager {
         applyConfiguration();
     }
 
-    private void applyConfiguration() {
+    /**
+     * Reads the configuration and applies it to the manager.
+     * <p>
+     *     No sychronization is done, so this method is not thread-safe.
+     *     It is intended to be called from the constructor and test classes only.
+     */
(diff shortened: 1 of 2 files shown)


toKrause:
Can you elaborate why the proposed fix in https://github.com/MyCoRe-Org/mycore/pull/2469 is not sufficient? Running the tests in that PR give no indications that resetting the manager is actually necessary (see attached log). 

The fix, as proposed here, still imposes requirements on the internal structure of the class under test. Unit tests shouldn't do that. Would it be acceptable to make the new method `applyConfiguration` public and rename it to `reinitialize`? Akin to the re-initialization method of `MCRURIResolver`. Better yet, un-singletonize `MCRPIManager`, so it becomes easier to test, and only provide a shared common instance, like it was done with `MCRURIResolver` in MCR-3348.


[test.log](https://github.com/user-attachments/files/19445560/test.log)

Mewel:
This fixes the bug for me, but I don't like that we use reflection for the test case.

toKrause:
Best thing to do would be to move from a singleton instance (accessor method `getInstance`) to a shared instance (accessor method `obtainInstance`) that is instantiated with a configuration obtained from configuration properties and use an appropriately configured instance for testing. Suggestion: https://github.com/MyCoRe-Org/mycore/pull/2473

sebhofmann (approved):

toKrause (approved):

PR #2473, opened 2025-03-26
MCR-3348 Update MCRPIManager and fix MCRPIManagerTest issue

toKrause:

### DIFF
--- a/mycore-mods/src/main/java/org/mycore/mods/identifier/MCRAbstractMODSMetadataService.java
+++ b/mycore-mods/src/main/java/org/mycore/mods/identifier/MCRAbstractMODSMetadataService.java
@@ -96,7 +96,7 @@ public class MCRAbstractMODSMetadataService
 
         String text = element.getTextNormalize();
 
-        return MCRPIManager.getInstance().getParserForType(getIdentifierType())
+        return MCRPIManager.obtainInstance().getParserForType(getIdentifierType())
             .parse(text)
             .map(MCRPersistentIdentifier.class::cast);
     }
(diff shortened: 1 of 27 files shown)


toKrause:
Best thing to fix the currently broken `MCRPIManagerTest` would be to move `MCRPIManager` from a singleton instance (accessor method `getInstance`) to a share instance (accessor method `obtainInstance`) that is instantiated with a configuration obtained from configuration properties and use an appropriately configured instance for testing. This removes the requirement to perform reflection based resetting for testing (which imposes requirements on the internal structure of the clas under test, a big no-no anyways; and, apparently, breaks unnecessarily easy).

Also, I don't see a reason why parsers must be instantiated every time they are used. All existing implementations are stateless and there is no reason why future implementations might be stateful.

There are also two issues with the existing tests:

1. `String mockString = MCRMockIdentifier.MOCK_SCHEME + "http://google.de/";` results in `MOCK:http://google.de/`, which is currently parsed by `MCRHandleParser` as a handle with prefix `MOCK:http:` and suffix `/google.de/`; `MCRMockIdentifierParser`  is never used. Changing to `MCRMockIdentifier.MOCK_SCHEME + "foobar";` fixes that, because no `/` delimiter exists any more.
2. Test method `testGet()` does two completely unrelated thing, the first of which is virtually identical to what is tested in `testParseIdentifier()`.

sebhofmann:
I don´t think its a good idea to expose the MCRPIManager Constructor to the public. It does not make sense to have more than one instance in an application and its unclear now if one should call the constructor or obtain the shared instance.

sebhofmann:
Also this is the 2024.06 branch which should be stable since 10 Months, the changes should be small to archieve the bugfix.

toKrause:
We could reduce the constructor visibility to "package private". That would still allow custom instances for testing, but alleviate the problem you mentioned. Would still be better then reflection voodoo in the tests.

PR #2474, opened 2025-03-26
MCR-3362 Remove accidentally added support for required attribute in case of map injection

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3362).


toKrause:
MCR-3362 didn't just add support for the new prefix pattern. By accident, proper support for the `required` attribute of `@MCRProperty` was also "cherry-picked". This, however, is not backwards compatible and should therefore be removed again.

Possommi (approved):

PR #2478, opened 2025-03-28
MCR-3245 Enable PMD rule LinguisticNaming

rsteph-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3245).

### DIFF
--- a/mycore-iiif/src/main/java/org/mycore/iiif/image/parser/MCRIIIFScaleParser.java
+++ b/mycore-iiif/src/main/java/org/mycore/iiif/image/parser/MCRIIIFScaleParser.java
@@ -59,7 +59,7 @@ public class MCRIIIFScaleParser {
 
         StringBuilder wBuilder = new StringBuilder();
         StringBuilder hBuilder = new StringBuilder();
-        getWidthAndHeightStrings(wBuilder, hBuilder);
+        retrieveWidthAndHeightStrings(wBuilder, hBuilder);
         String w = wBuilder.toString();
         String h = hBuilder.toString();
 
@@ -135,7 +135,7 @@ public class MCRIIIFScaleParser {
             (int) Math.ceil(percentValue / 100 * sourceRegionHeight));
     }
 
-    private void getWidthAndHeightStrings(StringBuilder wBuilder, StringBuilder hBuilder) {
+    private void retrieveWidthAndHeightStrings(StringBuilder wBuilder, StringBuilder hBuilder) {
         char[] chars = this.targetScale.toCharArray();
         boolean writeW = true;
         boolean first = true;
--- a/mycore-iiif/src/main/java/org/mycore/iiif/image/parser/MCRIIIFScaleParser.java
+++ b/mycore-iiif/src/main/java/org/mycore/iiif/image/parser/MCRIIIFScaleParser.java
@@ -57,24 +57,16 @@ public class MCRIIIFScaleParser {
             return parsePercentValue();
         }
 
-        StringBuilder wBuilder = new StringBuilder();
-        StringBuilder hBuilder = new StringBuilder();
-        retrieveWidthAndHeightStrings(wBuilder, hBuilder);
-        String w = wBuilder.toString();
-        String h = hBuilder.toString();
+        Dimension d = Dimension.build(this.targetScale);
 
-        try {
-            if (w.length() == 0) {
-                return scaleToHeight(Integer.parseInt(h));
-            } else if (h.length() == 0) {
-                return scaleToWidth(Integer.parseInt(w));
-            } else if (isBestFit()) {
-                return bestFit(Integer.parseInt(w), Integer.parseInt(h));
-            } else {
-                return scale(Integer.parseInt(w), Integer.parseInt(h));
-            }
-        } catch (NumberFormatException e) {
-            throw new IllegalArgumentException(h + " or " + w + " ist not a valid scale value!", e);
+        if (d.isWidthEmpty()) {
+            return scaleToHeight(d.height());
+        } else if (d.isHeightEmpty()) {
+            return scaleToWidth(d.width());
+        } else if (isBestFit()) {
+            return bestFit(d.width(), d.height());
+        } else {
+            return scale(d.width(), d.height());
         }
 
     }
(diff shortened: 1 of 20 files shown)


rsteph-de:
We need to talk about the changes in `MCRObjectCommands` ...
There have been two pairs of methods with similiar results:
- `showNextId()` vs  `getNextId()`
- `showLastId()` vs. `getNextId()`

The `get*()`have to be deleted to comply with this PMD rule.
We can discuss if we want to have the shorter (only the ID) or longer output (a sentence like *the last id of {base} is {mcrid}*) as output of the command.

mycore-iiif/src/main/java/org/mycore/iiif/image/parser/MCRIIIFScaleParser.java
```diff
@@ -59,7 +59,7 @@ public MCRIIIFImageTargetSize parseTargetScale() {
 
         StringBuilder wBuilder = new StringBuilder();
         StringBuilder hBuilder = new StringBuilder();
-        getWidthAndHeightStrings(wBuilder, hBuilder);
+        retrieveWidthAndHeightStrings(wBuilder, hBuilder);
```
Mewel: This doesn't really look better, retrieve is kind of like get. I would rewrite to:

```
Dimension dimension = getDimension(this.targetScale);

        try {
            if (dimension.isWidthEmpty()) {
                return scaleToHeight(dimension.getHeight());
            } else if (dimension.isHeightEmpty()) {
                return scaleToWidth(dimension.getWidth());
            } else if (isBestFit()) {
                return bestFit(dimension.getWidth(), dimension.getHeight());
            } else {
                return scale(dimension.getWidth(), dimension.getHeight());
            }
        } catch (NumberFormatException e) {
            throw new IllegalArgumentException(
                dimension.height + " or " + dimension.width + " ist not a valid scale value!", e);
        }
```
```

private Dimension getDimension(String scale) {
        StringBuilder wBuilder = new StringBuilder();
        StringBuilder hBuilder = new StringBuilder();

        char[] chars = scale.toCharArray();
        boolean writeW = true;
        boolean first = true;
        for (char currentChar : chars) {
            switch (currentChar) {
                case ',' -> {
                    first = false;
                    if (!writeW) {
                        throw new IllegalArgumentException("second , found in scale!");
                    }
                    writeW = false;
                }
                case '!' -> {
                    if (!first) {
                        throw new IllegalArgumentException("! should be located only in the first position of scale!");
                    }
                    first = false;
                }
                default -> {
                    first = false;
                    if (writeW) {
                        wBuilder.append(currentChar);
                    } else {
                        hBuilder.append(currentChar);
                    }
                }
            }
        }
        return new Dimension(wBuilder.toString(), hBuilder.toString());
    }

    private record Dimension(String width, String height) {

        boolean isWidthEmpty() {
            return width == null || width.isEmpty();
        }

        boolean isHeightEmpty() {
            return height == null || height.isEmpty();
        }

        int getWidth() {
            return Integer.parseInt(width);
        }

        int getHeight() {
            return Integer.parseInt(height);
        }

    }

```

mycore-mods/src/main/java/org/mycore/mods/rss/MCRRSSFeedImporter.java
```diff
@@ -135,10 +135,10 @@ public MCRRSSFeedImporter(String sourceSystemID) {
         field2queryID = MCRConfiguration2.getStringOrThrow(prefix + "Field2QueryID");
         xsl2BuildNotificationMail = MCRConfiguration2.getString(prefix + "XSL2BuildNotificationMail").orElse(null);
 
-        getPattern2FindID(prefix);
+        retrievePattern2FindID(prefix);
```
Mewel: Why not change it to:

```
        pattern2findID = getPattern2FindID(prefix);
    }

    private Pattern getPattern2FindID(String prefix) {
        String patternProperty = prefix + "Pattern2FindID";
        try {
            String pattern = MCRConfiguration2.getStringOrThrow(patternProperty);
            return Pattern.compile(pattern);
        } catch (PatternSyntaxException ex) {
            String msg = "Regular expression syntax error: " + patternProperty;
            throw new MCRConfigurationException(msg, ex);
        }
    }
```

mycore-wcms2/src/main/java/org/mycore/wcms2/resources/MCRWCMSFileBrowserResource.java
```diff
@@ -124,7 +124,7 @@ public String getFolders() throws IOException, ParserConfigurationException, SAX
         DocumentBuilder docBuilder = docFactory.newDocumentBuilder();
         File file = new File(MCRLayoutUtilities.getNavigationURL().getPath());
         Document doc = docBuilder.parse(file);
-        getAllowedPaths(doc.getDocumentElement());
+        retrieveAllowedPaths(doc.getDocumentElement());
```
Mewel: Better name would be: updateFolderList()

Mewel (changes requested):

Mewel:
> We need to talk about the changes in `MCRObjectCommands` ... There have been two pairs of methods with similiar results:
> 
> * `showNextId()` vs  `getNextId()`
> * `showLastId()` vs. `getNextId()`
> 
> The `get*()`have to be deleted to comply with this PMD rule. We can discuss if we want to have the shorter (only the ID) or longer output (a sentence like _the last id of {base} is {mcrid}_) as output of the command.

I would remove getNextId() and only use showNextId(), but I would remove the String "The next free ID  is {}" and only return the actual ID like in the old getNextId() method.

Robert: This is how it is currently done in this PR.

mycore-mods/src/main/java/org/mycore/mods/rss/MCRRSSFeedImporter.java
rsteph-de: that's better - I will modify the PR.

mycore-iiif/src/main/java/org/mycore/iiif/image/parser/MCRIIIFScaleParser.java
rsteph-de: I liked this approach and developed it further.
But in general I would like to keep changes as small as possible and do changes like this in a follow-up issue/PR.
My priority: Get all PMD rules working ASAP.

mycore-wcms2/src/main/java/org/mycore/wcms2/resources/MCRWCMSFileBrowserResource.java
rsteph-de: done!

Mewel (dismissed):

kkrebs (changes requested):

Mewel (approved):

PR #2482, opened 2025-03-30
MCR-3272 Enable PMD rule OverrideBothEqualsAndHashcode

rsteph-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3272).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/common/xsl/MCRParameterCollector.java
+++ b/mycore-base/src/main/java/org/mycore/common/xsl/MCRParameterCollector.java
@@ -386,6 +386,26 @@ public class MCRParameterCollector {
         return hashCode;
     }
 
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
+        if (obj == null) {
+            return false;
+        }
+        if (getClass() != obj.getClass()) {
+            return false;
+        }
+        MCRParameterCollector other = (MCRParameterCollector) obj;
+        return hashCode == other.hashCode
+            && modified == other.modified
+            && onlySetXSLParameters == other.onlySetXSLParameters
+            && setPropertiesFromConfiguration == other.setPropertiesFromConfiguration
+            && Objects.equals(parameters, other.parameters);
+
+    }
+
     private static final class SavePropertiesCacheHolder {
         private final static AtomicInteger COMPUTED_HASH_CODE = new AtomicInteger(0);
         private final static AtomicReference<UUID> PROPERTIES_CHANGE_LISTENER_ID = new AtomicReference<>();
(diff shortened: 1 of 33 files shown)


rsteph-de:
In most cases I created missing hashCode() and equals() methods via Eclipse automatically

mycore-base/src/main/java/org/mycore/common/xsl/MCRParameterCollector.java
```diff
@@ -386,6 +386,26 @@ public int hashCode() {
...
+        }
+        if (getClass() != obj.getClass()) {
+            return false;
+        }
+        MCRParameterCollector other = (MCRParameterCollector) obj;
+        return hashCode == other.hashCode
+            && modified == other.modified
+            && onlySetXSLParameters == other.onlySetXSLParameters
+            && setPropertiesFromConfiguration == other.setPropertiesFromConfiguration
+            && Objects.equals(parameters, other.parameters);
```
toKrause, line 404: All bets are probably off if `modified` is `true` for one of the compared `MCRParameterCollector`-instances, because that means, that the cached `hashCode` is stale.

Maybe
```
return hashCode() == other.hashCode()
            && onlySetXSLParameters == other.onlySetXSLParameters
            && setPropertiesFromConfiguration == other.setPropertiesFromConfiguration
            && Objects.equals(parameters, other.parameters);
```
That would ensure that `modified` is `false` for both instances and that the cached `hashCode` is up to date for both instances.


Also, the `hashCode()`-method contains this gem:

```
int result = LOGGER.hashCode();
```

I have no clue what the purpose of that is. Should you be able to figure it out, maybe add a comment. Bu I suspect, that the classic prime of `31` should also be fine.

mycore-base/src/main/java/org/mycore/datamodel/metadata/MCRMetaDefault.java
```diff
@@ -436,18 +436,32 @@ public int hashCode() {
...
+     *  
+     * 
+     */
     @Override
     public boolean equals(Object obj) {
         if (this == obj) {
             return true;
         }
-        if (obj == null || getClass() != obj.getClass()) {
+        if(obj == null) {
```
toKrause: Small formatting error (missing space).

toKrause (changes requested):

mycore-base/src/main/java/org/mycore/common/xsl/MCRParameterCollector.java
rsteph-de: I checked   `LOGGER.hashCode()`, which does `return getName().hashCode();`
Therefore I agree we can work with prime of `31`.

rsteph-de: The hashCode() function is used to generate an eTag to cache the output of an XSLT template which was called with this MCRParameterCollector.
There is proably no use case for the new `equals()` function, But we should make sure that it returns a reasonable and valid result.
I will follow your suggestion.

toKrause (dismissed):

kkrebs (changes requested):

PR #2500, opened 2025-04-08
MCR-3411 Allow to request display name of class in a certain language in MCRXMLFunctions.getDisplayName

Possommi:
[MCR-3411](https://mycore.atlassian.net/browse/MCR-3411).

[MCR-3411]: https://mycore.atlassian.net/browse/MCR-3411?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mycore-base/src/main/java/org/mycore/common/xml/MCRXMLFunctions.java
+++ b/mycore-base/src/main/java/org/mycore/common/xml/MCRXMLFunctions.java
@@ -690,7 +690,8 @@ public class MCRXMLFunctions {
      * @param categoryId the category id
      * @param lang the desired language for which to get the display name
      *
-     * @return the display name or an empty string when there is no label in the desired language
+     * @return the display name in the desired language, if the label for the desired language cannot be found the
+     *         method trys to deliver the default label. If this fails an empty string is returned.
      */
     public static String getDisplayName(String classificationId, String categoryId, String lang) {
         try {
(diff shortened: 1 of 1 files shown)


mycore-base/src/main/java/org/mycore/common/xml/MCRXMLFunctions.java
```diff
@@ -683,6 +683,29 @@ public static String getDisplayName(String classificationId, String categoryId)
...
+     *
+     * @return the display name or an empty string when there is no label in the desired language
+     */
+    public static String getDisplayName(String classificationId, String categoryId, String lang) {
+        try {
+            MCRCategoryID categID = new MCRCategoryID(classificationId, categoryId);
+            MCRCategoryDAO dao = MCRCategoryDAOFactory.getInstance();
+            MCRCategory category = dao.getCategory(categID, 0);
+            Optional<MCRLabel> label = category.getLabel(lang);
+            return label.isEmpty() ? "" : label.get().getText();
```
toKrause: I don't know the use case for this method no there is no further information given in the ticket description.

Wouldn't it be better to fall back to `category.getCurrentLabel()` if there is no label in the desired language, rather then an empty string?

Possommi: I just want an easy method for getting a label in a certain language with basically the same behaviour as `org.mycore.common.xml.MCRXMLFunctions#getDisplayName(java.lang.String, java.lang.String)`. That is the reason why an empty string is returned if a label cannot be resolved.

toKrause: That is why I'm asking what you are actually trying to achieve. Do you want

- return label in requested language or nothing (your current implementation) or
- return a label, preferably in the requested language (which would be similar to the existing function)?


`MCRXMLFunctions#getDisplayName(java.lang.String, java.lang.String)` uses `MCRCategory#getCurrentLabel()`, the implementation of which tries to find a label in multiple ways  (current language, default language, any configured application language, any actual language, any language) before it finally gives up. Your implementations doesn't do that.

If you want a function with basically the same behavior as the existing function, you should either fall back to the existing function or fall back to `MCRCategory#getCurrentLabel()` before returning an empty string. You should not immediately fall back to returning an empty string.

toKrause (approved):

PR #2512, opened 2025-04-17
MCR-3420 fix scopedValue can be null

Mewel:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3420).

* add null check and return value of default action


rsteph-de (approved):
OK.

PR #2517, opened 2025-04-22
MCR-3423 Improve resource path exceptions and testing

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3423).


mycore-base/src/main/java/org/mycore/resource/MCRResourcePath.java
```diff
@@ -99,25 +101,48 @@ public int hashCode() {
     }
 
     /**
-     * Creates a {@link MCRResourcePath} that represents the given path as a resource path
+     * Creates a {@link MCRResourcePath} that represents the given path as a resource path.
+     * <p>
+     * Returns en empty {@link Optional} in case of an empty path or in case
```
kkrebs: en -> an

mycore-base/src/main/java/org/mycore/resource/MCRResourcePath.java
```diff
@@ -99,25 +101,48 @@ public int hashCode() {
...
     public static Optional<MCRResourcePath> ofPath(Optional<String> path) {
-        return path.flatMap(MCRResourcePath::ofPath);
+        return Objects.requireNonNull(path, "Path must not be null").flatMap(MCRResourcePath::ofPath);
     }
 
     /**
-     * Creates a {@link MCRResourcePath} that represents the given path as a resource path
+     * Creates a {@link MCRResourcePath} that represents the given path as a resource path.
+     * <p>
+     * Returns en empty {@link Optional} in case of a missing or empty path or in case
```
kkrebs: en -> an

mycore-base/src/main/java/org/mycore/resource/MCRResourcePath.java
```diff
@@ -126,36 +151,85 @@ private static MCRResourcePath createResourcePath(String path) {
...
-        } else if (!path.startsWith("/")) {
-            return Optional.of(safePath("/" + path));
-        } else {
-            return Optional.of(safePath(path));
+    private static String cleanAndSafePath(String path, boolean orThrow) {
+        if (path == null || path.isEmpty()) {
+            if (orThrow) {
+                throw IllegalPathException.Code.MISSING_OR_EMPTY_PATH.createException(path);
+            }
+            return null;
```
rsteph-de: Mabe an explit else {} would make it clear, that the result is based on the `orThrow` parameter? - Or is this a PMD validation? - check other occurences as well

mycore-base/src/main/java/org/mycore/resource/MCRResourcePath.java
```diff
@@ -126,36 +151,85 @@ private static MCRResourcePath createResourcePath(String path) {
...
+        if (path.endsWith("/")) {
+            if (orThrow) {
+                throw IllegalPathException.Code.DIRECTORY_PATH.createException(path);
+            }
+            return null;
         }
+        if (!path.startsWith("/")) {
+            return safePath("/" + path);
+        }
+        return safePath(path);
```
rsteph-de, line 223: the called method allways throws an exception and does not return null` like the code before

rsteph-de (changes requested):

mycore-base/src/main/java/org/mycore/resource/MCRResourcePath.java
toKrause: There are two classes of errors.

- Invalid paths (`null`, empty or targeting a directory) and
- (if user supplied values are used to create them) potentially evil paths (containing empty (`//`), self reference (`/./`), or-most problematic-parent reference (`/../`) segments, or paths targeting forbidden resources).

The existing methods return an empty `Optional` (as a convenience to remove the need for for example null-checks before every invocation) for invalid paths and throw an error for potentially evil paths. The new methods always throw an error.

That's why the distinction between throwing an exception or returning `null` in `cleanAndSafePath` is only done in case of the invalid path checks.

I've extended the Javadoc to the public methods in this class to make this clear to the caller.

mycore-base/src/main/java/org/mycore/resource/MCRResourcePath.java
```diff
@@ -210,4 +286,53 @@ public Optional<String> asAbsoluteWebPath() {
 
     }
 
+    /**
+     * A {@link IllegalPathException} indicates that an illegal path is used as a resource or web resource path.
+     * <p>
+     * The {@link #code} specifies the nature of the violation.
+     */
+    public static class IllegalPathException extends IllegalArgumentException {
```
rsteph-de, line 294: Should'nt this be named MCRIllegalPathException

toKrause: Inner classes in MyCoRe do not necessarily have the MCR-prefix, for example: `MCRMailer.Email`, `MCRMailer.MessagePart`, `MCRTextResolver.Text`, `MCRXsonifyTransformer.SettingsBuilder`, `MCRCache.ModifiedHandle`.

This class would be references as `MCRResourcePath.IllegalPathException`, no need to repeat the `MCR`

rsteph-de: I was thinking of output with `LOGGER.error(e)`, where you would like to see that this is a MyCoRe releated exception.

rsteph-de (approved):
O.K.

PR #2519, opened 2025-04-22
MCR-3424 Use set instead of list as internal data structure

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3424).


erodde:
looks good to me

erodde (approved):

PR #2520, opened 2025-04-22
MCR-3425 Reinitialize Solr authenticators when corresponding properties change

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3425).


golsch (approved):
Looks good and MIR Wizard works.

PR #2527, opened 2025-04-28
MCR-3432 Updating classifications with the restapi can lead to broken classifications

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3432).


sebhofmann:
I also discovered a bad side effect https://mycore.atlassian.net/browse/MCR-3431 , which probably requires a bigger cleanup.

rsteph-de (approved):
OK

PR #2528, opened 2025-04-30
MCR-3433 Add tests for resource providers locators filters and selectors

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3433).


toKrause:
@yagee-de since this PR also makes som modifications to non-test-code, I suggest merging int main and not making this PR part of the upcoming LTS release. I therefore removed the "important" flag

kkrebs (approved):

PR #2537, opened 2025-05-14
MCR-3442 use write without result in neo4j

michael-becker:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3442).

We return a result object in a write operation, which is not allowed in neo4j (anymore).

I did as AI said and it worked: https://chatgpt.com/share/68246ad4-536c-8001-970f-3618997225bb


yagee-de (approved):

rsteph-de (approved):
OK

PR #2538, opened 2025-05-21
MCR-3445 increase request buffer size in solrj 

rsteph-de:

### DIFF
--- a/mycore-solr/src/main/java/org/mycore/solr/MCRSolrCore.java
+++ b/mycore-solr/src/main/java/org/mycore/solr/MCRSolrCore.java
@@ -192,7 +192,7 @@ public class MCRSolrCore {
 
                 Method mSetRequestBufferSize = jettyHttpClient.getClass()
                     .getDeclaredMethod("setRequestBufferSize", int.class);
-                mSetRequestBufferSize.invoke(jettyHttpClient);
+                mSetRequestBufferSize.invoke(jettyHttpClient, requestBufferSize);
             } catch (NoSuchMethodException | InvocationTargetException | IllegalAccessException e) {
                 LOGGER.error("Error increasing RequestBufferSize for Http2SolrClient", e);
             }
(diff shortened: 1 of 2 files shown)


mycore-solr/src/main/java/org/mycore/solr/MCRSolrCore.java
```diff
@@ -167,12 +169,32 @@ private static HttpSolrClientBase getSolrClientInstance(String baseSolrUrl, int
...
+            try {
+                Method mGetHttpClient = solrclient.getClass().getDeclaredMethod("getHttpClient");
+                mGetHttpClient.setAccessible(true);
+
+                // https://github.com/jetty/jetty.project/blob/jetty-11.0.x/jetty-client/src/main/java/org/eclipse/jetty/client/HttpClient.java
+                Object jettyHttpClient = mGetHttpClient.invoke(solrclient);
+                
+                Method mSetRequestBufferSize = jettyHttpClient.getClass()
+                    .getDeclaredMethod("setRequestBufferSize", int.class);
+                mSetRequestBufferSize.invoke(jettyHttpClient, 511 * 1024);
```
toKrause: Why this specific hard-coded value? Could be a configuration property for applications that need an even larger buffer size.

toKrause (changes requested):

mycore-solr/src/main/java/org/mycore/solr/MCRSolrCore.java
rsteph-de: Good idea. we can even use the exististence of the property to decide if the modifications are necessary.
(see commit)

toKrause (approved):

rsteph-de:
I am setting this back to draft. 
I added another PR which handles alto word coordinates in a separate Solr atomic update request

rsteph-de:
obsolete, word coordinates are handled via separate Solr atomic update request

PR #2566, opened 2025-07-10
Exception while build IIIF manifest if no title is set

bt1cu:


yagee-de:
Should probably merged the other way around

PR #2570, opened 2025-07-12
MCR-3445 send alto params as solr atomic update

rsteph-de:


rsteph-de:
I used the SolrIndexHandlerFactory to work with a new MCRSolrAltoFileIndexHandler,
which sends alto word coordinates as Solr atomic update.

see https://github.com/MyCoRe-Org/JSPDocportal/blob/main/jspdocportal-ir-module/src/main/java/org/mycore/jspdocportal/ir/solr/index/MCRSolrAltoFileIndexHandler.java

PR #2574, opened 2025-07-14
MCR-3472 remove MCREventhander properties after MCREventManagerTest run

rsteph-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3472).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/common/events/MCREventManager.java
+++ b/mycore-base/src/main/java/org/mycore/common/events/MCREventManager.java
@@ -44,6 +44,8 @@ import org.mycore.common.config.MCRConfigurationException;
  */
 public final class MCREventManager {
 
+    private static volatile MCREventManager instance;
+
     public static final String CONFIG_PREFIX = "MCR.EventHandler.";
 
     /** Call event handlers in forward direction (create, update) */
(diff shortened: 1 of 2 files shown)


Mewel:
This is not a fix in my eyes. The problem is the LazyInstanceHolder.

The test should look like this:

```
    @Test
    public void instance() {
        try {
            MCREventManager.getInstance();
        } catch (ExceptionInInitializerError e) {
            assertEquals("Configuration property MCR.EventHandler.Mode.Foo is not set.", e.getCause().getMessage());
        }
    }
```

Thats why I don't like this pattern for singletons. You're kind of violating the contract of the MCREventManager which wants to throw MCRConfigurationExceptions. Due to the fact that this is now initialized in static initialization, every exception will always be a ExceptionInInitializerError.

----

We should use the double locking pattern here:
```
    private static volatile MCREventManager instance;
    private MCREventManager() throws MCRConfigurationException {
    }
    public static MCREventManager getInstance() throws MCRConfigurationException {
        if (instance == null) {
            synchronized (MCREventManager.class) {
                if (instance == null) {
                    instance = new MCREventManager(); // can propagate exception
                }
            }
        }
        return instance;
    }
```

toKrause:
In the remaining  parts of MyCoRe, we now use the [Bill Pugh Singleton patern](https://www.baeldung.com/java-bill-pugh-singleton-implementation) instead of the double locking pattern everywhere.

PR #2580, opened 2025-07-17
MCR-3463 Use Java NIO API for file system resource provider

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3463).

_Replacement for https://github.com/MyCoRe-Org/mycore/pull/2563 which hat do be reverted due to administrative error._


kkrebs (approved):
already approved in #2563

PR #2582, opened 2025-07-17
MCR-3445 (2) make MCRSolrFileIndexHandler configurable

rsteph-de:


rsteph-de:
- code from `MCRSolrFileIndexHandler` was moved to `MCRSolrDefaultFileIndexHandler`
- `MCRSolrFileIndexHandler` was changed to an interface

Mewel (dismissed):

sebhofmann (changes requested):
If we merge this this should go to 2025.12(main). The better Way would be to overwrite the factory and then set IndexHandler.Factory

rsteph-de:
> If we merge this this should go to 2025.12(main). The better Way would be to overwrite the factory and then set IndexHandler.Factory

Thanks for the suggestion, that allows me to use the new behaviour in my 2025.06 LTS application without changes to the MyCoRe LTS branch. But in the future there should be ONE factory which creates configurable instances (as implemented here).

rsteph-de:
obsolete, problem solved by overriding the Indexhandlerfactory...

PR #2586, opened 2025-07-24
MCR-3441 Changes for modsperson

erodde:

Properly handle modsperson references. Search for DB entries for modsperson only if these references exist in mods. Further empty plugin point for the deletion of modsperson references before MCRMetadataManager.checkForActiveLinks() is called. Both shouldn't change behavior.


mycore-mods/src/main/java/org/mycore/mods/MCRMODSWrapper.java
```diff
@@ -200,6 +204,10 @@ public List<Element> getLinkedRelatedItems() {
         return getElements(LINKED_RELATED_ITEMS);
     }
 
+    public List<Element> getLinkedRelatedModsperson() {
```
toKrause: Nitpick:

The existing method `getLinkedRelatedItems`
- hasn't `mods` in it's name,
- has "related" ins its name, because it returns `mods:relatedItems` elements and
- is in plural form.

Better (because more fitting to the existing local style) might be `getLinkedPersons`.  
The XPath-constant should then be renamed `LINKED_PERSONS`.

The change to plural form should be done. The rest might be personal preference.

erodde: renamed methods for clarity. fun fact: correct plural of person is "people", but I'm using persons everywhere to avoid confusion :)

toKrause (approved):

erodde:
Im Falle von Merge-Konflikt: Reihenfolge Interface: Nach CheckDeletePermission, vor allem anderen.

PR #2588, opened 2025-07-25
MCR-3474 Zero-length salt causes JVM to crash on Windows

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3474).


rsteph-de (approved):
This fix works for me, thanks!

PR #2590, opened 2025-07-26
MCR-3476 fix AccessDeniedException in cleanup of tests (Windows)

rsteph-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3476).

### DIFF
--- a/mycore-base/src/test/java/org/mycore/common/MCRJPATestCase.java
+++ b/mycore-base/src/test/java/org/mycore/common/MCRJPATestCase.java
@@ -133,7 +133,7 @@ public class MCRJPATestCase extends MCRTestCase {
     }
 
     public void exportSchema() throws IOException {
-        doSchemaOperation(schema -> "create schema " + schema);
+        doSchemaOperation(schema -> "create schema if not exists " + schema);
         exportSchema("create");
     }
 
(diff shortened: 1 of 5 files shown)


mycore-base/src/test/java/org/mycore/common/MCRJPATestCase.java
```diff
@@ -133,7 +133,7 @@ public void setUp() throws Exception {
     }
 
     public void exportSchema() throws IOException {
-        doSchemaOperation(schema -> "create schema " + schema);
+        doSchemaOperation(schema -> "create schema if not exists " + schema);
```
rsteph-de: this prevents some false flag warnings if one of the tests was not cleaned up properly due to another error.

yagee-de: then it will leave test data in the database that is even harder to detect later

rsteph-de: it would be best, if we would start each test in a new environment, 
I've been looking some time into the database settings until I figured out this was not the problem

rsteph-de: but I can revert this change

yagee-de (approved):

PR #2592, opened 2025-07-28
MCR-3326 If objects could not be indexed a minimal document should be indexed instead

Mewel:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3326).

### DIFF
--- a/mycore-solr/src/main/java/org/mycore/solr/index/handlers/MCRSolrErrorIndexHandler.java
+++ b/mycore-solr/src/main/java/org/mycore/solr/index/handlers/MCRSolrErrorIndexHandler.java
@@ -1,0 +1,97 @@
+/*
+ * This file is part of ***  M y C o R e  ***
+ * See http://www.mycore.de/ for details.
+ *
+ * MyCoRe is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * MyCoRe is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with MyCoRe.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.mycore.solr.index.handlers;
+
+import java.io.IOException;
+import java.util.List;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.apache.solr.client.solrj.SolrClient;
+import org.apache.solr.client.solrj.SolrServerException;
+import org.apache.solr.client.solrj.request.UpdateRequest;
+import org.apache.solr.common.SolrInputDocument;
+import org.mycore.solr.MCRSolrConstants;
+import org.mycore.solr.MCRSolrUtils;
+import org.mycore.solr.index.MCRSolrIndexer;
+
+/**
+ * A specialized index handler responsible for indexing a minimal error document when a
+ * primary indexing operation fails. This ensures that indexing failures are discoverable
+ * directly within the Solr index.
+ * <p>
+ * The handler creates a document with the following fields:
+ * <ul>
+ *   <li><b>id</b>: The ID of the document that failed to index.</li>
+ *   <li><b>error_message</b>: A string containing the exception message.</li>
+ *   <li><b>error_stacktrace</b>: The stacktrace of the error.</li>
+ * </ul>
+ *
+ * @author Matthias Eichner
+ */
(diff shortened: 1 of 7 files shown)


mycore-solr/src/main/java/org/mycore/solr/index/handlers/document/MCRSolrInputDocumentHandler.java
```diff
@@ -31,54 +32,108 @@
...
+        } catch (Exception e) {
+            LOGGER.error(() -> "Error while creating solr document '" + id + "'.", e);
+            this.subHandlers.add(new MCRSolrErrorIndexHandler(id, e, solrClients));
+            return;
         }
+        String docId = String.valueOf(document.getFieldValue("id"));
+
+        LOGGER.info("Sending {} to SOLR...", docId);
         UpdateRequest updateRequest = getUpdateRequest(MCRSolrConstants.SOLR_UPDATE_PATH);
-        getSolrAuthenticationFactory().applyAuthentication(updateRequest,
```
toKrause, line 74: You removed `applyAuthentication` in two places and never added it back in. Is this an oversight, or is this actually not necessary?

mycore-solr/src/main/resources/components/solr/config/mycore.properties
```diff
@@ -77,6 +77,7 @@ MCR.Solr.ConcurrentUpdateSolrClient.ThreadCount=%MCR.Solr.Indexer.ThreadCount%
 
 MCR.Solr.SolrInputDocument.Factory=org.mycore.solr.index.document.MCRSolrTransformerInputDocumentFactory
 MCR.Solr.SolrInputDocument.Transformer=mycoreobject-solrdocument
+MCR.Solr.SolrInputDocument.MinimalFields=objectType,objectProject,derivateID,fileName,filePath,returnId,derivateMaindoc,created,modified,parent
```
toKrause: Would you consider adding `state` to the default list of minimal fields. Usually, each object always has a state and other components might (not unreasonably) assume, that that field also always exist. See for example [MIR-1474](https://mycore.atlassian.net/browse/MIR-1474).

mycore-solr/src/main/java/org/mycore/solr/index/handlers/MCRSolrErrorIndexHandler.java
```diff
@@ -0,0 +1,97 @@
...
+     * Creates and indexes a minimal document containing error information.
+     *
+     * @throws IOException if an I/O error occurs.
+     * @throws SolrServerException if a base Solr error occurs.
+     */
+    @Override
+    public void index() throws IOException, SolrServerException {
+        SolrInputDocument errorDoc = buildDocument(failedId, cause);
+
+        UpdateRequest errorRequest = getUpdateRequest(MCRSolrConstants.SOLR_UPDATE_PATH);
```
toKrause, line 81: I'm nit a Solr-export, so just wondering:

Why is sth. like 

```java
for (SolrClient solrClient : solrClients) {
      if (MCRSolrUtils.useNestedDocuments()) {
             MCRSolrIndexer.deleteById(solrClient, docId);
      }
}
```

not necessary in this handler?

mycore-solr/src/main/java/org/mycore/solr/index/handlers/document/MCRSolrInputDocumentHandler.java
Mewel: The applyAuthentication() is already done in getUpdateRequest(). The call was a duplicate.

mycore-solr/src/main/resources/components/solr/config/mycore.properties
Mewel: sure!

mycore-solr/src/main/java/org/mycore/solr/index/handlers/MCRSolrErrorIndexHandler.java
Mewel: Uhmm not sure if I understand you correctly, but it is there, in the for loop:
```
       for (SolrClient solrClient : clients) {
            try {
                if (MCRSolrUtils.useNestedDocuments()) {
                    MCRSolrIndexer.deleteById(solrClient, failedId);
                }
                errorRequest.process(solrClient);
                LOGGER.info("Successfully indexed error document for id '{}'.", failedId);
            } catch (Exception ex) {
                LOGGER.error(() -> "FATAL: Could not index error document for id '" + failedId + "'.", ex);
            }
        }
```

toKrause: Ups, sorry, I must have been blind, ...

Mewel: The problem with the code before was that it looped two times over the solr clients.
First it deleted all the documents and then it indexed it again. But if a problem arose in lets say the first one (an exception) then the other clients would be skipped, their document would have been deleted just because the first client threw an exception.

toKrause (approved):

PR #2596, opened 2025-07-28
MCR-3479 Harmonize and modernize MCRClassificationMappingEventHandler implementations

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3479).


sebhofmann:
https://github.com/MyCoRe-Org/mycore/pull/2467 already handles this

toKrause:
In https://github.com/MyCoRe-Org/mycore/pull/2467 there is still copious amounts of code duplication and some behavioral inconsistencies in `MCRDefaultClassificationMapper` and `MCRMODSClassificationMapper` (that certainly should be addressed in a separate PR)

sebhofmann:
Yes, but classification mapping is not done in EventHandlers anymore, since the Information/Metadata they create, can always be derived from existing metadata. 
It would be better to clean up the duplicate Code when my branch is merged.

PR #2606, opened 2025-08-07
MCR-3483 object lock for REST API

Mewel:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3483).

### DIFF
--- a/mycore-restapi/src/main/java/org/mycore/restapi/v2/MCRRestObjects.java
+++ b/mycore-restapi/src/main/java/org/mycore/restapi/v2/MCRRestObjects.java
@@ -47,6 +47,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.UUID;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
@@ -62,6 +63,7 @@ import org.mycore.access.MCRAccessManager;
 import org.mycore.common.MCRCoreVersion;
 import org.mycore.common.MCRException;
 import org.mycore.common.MCRPersistenceException;
+import org.mycore.common.MCRSessionMgr;
 import org.mycore.common.config.MCRConfiguration2;
 import org.mycore.common.content.MCRBaseContent;
 import org.mycore.common.content.MCRContent;
@@ -82,6 +84,8 @@ import org.mycore.datamodel.niofs.MCRPath;
 import org.mycore.datamodel.objectinfo.MCRObjectQuery;
 import org.mycore.datamodel.objectinfo.MCRObjectQueryResolver;
 import org.mycore.frontend.jersey.MCRCacheControl;
+import org.mycore.frontend.support.MCRObjectIDLockTable;
+import org.mycore.frontend.support.MCRObjectLock;
 import org.mycore.media.services.MCRThumbnailGenerator;
 import org.mycore.restapi.annotations.MCRAccessControlExposeHeaders;
 import org.mycore.restapi.annotations.MCRApiDraft;
@@ -106,6 +110,7 @@ import io.swagger.v3.oas.annotations.media.Schema;
 import io.swagger.v3.oas.annotations.parameters.RequestBody;
 import io.swagger.v3.oas.annotations.responses.ApiResponse;
 import io.swagger.v3.oas.annotations.tags.Tag;
+
 import jakarta.servlet.ServletContext;
 import jakarta.ws.rs.BadRequestException;
 import jakarta.ws.rs.Consumes;
@@ -115,6 +120,7 @@ import jakarta.ws.rs.ForbiddenException;
 import jakarta.ws.rs.GET;
 import jakarta.ws.rs.InternalServerErrorException;
 import jakarta.ws.rs.NotFoundException;
+import jakarta.ws.rs.PATCH;
 import jakarta.ws.rs.POST;
 import jakarta.ws.rs.PUT;
 import jakarta.ws.rs.Path;
@@ -138,6 +144,7 @@ import jakarta.xml.bind.annotation.XmlElementWrapper;
         description = "Operations on derivates belonging to metadata objects"),
     @Tag(name = MCRRestUtils.TAG_MYCORE_FILE, description = "Operations on files in derivates"),
 })
+@SuppressWarnings("PMD.AvoidDuplicateLiterals")
 public class MCRRestObjects {
 
     public static final String PARAM_AFTER_ID = "after_id";
@@ -747,12 +754,7 @@ public class MCRRestObjects {
     @MCRRequireTransaction
     public Response deleteObject(@PathParam(PARAM_MCRID) MCRObjectID id) {
         //check preconditions
-        if (!MCRMetadataManager.exists(id)) {
-            throw MCRErrorResponse.ofStatusCode(Response.Status.NOT_FOUND.getStatusCode())
-                .withErrorCode(MCRErrorCodeConstants.MCROBJECT_NOT_FOUND)
-                .withMessage(buildNotFoundMessage(id))
-                .toException();
-        }
+        checkExists(id);
         try {
             MCRMetadataManager.deleteMCRObject(id);
         } catch (MCRActiveLinkException e) {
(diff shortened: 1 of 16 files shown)


rsteph-de (dismissed):
MCRRestAPIObjects is getting longer ...

Is there a possibility to separate the Lock-API calls into another class?
This can be done in a follow-up PR which cleans up the REST-API code.

Otherwise O.K.

Mewel:
> MCRRestAPIObjects is getting longer ...
> 
> Is there a possibility to separate the Lock-API calls into another class? This can be done in a follow-up PR which cleans up the REST-API code.
> 
> Otherwise O.K.

I'm doing some changes right now anyway. I could also move it separate  if you like.

Edit: question is where too?

rsteph-de:
> I'm doing some changes right now anyway. I could also move it separate if you like.

This could show as "proof-of-concept" - that it is possible to have multiple RESTController classes  with the same path-prefix.

mycore-restapi/src/main/java/org/mycore/restapi/v2/MCRRestObjects.java
```diff
@@ -881,8 +875,78 @@ public Response getState(@PathParam(PARAM_MCRID) MCRObjectID id) {
...
+                description = "Lock information about an object",
+                content = @Content(
+                    schema = @Schema(implementation = MCRObjectLock.class)))
+        })
+    @SuppressWarnings("PMD.AvoidDuplicateLiterals")
+    public MCRObjectLock setLock(@PathParam(PARAM_MCRID) MCRObjectID id, MCRObjectLock requestBody) {
+        return lockService.setLock(id, requestBody);
+    }
+
+    @PATCH
```
yagee-de: Should be PUT, as PATCH should be used for partial updates. Here a complete MCRObjectLock is required.

mycore-restapi/src/main/java/org/mycore/restapi/v2/MCRRestObjects.java
```diff
@@ -881,8 +875,78 @@ public Response getState(@PathParam(PARAM_MCRID) MCRObjectID id) {
...
+
+    @POST
+    @Produces({ MediaType.APPLICATION_JSON, MediaType.APPLICATION_XML })
+    @Path("/{" + PARAM_MCRID + "}/lock")
+    @Operation(summary = "Lock object {" + PARAM_MCRID + "}.",
+        tags = MCRRestUtils.TAG_MYCORE_OBJECT,
+        requestBody = @RequestBody(content = @Content(schema = @Schema(implementation = MCRObjectLock.class))),
+        responses = {
+            @ApiResponse(responseCode = NOT_FOUND, description = "object is not found"),
+            @ApiResponse(responseCode = CONFLICT, description = "object is already locked"),
```
yagee-de: POST is for adding new uncounted resources. Here a maximum of 1 is allowed. So a better approach would be a PUT here.

mycore-restapi/src/main/java/org/mycore/restapi/v2/MCRRestObjects.java
```diff
@@ -881,8 +875,78 @@ public Response getState(@PathParam(PARAM_MCRID) MCRObjectID id) {
...
+
+    @DELETE
+    @Produces({ MediaType.APPLICATION_JSON, MediaType.APPLICATION_XML })
+    @Path("/{" + PARAM_MCRID + "}/lock")
+    @Operation(summary = "Unlock object {" + PARAM_MCRID + "}.",
+        tags = MCRRestUtils.TAG_MYCORE_OBJECT,
+        responses = {
+            @ApiResponse(responseCode = NOT_FOUND, description = "object or lock is not found"),
+            @ApiResponse(responseCode = CONFLICT,
+                description = "object is already unlocked or is locked by another user"),
```
yagee-de: What is the difference between "lock not found" and "already unlocked"?

mycore-restapi/src/main/java/org/mycore/restapi/v2/MCRRestV2App.java
```diff
@@ -22,6 +22,16 @@
...
 
+import com.fasterxml.jackson.jakarta.rs.json.JacksonXmlBindJsonProvider;
+import io.swagger.v3.jaxrs2.integration.JaxrsOpenApiContextBuilder;
+import io.swagger.v3.jaxrs2.integration.resources.OpenApiResource;
+import io.swagger.v3.oas.integration.OpenApiConfigurationException;
+import io.swagger.v3.oas.integration.SwaggerConfiguration;
+import io.swagger.v3.oas.models.OpenAPI;
+import io.swagger.v3.oas.models.info.Info;
+import io.swagger.v3.oas.models.info.License;
+import io.swagger.v3.oas.models.servers.Server;
```
yagee-de: import order has changed:
should be:
java
-empty line-
javax
-empty line-
org
-empty line-
com
-empty line-
rest

As far as I remember.

mycore-restapi/src/main/java/org/mycore/restapi/v2/service/MCRRestObjectLockService.java
yagee-de, line 1: Should be updated according to changes in REST API.

yagee-de (changes requested):

mycore-restapi/src/main/java/org/mycore/restapi/v2/MCRRestObjects.java
Mewel: This is a bit confusing. So you want a PUT on both setLock() and updateLock(). This obviosly doesn't work. Do you want a single putLock() method?

mycore-restapi/src/main/java/org/mycore/restapi/v2/MCRRestObjects.java
Mewel: The API doc is wrong here, should be "object lock is from another user".

mycore-restapi/src/main/java/org/mycore/restapi/v2/MCRRestObjects.java
```diff
@@ -202,6 +205,9 @@ public class MCRRestObjects {
     @Context
     UriInfo uriInfo;
 
+    @Inject
```
rsteph-de, line 208: This is the introduction of the Jersey Dependency Injection mechanismn in MyCoRe (first use).
In my opinion this is a architectural decision that adds "complexity" to the code and should be discussed first in an online meeting..
I tend to stay with "our" MCRConfiguration2.getInstance().

Mewel: I can see your point. And im ok with changing it.

But in general I prefer the default way and for me that would be using the @Inject method, cause its out of the box available without any boilerplate code (and tbh our property injection stuff comes with a lot of it...). The binding happens in MCRRestV2App. I could see using our property configuration there.

We also use the @Singleton annotation from the same package. So we kind of mix the stuff already.

mycore-restapi/src/main/java/org/mycore/restapi/v2/MCRRestV2App.java
```diff
@@ -71,6 +76,17 @@ protected String getVersion() {
         return "v2";
     }
 
+    private void registerServices() {
+        // MCRRestObjectLockService
+        register(new AbstractBinder() {
+            @Override
+            protected void configure() {
+                bind(MCRRestObjectLockServiceImpl.class)
+                    .to(MCRRestObjectLockService.class);
```
yagee-de: It's either hard coded (no interface) or configurable.

yagee-de (changes requested):
Good work, but a minor issue between interface, implementation and configuration. Maybe the interface is not necessary?

mycore-restapi/src/main/java/org/mycore/restapi/v2/MCRRestV2App.java
Mewel: I can make it configurable.
But an interface is in both cases good. For example if somebody decides to write tests for the MCRRestObjects class. Then you can easily mock the ObjectLockService interface.

yagee-de (dismissed):

sebhofmann (approved):

PR #2620, opened 2025-08-14
MCR-3490 Allow conditional mappings in x-mapping-values (for example, based on the parents genre)

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3490).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/common/xml/MCRXPathEvaluator.java
+++ b/mycore-base/src/main/java/org/mycore/common/xml/MCRXPathEvaluator.java
@@ -78,7 +78,7 @@ public class MCRXPathEvaluator {
     }
 
     public String replaceXPathOrI18n(String expression) {
-        return evaluateXPath(migrateLegacyI18nSyntaxToExtensionFunction(expression));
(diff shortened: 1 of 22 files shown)


toKrause:
Split into two PRs.

mycore-base/src/main/java/org/mycore/common/xml/MCRXPathEvaluator.java
```diff
@@ -97,11 +97,21 @@ private String migrateLegacyI18nSyntaxToExtensionFunction(String expression) {
         return updatedI18nExpression;
     }
 
+    @Deprecated(forRemoval = true)
```
sebhofmann, line 103: There should be a javadoc indicating that the function #evaluateFirstAsString is the replacement

sebhofmann (changes requested):
Just one little change.

mycore-base/src/main/java/org/mycore/common/xml/MCRXPathEvaluator.java
toKrause: good point

kkrebs (approved):

sebhofmann (approved):

PR #2621, opened 2025-08-18
MCR-3492 Test class relying on configuration to be initialized in preparation fails, if run individually

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3492).


kkrebs (approved):
looks good for me

PR #2626, opened 2025-08-19
MCR-3494 missing module handling in IViewConfig*.xsl

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3494).


bt1cu (approved):
tested with Leipzig Papyrus Project, it works

PR #2643, opened 2025-08-21
MCR-3441 expand MCRNameMerger for modsperson

erodde:
I expanded the MCRNameMerger in UBO for modsperson.

1. It can handle <mods:alternativeName> now and considers them when determining if two names are probably the same.
2. If first name is the same and one of the family names contains the other, the names are considered as "probablySameAs". This takes double-barreled names (Doppelnamen) into account.


erodde:
TODO: SpecializedNameMerger that can be configured. alternativeName should (probably) still be in the base MCRNameMerger, needs further consideration.

erodde:
PR is closed and replaced by https://github.com/MyCoRe-Org/mycore/pull/2774.

PR #2648, opened 2025-08-21
MCR-3498 Make default rules for xml cleaner configurable

fluetze:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3498).

### DIFF
--- a/mycore-xeditor/src/main/java/org/mycore/frontend/xeditor/cleanup/MCRCleaningRule.java
+++ b/mycore-xeditor/src/main/java/org/mycore/frontend/xeditor/cleanup/MCRCleaningRule.java
@@ -1,0 +1,53 @@
+package org.mycore.frontend.xeditor.cleanup;
+
+import java.util.List;
+
+import org.jdom2.Document;
+import org.jdom2.filter.Filters;
+import org.jdom2.xpath.XPathExpression;
+import org.jdom2.xpath.XPathFactory;
+import org.mycore.common.MCRConstants;
+
+class MCRCleaningRule {
+
+    private String xPathExprNodesToInspect;
+
+    private XPathExpression<Object> xPathNodesToInspect;
+
+    private XPathExpression<Object> xPathRelevancyTest;
+
+    MCRCleaningRule(String xPathExprNodesToInspect, String xPathExprRelevancyTest) {
+        this.xPathExprNodesToInspect = xPathExprNodesToInspect;
+        this.xPathNodesToInspect = XPathFactory.instance().compile(xPathExprNodesToInspect, Filters.fpassthrough(),
+            null,
+            MCRConstants.getStandardNamespaces());
+        this.xPathRelevancyTest = XPathFactory.instance().compile(xPathExprRelevancyTest, Filters.fpassthrough(), null,
+            MCRConstants.getStandardNamespaces());
+    }
+
+    public List<Object> getNodesToInspect(Document xml) {
+        return xPathNodesToInspect.evaluate(xml);
+    }
+
+    public boolean isRelevant(Object node) {
+        Object found = xPathRelevancyTest.evaluateFirst(node);
+        if (found == null) {
+            return false;
+        } else if (found instanceof Boolean b) {
+            return b;
+        } else {
+            return true; // something matching found
+        }
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        return obj instanceof MCRCleaningRule cleaningRule
+            && xPathExprNodesToInspect.equals(cleaningRule.xPathExprNodesToInspect);
+    }
+
+    @Override
+    public int hashCode() {
+        return xPathExprNodesToInspect.hashCode();
+    }
+}
--- a/mycore-xeditor/src/main/java/org/mycore/frontend/xeditor/cleanup/MCRCleaningRule.java
+++ b/mycore-xeditor/src/main/java/org/mycore/frontend/xeditor/cleanup/MCRCleaningRule.java
@@ -19,6 +19,7 @@
 package org.mycore.frontend.xeditor.cleanup;
 
 import java.util.List;
+import java.util.Objects;
 
 import org.jdom2.Document;
 import org.jdom2.filter.Filters;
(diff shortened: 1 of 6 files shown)


mycore-xeditor/src/main/java/org/mycore/frontend/xeditor/cleanup/MCRDefaultRules.java
```diff
@@ -0,0 +1,58 @@
...
+ * 
+ * @author Frank L\u00FCtzenkirchen
+ */
+class MCRDefaultRules {
+
+    private static final String CONFIG_PREFIX = "MCR.XEditor.XMLCleaner.DefaultRule.";
+    private static final String CONFIG_KEY_NODES_TO_INSPECT = "NodesToInspect";
+    private static final String CONFIG_KEY_RELEVANT_IF = "RelevantIf";
+
+    private static final List<MCRCleaningRule> DEFAULT_RULES = new ArrayList<>();
```
toKrause: This list should probably be unmodifiable.

mycore-xeditor/src/main/java/org/mycore/frontend/xeditor/cleanup/MCRCleaningRule.java
```diff
@@ -0,0 +1,74 @@
...
+ *   <li>The XPath to select the nodes to inspect for this rule
+ *   <li>The XPath expression to decide if this node is relevant or should be removed
+ * </ul>
+ * 
+ * @author Frank L\u00FCtzenkirchen
+ */
+class MCRCleaningRule {
+
+    /** The XPath to select the nodes to inspect for this rule */
+    private String xPathExprNodesToInspect;
```
toKrause: Nitpick: 

Your XPaths are named `xPathExpr....` and your `XPathExpressions` are named `xPath...`. I would probably switch that around.

mycore-xeditor/src/main/java/org/mycore/frontend/xeditor/cleanup/MCRCleaningRule.java
```diff
@@ -0,0 +1,74 @@
...
+            return false;
+        } else if (found instanceof Boolean b) {
+            return b;
+        } else {
+            return true; // something matching found
+        }
+    }
+
+    @Override
+    public boolean equals(Object obj) {
```
toKrause, line 81: Shouldn't `xPathExprRelevancyTest`also be part of `equals` and `hashCode`?

mycore-xeditor/src/main/java/org/mycore/frontend/xeditor/cleanup/MCRCleaningRule.java
```diff
@@ -0,0 +1,74 @@
+package org.mycore.frontend.xeditor.cleanup;
```
toKrause, line 19: Missing license headers in multiple files.

mycore-xeditor/src/main/java/org/mycore/frontend/xeditor/cleanup/MCRDefaultRules.java
```diff
@@ -0,0 +1,58 @@
...
+
+    static {
+        Map<String, String> config = MCRConfiguration2.getSubPropertiesMap(CONFIG_PREFIX);
```
toKrause, line 63: I'd like to generally promote usage of the configuration API instead of custom property parsing (not a big issue here, because we're only talking about two sings, but still, I'd like to fight the bane of MyCoRe that is the ungodly amount of different ways components feed themselves from the configuration), so just a suggestion:

Instantiation with `MCRConfiguration2:getInstanceOfOrThrow` (more convenient ways to instantiate collections like this will be added in the future, something like this should really be a one-liner)
```java
static {
    DEFAULT_RULES = MCRConfiguration2.getPropertiesMap()
        .keySet()
        .stream()
        .filter(s -> s.startsWith(CONFIG_PREFIX))
        .map(key -> key.substring(0, key.indexOf('.', key.indexOf('.', CONFIG_PREFIX.length()) + 1)))
        .distinct()
        .sorted()
        .map(rulePrefix -> MCRConfiguration2.getInstanceOfOrThrow(MCRCleaningRule.class, rulePrefix))
        .toList();
}
```

Configuration with annotations and a `MCRConfigurationProxy`
```java
@MCRConfigurationProxy(proxyClass = MCRCleaningRule.Factory.class)
public final class MCRCleaningRule {

    public static final String CONFIG_KEY_NODES_TO_INSPECT = "NodesToInspect";

    public static final String CONFIG_KEY_RELEVANT_IF = "RelevantIf";

    /** The XPath to select the nodes to inspect for this rule */
    private String xPathExprNodesToInspect;

    ...

    @Override
    public int hashCode() {
        return xPathExprNodesToInspect.hashCode();
    }

    public static class Factory implements Supplier<MCRCleaningRule> {

        @MCRProperty(name = CONFIG_KEY_NODES_TO_INSPECT)
        public String nodesToInspect;

        @MCRProperty(name = CONFIG_KEY_RELEVANT_IF)
        public String relevantIf;

        @Override
        public MCRCleaningRule get() {
            return new MCRCleaningRule(nodesToInspect, relevantIf);
        }

    }

}
```

toKrause (changes requested):

toKrause (approved):

mycore-xeditor/src/main/java/org/mycore/frontend/xeditor/cleanup/MCRDefaultRules.java
fluetze: Thx for the tutorial. I will use that at other opportunities.

sebhofmann: You can use 

```
      MCRConfiguration2.<MCRCleaningRule>getInstances(CONFIG_PREFIX).values().stream().map(c-> {
        try {
          return c.call();
        } catch (Exception e) {
          throw new MCRConfigurationException("Could not get instance for " + name, e);
        }
      }).toList();
```

for that first code snipped, or you can just store the callable and create new instances when needed.:
```
MCRConfiguration2.<MCRCleaningRule>getInstances(CONFIG_PREFIX).values()
```

toKrause: I don't think that that works with the order-preserving way (that still allows entries with meanigful names) the config keys are designed here.

PR #2683, opened 2025-09-22
MCR-3517 resolve build issues

yagee-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3517).

This pull request includes several updates focused on improving build tooling and simplifying code in the `mycore-base` module. The most significant changes are updates to Maven plugin versions, enhancements to build configuration, and minor code cleanups.

**Build tooling and configuration:**

* Upgraded the `git-commit-id-maven-plugin` version from `5.0.0` to `9.0.2` in `mycore-base/pom.xml` for improved compatibility and features.
* Added the `<pmd.analysisCache>` property set to `true` in `pom.xml` to enable caching for PMD analysis, which can improve build performance.
* Added `<skipSourcesUnchanged>true</skipSourcesUnchanged>` to the Hibernate processor configuration in `pom.xml` to skip processing unchanged sources and speed up incremental builds.

**Codebase simplification:**

* Updated logging in the `getSource` method of `MCRConfigurationDir.java` to use `LogManager.getLogger()` directly instead of the removed static `LOGGER` field.

### DIFF
--- a/mycore-base/src/main/java/org/mycore/common/config/MCRConfigurationDir.java
+++ b/mycore-base/src/main/java/org/mycore/common/config/MCRConfigurationDir.java
@@ -25,7 +25,6 @@ import java.security.ProtectionDomain;
 import java.util.Locale;
 
 import org.apache.logging.log4j.LogManager;
-import org.apache.logging.log4j.Logger;
 import org.mycore.common.MCRUtils;
 
 import jakarta.servlet.ServletContext;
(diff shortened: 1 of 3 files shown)


mycore-base/src/main/java/org/mycore/common/config/MCRConfigurationDir.java
```diff
@@ -115,7 +112,7 @@ private static String getSource(Class<MCRConfigurationDir> clazz) {
         ProtectionDomain protectionDomain = clazz.getProtectionDomain();
         CodeSource codeSource = protectionDomain.getCodeSource();
         if (codeSource == null) {
-            LOGGER.error(() -> "Cannot get CodeSource for class " + clazz.getSimpleName());
```
sebhofmann, line 118: Why you remove that field? The PR Says its unused, but that is not true.

sebhofmann (changes requested):

mycore-base/src/main/java/org/mycore/common/config/MCRConfigurationDir.java
yagee-de: Required for MCRCoreVersion class as LOGGER is not initialized

sebhofmann (approved):

PR #2698, opened 2025-10-07
MCR-3441 delete MCRPreDeleteAction as not necessary anymore

erodde:
As deletion of modsperson objects is now managed differently in UBO, this action is not needed anymore and can be removed.


rsteph-de:
Just to check... The property was not introduced in one of the mycore.properties?

sebhofmann (approved):

erodde:
> Just to check... The property was not introduced in one of the mycore.properties?

I rechecked the original commit, but no. My plan was to use the interface in UBO.

PR #2704, opened 2025-10-15
MCR-3532 Log message uniqueness filter not working in 2024.06.x

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3532).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/common/log4j2/filter/MCRUniqueFilter.java
+++ b/mycore-base/src/main/java/org/mycore/common/log4j2/filter/MCRUniqueFilter.java
@@ -33,7 +33,7 @@ import org.apache.logging.log4j.core.config.plugins.Plugin;
 import org.apache.logging.log4j.core.config.plugins.PluginAttribute;
 import org.apache.logging.log4j.core.config.plugins.PluginFactory;
 import org.apache.logging.log4j.core.filter.AbstractFilter;
-import org.apache.logging.log4j.message.ParameterizedMessage;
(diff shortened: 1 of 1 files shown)


toKrause:
Related PR: https://github.com/MyCoRe-Org/mycore/pull/2025

mycore-base/src/main/java/org/mycore/common/log4j2/filter/MCRUniqueFilter.java
```diff
@@ -56,17 +56,17 @@ private MCRUniqueFilter(int cacheSizePerFormat, Result onMatch, Result onMismatc
...
-            List<Object> parameters = Arrays.asList(message.getParameters());
-            Set<List<Object>> cache = getCache(message.getFormat());
-            if (cache.contains(parameters)) {
-                return onMatch;
-            } else {
-                cache.add(parameters);
-            }
+        Message message = event.getMessage();
+        List<Object> parameters = Arrays.asList(message.getParameters());
+        Set<List<Object>> cache = getCache(message.getFormat());
```
toKrause, line 63: `Message#getFormat()` is deprecated on the interface `Message`, but not on the ~30 implementations, hence the `@SuppressWarnings("deprecation")`. We could switch over the ~30 implementations and do the same thing in each case, but I think that would be absurd. This approach is much more pragmatic, even if not perfect. But we have to live with the design choices of Log4j and there is no indication that the method in question will be removed.

sebhofmann (approved):

PR #2707, opened 2025-10-16
MCR-3533 CLI unable to process commands with trailing white-space

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3533).

### DIFF
--- a/mycore-base/src/main/java/org/mycore/frontend/cli/MCRCommandLineInterface.java
+++ b/mycore-base/src/main/java/org/mycore/frontend/cli/MCRCommandLineInterface.java
@@ -38,6 +38,7 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Queue;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.regex.Pattern;
 import java.util.stream.Stream;
 
 import org.apache.commons.text.StringSubstitutor;
@@ -147,25 +148,14 @@ public class MCRCommandLineInterface {
     private static void readCommandFromArguments(String[] args) {
         if (args.length > 0) {
             interactiveMode = false;
-
-            String line = readLineFromArguments(args);
-            addCommands(line);
+            addCommands(String.join(" ", args));
         }
     }
 
     private static void addCommands(String line) {
-        Stream.of(line.split(";;"))
-            .map(String::trim)
+        Stream.of(line.split(Pattern.quote(" ;; ")))
             .filter(s -> !s.isEmpty())
             .forEachOrdered(COMMAND_QUEUE::add);
-    }
-
-    private static String readLineFromArguments(String[] args) {
-        StringBuilder sb = new StringBuilder();
-        for (String arg : args) {
-            sb.append(arg).append(' ');
-        }
-        return sb.toString();
     }
 
     private static void initSession() {


mycore-base/src/main/java/org/mycore/frontend/cli/MCRCommandLineInterface.java
```diff
@@ -147,27 +148,16 @@ public static void main(String[] args) {
...
-
-            String line = readLineFromArguments(args);
-            addCommands(line);
+            addCommands(String.join(" ", args));
         }
     }
 
     private static void addCommands(String line) {
-        Stream.of(line.split(";;"))
-            .map(String::trim)
```
rsteph-de, line 158: Why is .map(String::trim)`` removed?
I use this class to process commands from an xml file (with linebreaks and indention).
That means the commands need to be cleaned up. 
I could not find code which does that later.

rsteph-de: or maybe use another pattern like line.split("\\s+;;\\s+") - (syntax not guaranted)

rsteph-de (changes requested):
see comment

mycore-base/src/main/java/org/mycore/frontend/cli/MCRCommandLineInterface.java
toKrause: > Why is .map(String::trim)`` removed?

That ist the main point of this PR. Some commands have a necessary trailing space character. The trimming makes it impossible for the CLI to process such commands. See Jira ticket for an example.

toKrause: Can you provide an example XML file and instructions on how to feed it to the CLI?

toKrause: > or maybe use another pattern like line.split("\s+;;\s+") - (syntax not guaranted)

That would also prevent the CLI to process a command with trailing white-space, if it is not the last command in the chain.

rsteph-de: I forgot that you mentioned the use case with the trailing space in our last telco.

I use the CLI for initializing / configuring my applications with Ant scripts:
e.g. https://github.com/MyCoRe-Org/professorenkatalog-sample/blob/301790315f61dac9311b8e7385460b210865c49d/build.xml#L71

rsteph-de: AI suggests 2 different codes for trimming and non trimming separators.
Since I have the rarer use case I suggest `;;;` for the trimming variant.

A command ending with `" "` is a bad idea.

toKrause: > A command ending with `" "` is a bad idea.

I agree, but, sadly,  they do exist, ... and should be supported.

toKrause: I'm not an Ant-user, so this is more of a guess,  but would the following work for you in conjunction with this PR?

```xml
<java fork="true" dir="${CLI_HOME}/lib" classpathref="CLI_CLASSPATH" classname="org.mycore.frontend.cli.MCRCommandLineInterface">
	<sysproperty key="MCR.AppName" value="${APP_NAME}" />
	<arg value="update all classifications from directory ${CONFIG_DIR}/classification" />
	<arg value=";;" />
	<arg value="update classification from uri resource:config/profkat-base/classification/derivate_types.xml" />
	<arg value=";;" />
	<arg value="update classification from uri resource:config/profkat-base/classification/state.xml" />
</java>
```
Or maybe, depending on how Ant concats the args:

```xml
<java fork="true" dir="${CLI_HOME}/lib" classpathref="CLI_CLASSPATH" classname="org.mycore.frontend.cli.MCRCommandLineInterface">
	<sysproperty key="MCR.AppName" value="${APP_NAME}" />
	<arg value="update all classifications from directory ${CONFIG_DIR}/classification" />
	<arg value=" ;; " />
	<arg value="update classification from uri resource:config/profkat-base/classification/derivate_types.xml" />
	<arg value=" ;; " />
	<arg value="update classification from uri resource:config/profkat-base/classification/state.xml" />
</java>
```

rsteph-de (approved):

PR #2709, opened 2025-10-22
MCR-3512 NullPointerException when publishing a new publication to orcid

Possommi:
[MCR-3512](https://mycore.atlassian.net/browse/MCR-3512).

[MCR-3512]: https://mycore.atlassian.net/browse/MCR-3512?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mycore-orcid2/src/main/java/org/mycore/orcid2/metadata/MCRORCIDPutCodeInfo.java
+++ b/mycore-orcid2/src/main/java/org/mycore/orcid2/metadata/MCRORCIDPutCodeInfo.java
@@ -115,11 +115,11 @@ public class MCRORCIDPutCodeInfo {
     /**
      * Sets other put codes.
      *
-     * @param putCodes other put codes
+     * @param putCodes other put codes, must not be <code>null</code>
      */
     public void setOtherPutCodes(long[] putCodes) {
         if(putCodes == null) {
-            return;
+            throw new IllegalArgumentException("putCodes must not be null");
         }
         this.otherPutCodes = Arrays.copyOf(putCodes, putCodes.length);
     }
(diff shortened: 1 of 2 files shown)


toKrause (approved):

golsch (changes requested):
The problem is at this point: https://github.com/MyCoRe-Org/mycore/blob/2024.06.x/mycore-orcid2/src/main/java/org/mycore/orcid2/metadata/MCRORCIDMetadataUtils.java#L151

Here, we have to ensure that the array is emptied, because null is not actually the desired behavior.

mycore-orcid2/src/main/java/org/mycore/orcid2/metadata/MCRORCIDPutCodeInfo.java
```diff
@@ -118,6 +118,9 @@ public long[] getOtherPutCodes() {
      * @param putCodes other put codes
      */
     public void setOtherPutCodes(long[] putCodes) {
+        if(putCodes == null) {
+            return;
+        }
```
golsch, line 123: This should be removed.

golsch (changes requested):

mycore-orcid2/src/main/java/org/mycore/orcid2/metadata/MCRORCIDPutCodeInfo.java
Possommi: That is a valid `null` check. Maybe throwing an `IllegalArgumentException` would be an alternative.

golsch: That's right. But if someone tries to reset the array with null again, there won't be an error. Then the array wouldn't be reset and you could spend hours debugging.

golsch (approved):

PR #2712, opened 2025-10-24
MCR 3537 Update orcid works is no longer possible

golsch:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3537).

requires https://github.com/MyCoRe-Org/mycore/pull/2709


Possommi (approved):
works when https://github.com/MyCoRe-Org/mycore/pull/2709 is applied

PR #2715, opened 2025-10-29
MCR-3542 WCMS not working

Mewel:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3542).

* fix wcmsType due to PMD changes
* fix const/let bug while writing
* use new 1.7.12 version to get rid of i18n errors


toKrause (approved):
I've tested this PR with a fresh install of MIR 2024.06 and the problem seems to be fixed.

PR #2716, opened 2025-10-29
MCR-3539 fix session listing

Mewel:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3539).


toKrause (approved):
I've tested this PR with a fresh install of MIR 2025.06 and the problem seems to be fixed.

PR #2724, opened 2025-11-10
MCR-3551 Allow hiding of publicly available derivates

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3551).


rsteph-de (changes requested):
This is a new use case for my utility method to create the boolean result of an URIResolver.

toKrause:
> This is a new use case for my utility method to create the boolean result of an URIResolver.

Good point. Fixed.

rsteph-de (approved):

PR #2725, opened 2025-11-11
MCR-3545 Category-Link-Table uses normalized object

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3545).


sebhofmann:
The Objects are now expanded before creating links to classifications. The link to objects is now added before the link to classifications, because the links are maybe needed in the object expansion.
The `MCRMODSLinkProviderTest` is now broken, because the `link_changed` event is also triggered when a object is created (not only modyfied). 
The `MCRMODSLinkProviderTest`  failes because the object links to a modsperson which does not exist.
We need to decide if Objects should always exists for all `link_changed` events and adapt the test or if we want to treat it as an error.

sebhofmann:
We decided that the links should always exist, so i fixed the test by creating a modsperson.

rsteph-de (approved):

PR #2728, opened 2025-11-11
MCR-3554 Updated toString() in MCRUser

Possommi:
[MCR-3554](https://mycore.atlassian.net/browse/MCR-3554).

[MCR-3554]: https://mycore.atlassian.net/browse/MCR-3554?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mycore-user2/src/main/java/org/mycore/user2/MCRUser.java
+++ b/mycore-user2/src/main/java/org/mycore/user2/MCRUser.java
@@ -499,7 +499,7 @@ public class MCRUser implements MCRUserInformation, Cloneable, Serializable {
     @Transient
     @Override
     public String toString() {
-        return getUserID();
+        return getUserID() + ": " + this.userName + "@" + this.realmID;
     }
 
     @Transient


mycore-user2/src/main/java/org/mycore/user2/MCRUser.java
```diff
@@ -499,7 +499,7 @@ public int hashCode() {
     @Transient
     @Override
     public String toString() {
-        return getUserID();
+        return getUserID() + ": " + this.userName + "@" + this.realmID;
```
toKrause, line 502: Why? 

The method `getUserID` already returns the concatenation of username and realm ID.

Possommi: You are right, I only checked 2023.06.x

PR #2731, opened 2025-11-11
MCR-3443 Solr classification core name should be optional

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3443).


chemicus (approved):
Seems to be a reasonable bugfix. I'll add a small refactoring in a separate commit which may or may not be accepted.

kkrebs (approved):

PR #2735, opened 2025-11-13
MCR-3557 Fix default transformer in XEditor post processor

golsch:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3557).


sebhofmann (approved):

toKrause:
I've tested this PR and can confirm that it solves the problem.

fluetze (changes requested):
I can not approve this.
Yes, it solves the problem, but by changing the transformer used.
We have to question why MCRXSL2XMLTransformer does not the job.
It seems the suclass MCRXSL2XMLTransformer does not behave as its parent class.
Both do a simple XSL transformation here. 
I fear there is  bug hidden in the transformation, which we ignore by "solving" its symptoms.

golsch:
My observation: If the result of a transformation is text THAT looks like XML, the `MCRXSL2XMLTransformer` still treats it as text and cannot create a Document from the result. However, the `MCRXSLTransformer` seems to handle this "correctly". The question is which behavior is actually correct and which one we really want.

PR #2736, opened 2025-11-14
MCR-3391 mycore-junit4 should contain TestCase classes

yagee-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3391).

No classes were present in artifact of `mycore-junit4` before PR.


toKrause:
The module `mycore-junit4` was created by moving the old test-classes from `mycore-base` without any structural modifications. They were therefore kept in `/src/test/java`. But you are right, currently they do not end up in any artifact.

An alternate approach to the structural modifications you are proposing could be to add the following configuration to the `pom.xml` of `mycore-junit4` (also copied from `mycore-base`):

```xml
<plugin>
  <groupId>org.apache.maven.plugins</groupId>
  <artifactId>maven-jar-plugin</artifactId>
  <executions>
    <execution>
      <goals>
        <goal>test-jar</goal>
      </goals>
    </execution>
  </executions>
</plugin>
```

Then, the artifact providing the Junit4-helper-clases for MyCoRe would still be included with `<type>test-jar</type>`, just as before.


I'm not sure which approach is better.

yagee-de:
I see your point that this results in very little change to the code that uses it, but change is already needed. So therefore I suggest that test code should test the artifact. But as there is no code in the artifact the former test code now should be the artifact instead. I hope this makes sense. At this moment the artifact "mycore-junit4" does not contain any code helping with Junit4 legacy tests.

toKrause (approved):

PR #2738, opened 2025-11-18
MCR-3559 Return 404 http status code in MCRUserServlet when user cannot be found

Possommi:
[MCR-3559](https://mycore.atlassian.net/browse/MCR-3559).

[MCR-3559]: https://mycore.atlassian.net/browse/MCR-3559?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mycore-user2/src/main/java/org/mycore/user2/MCRUserServlet.java
+++ b/mycore-user2/src/main/java/org/mycore/user2/MCRUserServlet.java
@@ -135,7 +135,7 @@ public class MCRUserServlet extends MCRServlet {
         if (currentUser == null) {
             String uid = userID == null ? MCRSessionMgr.getCurrentSession().getUserInformation().getUserID() : userID;
             String msg = MCRTranslation.translate("component.user2.UserServlet.currentUserUnknown", uid);
-            res.sendError(HttpServletResponse.SC_FORBIDDEN, msg);
+            res.sendError(HttpServletResponse.SC_NOT_FOUND, msg);
             return false;
         }
         return true;


toKrause:
Is the current behavior (returning the same error code, independent of the reason why a user can't be accessed)  a bug or a deliberate security consideration? The current behavior might be intentional. Otherwise, attackers might be able to use the error codes to determine whether a user account exists or not.

Possommi:
> Is the current behavior (returning the same error code, independent of the reason why a user can't be accessed) a bug or a deliberate security consideration? The current behavior might be intentional. Otherwise, attackers might be able to use the error codes to determine whether a user account exists or not.

I think that was just laziness. It always returned 403. Now 403 is returned if you don't have the right to access the resource. 404 is returned when you have access rights but the requested user cannot not be found.

toKrause:
> Now 403 is returned if you don't have the right to access the resource. 404 is returned when you have access rights but the requested user cannot not be found.

I'm not sure that that is true. In `showUser`, `checkUserIsNotNull` is called first, before any permission check. Therefore, 404 is returned when the requested user can not be found, whether the calling user has access-rights or not.

toKrause:
@yagee-de wrote the code in question 13 years ago. Perhaps he still remembers whether he acted intentionally or was "just lazy"

golsch:
In my opinion, a general error is not appropriate here. I prefer to remove the error from the method and handle it in the respective methods.

yagee-de (approved):

toKrause (approved):

PR #2743, opened 2025-11-21
MCR-3562 Allow to request citation in language specific format

Possommi:
[MCR-3562](https://mycore.atlassian.net/browse/MCR-3562).

[MCR-3562]: https://mycore.atlassian.net/browse/MCR-3562?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mycore-csl/src/main/resources/components/csl/config/mycore.properties
```diff
@@ -1,5 +1,6 @@
 MCR.CSL.NonDroppingParticles=der bei,bei,der
 MCR.CSL.DroppingParticles=van,von,zur,de,zu,de,di,del,dello,della,dei,delle,da,dal,degli,dalla,zur,of,van\u2019t,du,la
+MCR.CSL.defaultLocale = en
```
toKrause: Should be `MCR.CSL.DefaultLocale`.

mycore-csl/src/main/java/org/mycore/csl/MCRCSLTransformer.java
```diff
@@ -75,40 +98,49 @@ public MCRContent transform(MCRContent source) {
...
     }
 
-    private void returnTransformerInstance(MCRCSLTransformerInstance instance, String style, String format) {
+    private void returnTransformerInstance(MCRCSLTransformerInstance instance, String style, String format,
+        Language language) {
         try {
             instance.getCitationProcessor().reset();
             instance.getDataProvider().reset();
             instance.getCitationProcessor().setOutputFormat(format);
+            if ("xml".equals(format)) {
```
toKrause: Why is this part of this ticket? Doesn't seem to have anything to do with the language attribute.

Possommi: Actually this pr makes only sense for me when the csl transformer supports the `MCRCSLXMLOutputFormat` which is also introduced here. I need this format and the language support for UBO to generate solr responses with content generated by the `MCRCSLTransformer`

toKrause: I don't doubt this, but the two aspects handled in this PR could/should have been two separate tickets and PRs.

Possommi: I see where this is going. I removed the references and created an extra ticket [MCR-3588](https://mycore.atlassian.net/browse/MCR-3588)

toKrause (approved):

PR #2764, opened 2025-12-04
MCR-2897 Workaround to allow resource protocol in mycore-fo

erodde:
[Link to jira](https://mycore.atlassian.net/browse/MCR-2897).

Please check if this is an alternative to migrating to Xalan. I'm not sure the XML parser is even the cause of this specific problem.

### DIFF
--- a/mycore-fo/src/main/java/org/mycore/component/fo/common/fo/MCRFoFormatterFOP.java
+++ b/mycore-fo/src/main/java/org/mycore/component/fo/common/fo/MCRFoFormatterFOP.java
@@ -46,6 +46,7 @@ import org.apache.fop.apps.Fop;
 import org.apache.fop.apps.FopFactory;
 import org.apache.fop.apps.FopFactoryBuilder;
 import org.apache.fop.apps.MimeConstants;
+import org.apache.fop.apps.io.InternalResourceResolver;
 import org.apache.fop.configuration.Configuration;
 import org.apache.fop.configuration.ConfigurationException;
 import org.apache.fop.configuration.DefaultConfigurationBuilder;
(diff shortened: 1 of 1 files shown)


michael-becker:
I tested it locally without the fix ad get the mentioned exception. Using the fix, it works for me. A slighty less hacky solution would be to separate the environmental profile, e.g.

```java
final URI baseUri = URI.create("resource:/");
final EnvironmentProfile env = EnvironmentalProfileFactory.createRestrictedIO(baseUri, resolver);
final String foCfg = MCRConfiguration2.getString("MCR.LayoutService.FoFormatter.FOP.config").orElse("");

final FopFactoryBuilder fopFactoryBuilder;
if (StringUtils.isBlank(foCfg)) {
  fopFactoryBuilder = new FopFactoryBuilder(env);
} else {
  fopFactoryBuilder = getBuilderFromConfiguration(foCfg, baseUri, env);
}
```

and

```java
public static FopFactoryBuilder getBuilderFromConfiguration(final String foCfg, URI baseUri,
        EnvironmentProfile env) {
        final FopFactoryBuilder fopFactoryBuilder;
        try {
            URL configResource = MCRResourceHelper.getResourceUrl(foCfg);
            URLConnection con = configResource.openConnection();
            try (InputStream is = con.getInputStream()) {
                FopConfParser parser = new FopConfParser(is, baseUri, env);
                fopFactoryBuilder = parser.getFopFactoryBuilder();
            }

... (hyphenation workaround)

            return fopFactoryBuilder;
        } catch (Exception e) {
            throw new MCRException("Exception while loading FOP configuration from " + foCfg, e);
        }
    }
```

But I think, your solution is also appropriate.

michael-becker (approved):

erodde:
I tried implementing the suggested changes, but i would need to change quite a few things in the MCRFoFormatterFOP. As I'm not firm with apache-fop, I cannot say what behavior would change if we instantiate the `fopFactory` via the `FopConfParser` instead of directly setting the config in the `FopFactoryBuilder`. The hyphenation workaround also depends on the `org.apache.fop.configuration.Configuration` (`cfg` variable), complicating the method rewrite. 

The workaround is suggested similarily in https://issues.apache.org/jira/browse/FOP-3109 and https://github.com/fugerit79/fop-custom-resource-resolver-poc/commit/d2f8ec60b69d160442c16025ba4aff9c99c863a1 .

I would suggest using the workaround for now, and overhaul the MCRFoFormatterFOP in the long term. The current version of apache-fop at the time of writing is 2.11, we are using 2.10.

erodde:
Rework of MCRFoFormatterFOP tracked through https://mycore.atlassian.net/browse/MCR-3591

PR #2765, opened 2025-12-05
MCR-3571 links leading to derivates are not handled properly

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3571).


toKrause:
It's unclear to me why the changes in `MCRPersistentIdentifierEventHandler` are necessary, and the Jira ticket doesn't include a detailed description of the error. What error is fixed by this change, and how can I test it?

toKrause (approved):

PR #2776, opened 2025-12-17
MCR-3581! Improve factory methods in mycore-acl

golsch:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3581).

## Ticket & Documentation

## Testing

## MCR Conventions & Metadata

## Multi-Repo Considerations
  - [x] If yes, is it already created?


mycore-acl/src/main/java/org/mycore/mcr/acl/accesskey/service/MCRAccessKeyServiceProvider.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+
+package org.mycore.mcr.acl.accesskey.service;
+
+import org.mycore.common.config.MCRConfiguration2;
+
+/**
+ * Provides a centralized access point for the shared instance of {@link MCRAccessKeyService}.
+ */
+public final class MCRAccessKeyServiceProvider {
```
toKrause: Would you consider renaming this class `MCRAccessKeyServiceManager`? Even though it currently only _provides_ access to the configured `MCRAccessKeyService`, this class might be a good home for potential future methods that work with `MCRAccessKeyService` instances.

It would also more closely follow existing naming conventions established in MyCoRe, for example: `MCRPIServiceManager`, `MCRProcessableManager`, `MCRTikaManager`.

toKrause: Or, to avoid holding a singleton instance in a class other then the singletons class altogether (which is unusual from a design-pattern perspective and afaik. currently done nowhere else in MyCoRe), remove this class and put the singleton holder into the interface, see `MCRObjectQueryResolver`.

(Even better would be not having polymorph singletons, but I think that ship has sailed in MyCoRe a long time ago.)

golsch: I have decided to simply keep the instance directly in `AccessKeyServiceImpl` after all. An independent approach would be nice to have, but I don't really expect further implementations in the future and therefore don't want to make it unnecessarily complex.

toKrause: But  `MCRAccessKeyService` is an interface an not a class, so we should expect other implementations (say, a mock-implementation for testing). At least, other implementations should be possible.

And I don't see how using a different implementations is supposed to work with the current version of the PR. 

The property `MCR.ACL.AccessKey.Service.Class` is evaluated in `MCRAccessKeyServiceImpl` with 
`MCRConfiguration2.getInstanceOfOrThrow(MCRAccessKeyServiceImpl.class, INSTANCE_PROPERTY);`. That means, that all implementations have to be a subclass of `MCRAccessKeyServiceImpl`. Other implementations are impossible.

Also, wherever the `MCRAccessKeyService` is used, it is now acquired with `MCRAccessKeyServiceImpl.obtainInstance()`, which tightly couples the usage of an interface to one specific implementation.

The example in `MCRObjectQueryResolver` is certainly unusual and a bit ugly, but it doesn't introduce these problems.

golsch: You're right. But I don't like the approach via the interface either. May we should define a general approach for this problem. Something like a method that provides the concrete implementation via the `mycore.properties` using an interface / class as paramter, for example `MCRConfiguration2#getSingleInstance(Class<T>)`.
Then the problem would be solved once and for all. What do you mean? I would make a specific proposal.

toKrause: Another approach could be to turn `MCRAccessKeyService` into a final class and use the adapter pattern pioneered by `MCRXMLMetadataManager`.

golsch: I have now switched to the approach via the interface.

toKrause (approved):

PR #2830, opened 2026-02-26
MCR-3625 fix mcrversion XSLT3 function and URI-Resolver

rsteph-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3625).

## Ticket & Documentation

- [x] For UI changes: before & after screenshots are attached.

## Bugfix-Specific Checks

## Testing
- x] The feature behaves as described in the ticket.

## MCR Conventions & Metadata

## Multi-Repo Considerations

### DIFF
--- a/mycore-base/src/main/java/org/mycore/common/xml/MCRURIResolver.java
+++ b/mycore-base/src/main/java/org/mycore/common/xml/MCRURIResolver.java
@@ -1455,7 +1455,9 @@ public final class MCRURIResolver implements URIResolver {
                     case "version" -> MCRCoreVersion.getVersion();
                     case "completeVersion" -> MCRCoreVersion.getCompleteVersion();
                     case "revision" -> MCRCoreVersion.getRevision();
-                    default -> MCRCoreVersion.getCompleteVersion();
+                    default -> throw new IllegalArgumentException(
+                        "Invalid parameter for MCRVersionResolver: " + versionType);
+
                 });
             return new JDOMSource(versionElement);
         }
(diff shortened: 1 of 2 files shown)


rsteph-de:
There is no harm in renaming the XSLT3 function, with the code missing in MCRVersionURIResolver it has never been working before.

mycore-base/src/main/java/org/mycore/common/xml/MCRURIResolver.java
```diff
@@ -1453,6 +1453,7 @@ public Source resolve(String href, String base) throws TransformerException {
                     case "abbrev" -> MCRCoreVersion.getAbbrev();
                     case "branch" -> MCRCoreVersion.getBranch();
                     case "version" -> MCRCoreVersion.getVersion();
+                    case "completeVersion" -> MCRCoreVersion.getCompleteVersion();
```
golsch, line 1456: The URIResolver definitely worked because the `default` case is `MCRCoreVersion.getCompleteVersion()`. `document('version')` therefore returns the "complete version." After your change, you get it explicitly or implicitly as the default, which I don't think is very good. Accordingly, an exception should now be thrown by default, but that would be an API change.

golsch (changes requested):

mycore-base/src/main/java/org/mycore/common/xml/MCRURIResolver.java
rsteph-de: Good point, I go for the exception and accept the API change. 
The MCRVersion-URIResolver is now internal and should only be used by the corresponding XSLT3-function

yagee-de: The way it is implemented is backward compatible and should not cause any problems. Also pairing is not possible, e.g. `case "completeVersion", default -> …`. So I don't see an issue here. It is perfectly valid that two cases causes the same result.

rsteph-de: as described above, backward compatiblity is a minor issue, because the XSLT3 function should be used. I prefer the cleaner code.

golsch (approved):

PR #2857, opened 2026-03-13
MCR-3641 MyCoReWebPage.xsl does not support resolving of i18n values in attributes

Possommi:
[MCR-3641](https://mycore.atlassian.net/browse/MCR-3641).

## Ticket & Documentation

## Bugfix-Specific Checks

## Testing

## MCR Conventions & Metadata

## Multi-Repo Considerations

[MCR-3641]: https://mycore.atlassian.net/browse/MCR-3641?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mycore-base/src/main/resources/xsl/MyCoReWebPage.xsl
+++ b/mycore-base/src/main/resources/xsl/MyCoReWebPage.xsl
@@ -145,6 +145,16 @@
     </xsl:choose>
   </xsl:template>
 
+  <xsl:template match="@*[contains(. ,'{i18n:')][contains(. ,'}')]">
+    <xsl:variable name="before-key" select="substring-before(. ,'{i18n:')"/>
+    <xsl:variable name="after-key" select="substring-after(. ,'}')"/>
+    <xsl:variable name="key" select="substring-before(substring-after(. ,'{i18n:'), '}')"/>
+
+    <xsl:attribute name="{local-name()}">
+      <xsl:value-of select="concat($before-key, i18n:translate($key), $after-key)" disable-output-escaping="yes"/>
+    </xsl:attribute>
+  </xsl:template>
+
   <!-- =============================================================================== -->
 
   <!-- Dynamic includes -->
--- a/mycore-base/src/main/resources/xslt/MyCoReWebPage.xsl
+++ b/mycore-base/src/main/resources/xslt/MyCoReWebPage.xsl
@@ -144,6 +144,16 @@
     <xsl:value-of select="mcri18n:translate(@key)" />
   </xsl:template>
 
+  <xsl:template match="@*[contains(. ,'{i18n:')][contains(. ,'}')]">
+    <xsl:variable name="before-key" select="substring-before(. ,'{i18n:')"/>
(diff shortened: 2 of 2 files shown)


fluetze:
Actually, we already have a solution for this for years. It is called i18n-transformation-step.xsl in mycore-base and heavily used in DuEPublico. For example, see https://github.com/ub-duisburg-essen/duepublico/blob/main/duepublico-extras/src/main/resources/META-INF/resources/content/index.xml

Possommi:
> Actually, we already have a solution for this for years. It is called i18n-transformation-step.xsl in mycore-base and heavily used in DuEPublico. For example, see https://github.com/ub-duisburg-essen/duepublico/blob/main/duepublico-extras/src/main/resources/META-INF/resources/content/index.xml

so `|code:mir.img.preview|` would work for resolving in attributes, i guess. If I don't need to provide an extra stylesheet I am more than happy.

toKrause (commented):
Does this only support a single `{i18n:...}`-replacement in a given attribute, or am I missing something? 

I think, multiple replacements should be supported for this to be a well-rounded feature.

toKrause:
> i18n-transformation-step.xsl

This stylesheet is not part of `mycore-base` anymore. It has been moved to the legacy module.

fluetze:
> > i18n-transformation-step.xsl
> This stylesheet is not part of `mycore-base` anymore. It has been moved to the legacy module.

Ok, then it is a good idea to reinvent the functionality again. ;-)

From my point of view, i18n is best implemented as a separate transformation step. That way it will work with static pages, xeditors etc. - If nobody wants to do it that way, I can move the xsl to DuEPublico.

Possommi:
> > > i18n-transformation-step.xsl
> > > This stylesheet is not part of `mycore-base` anymore. It has been moved to the legacy module.
> 
> Ok, then it is a good idea to reinvent the functionality again. ;-)
> 
> From my point of view, i18n is best implemented as a separate transformation step. That way it will work with static pages, xeditors etc. - If nobody wants to do it that way, I can move the xsl to DuEPublico.

in 2025.12.x it is still part, so I can use it

toKrause:
I didn't remove that stylesheet. Just noticed, that it doesn't exist (anymore).

But, ...

Independently reusable features do not fit well with the MyCoRe core design principles ("Do repeat yourself", "Functionality by obscurity", ...), so getting rid of that stylesheet was probably the correct decision.

:)

Possommi:
> Does this only support a single `{i18n:...}`-replacement in a given attribute, or am I missing something?
> 
> I think, multiple replacements should be supported for this to be a well-rounded feature.

Although I get your point I do not need more at the moment and won't implement things I don't need right now. If this pr cannot get into mcr I will just have this feature in my application module, which would be ok for me.

toKrause (changes requested):
I think supporting only up to one replacement per attribute would be too much of a pitfall for this to be a well-rounded feature.

Possommi:
For now I will have it in my application module. I might come back to this when migrating away from xalan.

PR #2861, opened 2026-03-17
MCR-3646 Harmonize edge cases and improve error and log messages of annotation based configuration

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3646).

## Ticket & Documentation

Behavior of some edge cases changes. See ticket for complete list. Migration guide follows.

## Testing

See ticket. Code contains exhaustive list of tests cases.

## MCR Conventions & Metadata

## Multi-Repo Considerations


sebhofmann (changes requested):
The 9600 lines across these six files aren't really maintainable. They're copies of the same structure, walking the truth table cell by cell. When the rules need to change, nobody's going to want to touch this.

The spec is ~10 rules plus 3 log triggers, driven by the same four input axes. Could we encode those rules once in the test as expectedOutcome(required, classProp, sentinel, valueProp) returning the expected instance and logs, then loop over all input combinations? A rule change is one edit, not 540. The error message structure (type, property, class, field, reason) can be a single helper applied at each exception site.

If that feels too much, the lighter version is `@ParameterizedTest` walking the same combinations. That needs the test setup to move off `@MCRTestConfiguration` / `@MCRTestProperty` though, since annotations can't take runtime parameters. Properties would be set programmatically in `@BeforeEach` (or per case) from the arguments.

toKrause:
I'm not sure how practical parameterization would be, since different test cases use differently annotated configured classes, and I didn't want to develop a complete testing framework. It was much simpler to test each case individually than to invent something clever.

toKrause:
... but I'll look into it.

toKrause:
I've reworked one class as an example: https://github.com/MyCoRe-Org/mycore/pull/2861/changes#diff-8797c2630ef7a30469903ddb89e83200e10c8891a11c7b85c47ae6e5b2e7a581

toKrause:
If this style is okay for you, I'll rework the rest accordingly.

sebhofmann:
Yes that style is okay for me. Just notices some typos shoud -> should.

mycore-base/src/main/java/org/mycore/common/config/MCRConfigurableInstanceHelper.java
```diff
@@ -229,19 +234,96 @@ private static String targetTypeName(Target<?> target) {
...
+        if (Objects.equals("", annotationName)) {
+            return configuration.name().canonical();
+        } else {
+            return configuration.name().canonical() + "." + annotationName;
+        }
+    }
+
+    private static MCRConfigurationException incompatibilityException(String property, Target<?> target,
+        Class<?> annotationValueClass, Object instance) {
+        throw new MCRConfigurationException("Instance of class " + instance.getClass().getName()
```
sebhofmann: This helper method should return the Exception or have void as return type. I is also inconsistent compared to `nonIntegerKeyException` and `inconsistentKeysException`

mycore-base/src/main/java/org/mycore/common/config/MCRConfigurableInstanceHelper.java
```diff
@@ -229,19 +234,96 @@ private static String targetTypeName(Target<?> target) {
...
+            + "', configured in " + property + ", is incompatible with annotation value class "
+            + annotationValueClass.getName() + " for target " + targetTypeName(target)
+            + " '" + target.name() + "' in configured class " + target.declaringClass().getName());
+    }
+
+    private static MCRConfigurationException missingException(String property, Target<?> target,
+        String description) {
+        throw new MCRConfigurationException(
```
sebhofmann: This helper method should return the Exception or have void as return type. I is also inconsistent compared to `nonIntegerKeyException` and `inconsistentKeysException`

mycore-base/src/main/java/org/mycore/common/config/MCRConfigurableInstanceHelper.java
```diff
@@ -229,19 +234,96 @@ private static String targetTypeName(Target<?> target) {
...
+            capitalize(description) + ", configured in " + property + " (and its sub-properties)," +
+                " for target " + targetTypeName(target) + " '" + target.name() + "' in configured class "
+                + target.declaringClass().getName()
+                + " is missing");
+    }
+
+    private static MCRConfigurationException emptyException(String property, Target<?> target, String description) {
+        throw new MCRConfigurationException(
```
sebhofmann: This helper method should return the Exception or have void as return type. I is also inconsistent compared to `nonIntegerKeyException` and `inconsistentKeysException`

mycore-base/src/main/java/org/mycore/common/config/MCRConfigurableInstanceHelper.java
```diff
@@ -229,19 +234,96 @@ private static String targetTypeName(Target<?> target) {
...
+            capitalize(description) + ", configured in " + property + " (and its sub-properties)," +
+                " for target " + targetTypeName(target) + " '" + target.name() + "' in configured class "
+                + target.declaringClass().getName()
+                + " is empty");
+    }
+
+    private static MCRConfigurationException emptyRawException(String property, Target<?> target, String description) {
+        throw new MCRConfigurationException(
```
sebhofmann: This helper method should return the Exception or have void as return type. I is also inconsistent compared to `nonIntegerKeyException` and `inconsistentKeysException`

mycore-base/src/main/java/org/mycore/common/config/MCRConfigurableInstanceHelper.java
```diff
@@ -1299,7 +1449,59 @@ public Map<String, String> get(MCRInstanceConfiguration configuration, Target<?>
...
+
+        protected Object getInstance(String property, Target<?> target, Class<?> valueClass,
+            MCRInstanceConfiguration nestedConfiguration, MCRSentinel sentinel) {
+
+            Set<Option> options = createOptions(valueClass);
+            boolean implicitValueClass = Modifier.isFinal(valueClass.getModifiers());
+
+            if (sentinel.enabled()) {
+                boolean sentinelValue = sentinel.defaultValue();
+                String configuredSentinelValue = nestedConfiguration.properties().remove(sentinel.name());
```
sebhofmann, line 1463: This modifies the the `nestedConfiguration.properties()`, does this get evaluated just once? Maybe better use `properties().get(sentinel.name())`.

mycore-base/src/main/java/org/mycore/common/config/MCRConfigurableInstanceHelper.java
```diff
@@ -229,19 +234,96 @@ private static String targetTypeName(Target<?> target) {
...
+                + target.declaringClass().getName()
+                + " is empty");
+    }
+
+    private static MCRConfigurationException nonIntegerKeyException(String property, Target<?> target,
+        String key, String description, NumberFormatException exception) {
+        return new MCRConfigurationException(
+            capitalize(description) + ", configured in " + property + " (and its sub-properties)," +
+                " for target " + targetTypeName(target) + " '" + target.name() + "' in configured class "
+                + target.declaringClass().getName() + " is has element with non-integer key " + key,
```
sebhofmann: Typo: "is has element with non-integer key" -> "has element with non-integer key".

mycore-base/src/main/java/org/mycore/common/config/MCRConfigurableInstanceHelper.java
```diff
@@ -901,29 +983,70 @@ public String get(MCRInstanceConfiguration configuration, Target<?> target) {
...
+
+            if (sentinel.enabled()) {
+                boolean sentinelValue = sentinel.defaultValue();
+                String configurationValue = properties.get(prefix + "." + sentinel.name());
+                if (configurationValue != null) {
+                    sentinelValue = Boolean.parseBoolean(configurationValue);
+                }
+                if (sentinelValue == sentinel.rejectionValue()) {
+                    if (LOGGER.isInfoEnabled()) {
+                        LOGGER.info("[SENTINEL] Ignoring {} {} and all sup-properties",
```
sebhofmann: Typo: sub-properties

mycore-base/src/main/java/org/mycore/common/config/MCRConfigurableInstanceHelper.java
```diff
@@ -169,25 +167,32 @@ public static <S> S getInstance(Class<S> superClass, MCRInstanceConfiguration co
...
         }
         Class<S> targetClass = getClass(configuration.name().actual(), configuration.className());
-        Object instance = createInstanceDirectorViaProxy(targetClass, configuration);
+        Object instance = createInstanceDirectOrViaProxy(targetClass, configuration);
         if (superClass.isAssignableFrom(instance.getClass())) {
             return superClass.cast(instance);
         } else {
-            throw new MCRConfigurationException("Configured instance of class " + instance.getClass().getName()
+            throw new MCRConfigurationException("Instance of class " + instance.getClass().getName()
                 + " is incompatible with intended super class " + superClass.getName() + "'"
```
sebhofmann: Missing opening `'`

mycore-base/src/main/java/org/mycore/common/config/MCRConfigurableInstanceHelper.java
sebhofmann: Missing opening `'`

sebhofmann (changes requested):

mycore-base/src/main/java/org/mycore/common/config/MCRConfigurableInstanceHelper.java
toKrause: The configuration is only used for one instantiation process. it is not reused and can therefore safely be modified. The idea here is to remove the sentinel value, so it isn't available anymore as a configurable value (same as with the `Class` property, which is also not part of `properties()`).

This is intended to prevent a situation where one property influences two things. Example:

```java
    @MCRInstance(name = FOOBAR_KEY, valueClass = FooBar.class, required = false, sentinel = @MCRSentinel(name = "Active"))
    public Foobar foobar;
    
    public static class Foobar {
        
        @MCRProperty(name = "Class")
        private String classString;
        
        @MCRProperty(name = "Active")
        private String activeString;
        
    }
```

Both of these properties should not be used as a configuration value for the instance, because they are already used as as a configuration value for the instantiation process.

sebhofmann (approved):

PR #2865, opened 2026-03-17
MCR-3645 refactor solr integration to use configurable class instances mcrsolrindexmanager

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3645).

## Ticket & Documentation

## Testing

## MCR Conventions & Metadata

## Multi-Repo Considerations
  - [x] If yes, is it already created? see -> https://github.com/MyCoRe-Org/mir/pull/1308


mycore-solr/src/main/java/org/mycore/solr/MCRSolrIndexManager.java
```diff
@@ -0,0 +1,73 @@
...
+
+package org.mycore.solr;
+
+import java.util.List;
+import java.util.Optional;
+
+import org.mycore.common.config.MCRConfiguration2;
+import org.mycore.common.config.MCRConfigurationException;
+
+public interface MCRSolrIndexManager {
```
Mewel: * the "Manager" part is a bit missleading for an interface. Would probably prefer Service or something.
* should have javadoc
* This class does multiple things and should be splitted into seperate ones:
  * a manager with the obtainInstance() stuff
  * a interface
* also this class reads from SOLR_COLLECTION_MANAGER_PROPERTY, getMainIndex() resolves via MCRSolrConstants.MAIN_CORE_TYPE, so within this one interface: "Index" (class name), "Collection" (constant), and "Core" (constant) are all used for the same concept

mycore-solr/src/main/java/org/mycore/solr/MCRSolrIndex.java
```diff
@@ -0,0 +1,73 @@
...
+import java.util.Optional;
+import java.util.Set;
+
+import org.apache.solr.client.solrj.SolrClient;
+import org.apache.solr.client.solrj.impl.ConcurrentUpdateBaseSolrClient;
+
+/**
+ * Describes a Collection or Core on a SolrServer
+ */
+public interface MCRSolrIndex {
```
Mewel, line 29: The name confuses me, cause it doesn't really say what it does. Maybe something like MCRSolrEndpoint would be better.

mycore-solr/src/main/java/org/mycore/solr/MCRConfigurableIndexManager.java
```diff
@@ -0,0 +1,93 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.mycore.common.config.annotation.MCRConfigurationProxy;
+import org.mycore.common.config.annotation.MCRInstanceMap;
+import org.mycore.common.events.MCRShutdownHandler;
+import org.mycore.solr.MCRConfigurableIndexManager.ConfigAdapter;
+
+@MCRConfigurationProxy(
+    proxyClass = ConfigAdapter.class)
+public class MCRConfigurableIndexManager implements MCRSolrIndexManager {
```
Mewel: The naming is confusing again. The "Configurable" prefix describes how the class is wired up (via @MCRConfigurationProxy). There's no       
  non-configurable counterpart, so the prefix doesn't distinguish anything meaningful. It's an implementation detail leaking into the class name.

From my understanding this is the standalone implementation, because you also have a MCRModifiableSolrIndexManager (i guess for the cloud). Probably the name should reflect that.

So better would be:
MCRSolrStandaloneIndexService
MCRSolrCloudIndexService

or will the cloud version also extending the MCRConfigurableIndexManager? Cause there is no implementation of MCRModifiableSolrIndexManager yet. Very confusing.

Also missing javadoc

mycore-solr/src/main/java/org/mycore/solr/cloud/collection/MCRConfigurableSolrCloudCollection.java
```diff
@@ -0,0 +1,296 @@
...
+import org.mycore.common.config.annotation.MCRProperty;
+import org.mycore.solr.MCRAbstractHttpBasedIndexConfigAdapter;
+import org.mycore.solr.MCRIndexType;
+import org.mycore.solr.MCRSolrUtils;
+import org.mycore.solr.auth.MCRSolrAuthenticationLevel;
+import org.mycore.solr.auth.MCRSolrAuthenticationManager;
+import org.mycore.solr.cloud.collection.MCRConfigurableSolrCloudCollection.ConfigAdapter;
+
+@MCRConfigurationProxy(proxyClass = ConfigAdapter.class)
+public class MCRConfigurableSolrCloudCollection implements MCRSolrCloudCollection {
```
Mewel, line 47: Some thoughts about the hierarchy:
MCRSolrIndex (interface)                                                                                                                                                              
  ├── MCRConfigurableSolrCore          ← directly implements MCRSolrIndex, no sub-interface
  └── MCRSolrCloudCollection           ← sub-interface of MCRSolrIndex                                                                                                                  
      └── MCRConfigurableSolrCloudCollection

This seem inconsistent. The cloud version has an extra interface while the standalone version doesnt. Also the "Configurable" part again is not required.

I would rm the MCRSolrCloudCollection interface and just do two implementations:
* MCRSolrCore
* MCRSolrCloudCollection

mycore-solr/src/main/java/org/mycore/solr/cloud/collection/MCRConfigurableSolrCloudCollection.java
```diff
@@ -0,0 +1,296 @@
...
+    }
+
+    @Override
+    public Optional<ConcurrentUpdateBaseSolrClient> getConcurrentClient() {
+        // You should use the normal Cloud Client, when using SolrCloud
+        return Optional.empty();
+    }
+
+    @Override
+    public Set<MCRIndexType> getCoreTypes() {
```
Mewel: getCoreTypes() is defined on MCRSolrIndex and inherited by MCRConfigurableSolrCloudCollection. "Core" is standalone terminology — calling it on a cloud collection is a terminology mismatch. The field in the cloud impl is even named coreTypes. Something like getIndexTypes() would be neutral across both modes.

I just saw, the return type is also MCRIndexType. So the name should reflect that.

mycore-solr/src/main/java/org/mycore/solr/MCRAbstractHttpBasedIndexConfigAdapter.java
```diff
@@ -0,0 +1,188 @@
...
+    @MCRProperty(name = "UseJettyHttpClient", defaultName = MCRSolrDefaultPropertyConstants.USE_JETTY_HTTP_CLIENT)
+    public void setUseJettyHttpClient(String useJettyHttpClient) {
+        this.useJettyHttpClient = Boolean.parseBoolean(useJettyHttpClient);
+    }
+
+    public boolean useJettyHttpClient() {
+        return useJettyHttpClient;
+    }
+
+    protected HttpSolrClientBuilderBase getBuilder() {
```
Mewel: Raw use of parameterized class 'HttpSolrClientBuilderBase'. Should be HttpSolrClientBuilderBase<?, ?>.
This is just an example, this should be fixed in the whole code base.

Mewel (changes requested):

mycore-solr/src/main/java/org/mycore/solr/MCRSolrIndex.java
sebhofmann: I also considered `MCRSolrEndpoint` but i think that would be more confusing, since the MCRSolrIndex describes a collection or a core and `MCRSolrEndpoint` sounds like just a solr server and nothing more.

mycore-solr/src/main/java/org/mycore/solr/cloud/collection/MCRConfigurableSolrCloudCollection.java
sebhofmann: This is not inconsistent.
The interfaces which is used in mycore is MCRSolrIndex. 

 The `MCRSolrCloudCollection` adds addional generic parameters which are required to create a Index in a solr cloud. Creating indexes is only possible in the Cloud, so it does not make sense to have that parameters in `MCRSolrIndex`.

The `MCRConfigurableIndexManager`, `MCRConfigurableSolrCore` and `MCRConfigurableSolrCloudCollection` are the actual implementations which use `MCRConfiguration`.

There could be another implementation which only allows to add Cores by using CLI Commands.

mycore-solr/src/main/java/org/mycore/solr/cloud/collection/MCRConfigurableSolrCloudCollection.java
sebhofmann: I fixed that naming inconsistency by renaming the coreTypes to IndexTypes

mycore-solr/src/main/java/org/mycore/solr/MCRConfigurableIndexManager.java
sebhofmann: I added some Javadocs to the `MCRSolrIndexManager`. The Configurable part is the clue of this implementation. The Indexes are provided by MCRConfiguration. I also removed the `MCRModifiableSolrIndexManager` interface, since it is not used anymore, since i decided to not allow the configuration by CLI-Commands.

mycore-mods/src/main/java/org/mycore/mods/MCRMODSEmbargoReleaseCronjob.java
```diff
@@ -58,7 +57,7 @@ public String getDescription() {
     @Override
     public void runJob() {
 
-        if (MCRConfiguration2.getString("MCR.Solr.ServerURL").isEmpty()) {
+        if(MCRSolrIndexManager.obtainInstance().getMainIndex().isPresent()) {
```
Mewel: This should be the other way around right?

mycore-solr/src/main/java/org/mycore/solr/proxy/MCRSolrProxyServlet.java
```diff
@@ -230,55 +221,52 @@ private void handleQuery(String queryHandlerPath, HttpServletRequest request, Ht
...
-                    OutputStream servletOutput = resp.getOutputStream();
-                    solrResponseStream.transferTo(servletOutput);
-                } else {
-                    MCRStreamContent solrResponse = new MCRStreamContent(solrResponseStream,
-                        solrRequest.uri().toString(),
-                        "response");
-                    MCRLayoutService.obtainInstance().doLayout(request, resp, solrResponse);
-                }
+            NamedList<Object> solrResponse = solrIndex.getClient().request(queryRequest);
+            InputStream is = (InputStream) solrResponse.get("stream");
```
Mewel: Not sure if the stream needs to be closed here with try-with-resource:
try (InputStream is = (InputStream) solrResponse.get("stream")) {

mycore-solr/src/main/resources/components/solr/config/deprecated.properties
```diff
@@ -26,3 +26,18 @@ MCR.Module-solr.JoinQueryFields=MCR.Solr.JoinQueryFields
...
+
+MCR.Solr.SolrClient.ConnectionTimeout=MCR.Solr.Default.Client.ConnectionTimeout
+MCR.Solr.SolrClient.SocketTimeout=MCR.Solr.Default.Client.RequestTimeout && MCR.Solr.Default.Client.IdleTimeout
+
+MCR.Solr.Core.main.ConfigSetTemplate=MCR.Solr.IndexManager.Index.main.ConfigSetTemplate
+MCR.Solr.Core.main.Type=MCR.Solr.IndexManager.Index.main.CoreTypes
+
+MCR.Solr.Core.classification.ConfigSetTemplate=MCR.Solr.IndexManager.Index.classification.ConfigSetTemplate
+MCR.Solr.Core.classification.Type=MCR.Solr.IndexManager.Index.classification.CoreTypes
+
```
Mewel, line 52: Missing deprecated properties:
MCR.Solr.Core.main.Name=jportal-main
MCR.Solr.Core.classification.Name=jportal-classification

Side note: I also found that those properties defined in the solr-xeditor-templates.xsl as xsl:param:
  <xsl:param name="MCR.Solr.ServerURL" />
  <xsl:param name="MCR.Solr.Core.main.Name" />
  <xsl:param name="MCR.Solr.Core.main.ServerURL" />
  <xsl:param name="MCR.Solr.Core.classification.Name" />
  <xsl:param name="MCR.Solr.Core.classification.ServerURL" />

Mewel (changes requested):

mycore-solr/src/main/java/org/mycore/solr/MCRSolrIndexRegistryManager.java
```diff
@@ -0,0 +1,84 @@
...
+    public static void reloadRegistry() {
+        synchronized (MCRSolrIndexRegistryManager.class) {
+            if (instance != null) {
+                instance.closeIndexes();
+            }
+            instance = null;
+            obtainRegistry();
+        }
+    }
+
```
Mewel, line 80: Sorry to open this again. Maybe we can add convenient methods. It seems like we use this alot:
```
public static MCRRegistryIndex requireMainIndex() {
  return MCRSolrIndexRegistryManager.obtainRegistry().requireMainIndex()
}
```

or maybe even getMainClient() and getClassificationClient()

Mewel (changes requested):

Mewel (approved):

PR #2868, opened 2026-03-18
MCR-3649 support more parameter types via casting

yagee-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3649).

documented in ticket


rsteph-de (approved):
good idea!

PR #2880, opened 2026-03-25
MCR-3657 Add new module mycore-solr-api for Solr proxy REST API

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3657).

## Ticket & Documentation

## Testing

## MCR Conventions & Metadata
- [] If the **public API** has changed:

## Multi-Repo Considerations


yagee-de:
What is the benefit you want to achieve?
Here some findings:

- You check permission against QueryHandler path, but qt query parameter has higher priority for SOLRJ
- You do not support POST for larger queries
- You buffer SOLR response in memory instead of streaming the result to the client, this costs performance and memory
- the API differs from SOLRs API, if it's really needed it should be a small wrapper around a request to the SOLR server like the SolrProxyServlet but without relying on a session (cookie)
- SOLR API 'solr/{core}/select' vs. 'api/solr/select?core={core}' should also be reconsidered.
- Support for MCRErrorResponse and MCRExceptionMapper may also provide benefit
- How can I discover cores or query handler?
- How is authentication handled?
- Authorization should be an aspect and not be part of the implementation

sebhofmann:
> What is the benefit you want to achieve?

I need a alternative to `SolrProxyServlet` and `api/v1/solr`, which works with the rest authentication and does the access checking like `SolrProxyServlet`. 

> You do not support POST for larger queries

Not a requirement for me now. Post is usually used to write/persist.

> You buffer SOLR response in memory instead of streaming the result to the client, this costs performance and memory

Valid point, ill change that.


We can discuss the other points in the next meeting.

rsteph-de (approved):
looks good for me

PR #2897, opened 2026-04-02
MCR-3664 Singleton detection in getInstanceOf checks superClass instead of configured class

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3664).

## Ticket & Documentation

## Bugfix-Specific Checks

## Testing
- [ ] I have tested the changes locally.
- [ ] The feature behaves as described in the ticket.

## MCR Conventions & Metadata

## Multi-Repo Considerations

### DIFF
--- a/mycore-base/src/main/java/org/mycore/common/config/MCRConfiguration2.java
+++ b/mycore-base/src/main/java/org/mycore/common/config/MCRConfiguration2.java
@@ -197,7 +197,7 @@ public class MCRConfiguration2 {
      * @throws MCRConfigurationException if the class can not be loaded or instantiated
      */
     public static <S> Optional<S> getInstanceOf(Class<S> superClass, String name) throws MCRConfigurationException {
-        if (MCRConfigurableInstanceHelper.isSingleton(superClass)) {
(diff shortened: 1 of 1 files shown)


sebhofmann (approved):
Fix is correct and minimal — restores the original behavior by resolving the `@Singleton` annotation from the actually configured class (via property name) instead of the passed-in super class.

A few notes:

1. **Merge flow**: The same bug exists in `2025.06.x` and `main`, where the private `getInstanceOf` method has an additional `Set<Option>` parameter. The fix will need to be adapted during the normal merge flow (2024.06.x → 2025.06.x → main) — expect a compile error due to the different method signature.

2. **Test**: It would be nice to add a test in `main` (JUnit 5) that verifies the singleton behavior works correctly when using `getInstanceOf(SuperClass.class, propertyName)` with a `@Singleton`-annotated implementation class — to prevent future regressions.

3. **Checklist**: "fixed-version is correctly set" is checked, but MCR-3664 has no fix version set yet. Should be updated to match the target branch `2024.06.x`.

rsteph-de (approved):
OK

PR #2907, opened 2026-04-10
MCR-3676 XSLT rendering failures caused by invalid MCRObjects

Mewel:
* use MCRXMLMetadataManager to retrieve content directly from hard drive if validation or expanding fails

[Link to jira](https://mycore.atlassian.net/browse/MCR-3676).

## Ticket & Documentation

## Bugfix-Specific Checks

## Testing

## MCR Conventions & Metadata

## Multi-Repo Considerations

### DIFF
--- a/mycore-base/src/main/java/org/mycore/common/xml/MCRURIResolver.java
+++ b/mycore-base/src/main/java/org/mycore/common/xml/MCRURIResolver.java
@@ -178,7 +178,7 @@ public final class MCRURIResolver implements URIResolver {
     public static MCRURIResolver obtainInstance() {
         return SHARED_INSTANCE;
     }
-    
+
     public static Map<String, String> getParameterMap(String key) {
         String[] param;
         StringTokenizer tok = new StringTokenizer(key, "&");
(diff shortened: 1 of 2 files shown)


sebhofmann (changes requested):
The implementation should not fall back to the basic xml, i think thats bad api design. The caller can use the notNull or the exception resolver to prevent rendering failures.

Mewel:
> The implementation should not fall back to the basic xml, i think thats bad api design. The caller can use the notNull or the exception resolver to prevent rendering failures.

I think your right, its not the best design, but then we should add this at least to the migration guide. Because "mcrobject:" resolver now behaves differently. You would also need to add this to every "mcrobject:" call.
And even if you add this (catchEx:mcrobject), you're getting some exception xml stacktrace back instead of an mcrobject which maybe destroys your xsl transformation after.

I would probably prefer another URI resolver which resolves the object from disk. Like mcrobjectfallback: or something.

---
What about the MCRObjectServlet. You can't use XSL.Style=xml to see the xml object. Even if you add the expanded=false parameter you still just see an exception.

sebhofmann:
> Whats the exception in this case? It should work, because it uses MCRbaseContent and not MCRExpandedObjectManager.

You're right, it looks actually ok with expanded = false

> Catching the exception is only fighting the symptoms, you´ll need to fix the object anyway.

Can't really do that with 2.5 million objects. It should be fault tolerant. I spoke with Thomas about it, he said it would be best if mcrobject: reads the expanded object directly from disk. There should be no API inbetween. This would have two advantages. 1. its faster, 2. its like to old API

PR #2922, opened 2026-04-21
MCR-3668 Remove MCRConfigurableInstanceHelper, move value-class-determination into MCRInstanceConfiguration

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3668).

# Pull Request Checklist (Author)

Please go through the following checklist before assigning the PR for review:

## Ticket & Documentation
- [x] The issue in the ticket is clearly described and the solution is documented.
- [x] Design decisions (if any) are explained.
- [x] The ticket references the correct source and target branches.
- [x] The `fixed-version` is correctly set in the ticket and matches the PR's target branch (`main`).

## Feature & Improvement Specific Checks
- [ ] Instructions on how to test or use the feature are included or linked (e.g. to documentation).
- [ ] For UI changes: before & after screenshots are attached.
- [ ] New features or migrations are documented.
- [ ] Does this change affect existing applications, data, or configurations?
  - [ ] Yes: Is a migration required? If yes, describe it.
  - [ ] Breaking change is marked in the **commit message**.

## Testing
- [x] I have tested the changes locally.
- [x] The feature behaves as described in the ticket.
- [x] Were existing tests modified?
  - [x] Yes: explain the changes for reviewers: API changes

## MCR Conventions & Metadata
- [ ] [MCR naming conventions](https://www.mycore.de/documentation/developer/conventions/) are followed
- [ ] If the **public API** has changed:
  - [ ] Old API is deprecated or a migration is documented.
  - [ ] If not, no action needed.
- [ ] Java license headers are added where necessary.
- [ ] Javadoc is written for non-self-explanatory classes/methods (Clean Code).
- [ ] All configuration options are documented in Javadoc and `mycore.properties`.
- [ ] No default properties are hardcoded — all set via `mycore.properties`.

## Multi-Repo Considerations
- [ ] Is an equivalent PR in `MIR` required?
  - [ ] If yes, is it already created?

PR #2924, opened 2026-04-21
MCR-3670 Remove instance name suffixes other than .Class

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3670).

## Ticket & Documentation

## Testing

## MCR Conventions & Metadata

## Multi-Repo Considerations


sebhofmann (changes requested):
### Code review

The core `MCRInstanceName` / `MCRInstanceConfiguration` / `MCRConfiguration2` refactor and the bulk of the property renames are consistent. However, a few module-specific refactors introduce runtime breakage. Requesting changes for the following:

**1. Solr index registry can no longer be instantiated**

`SOLR_INDEX_REGISTRY_PROPERTY` was changed from `SOLR_INDEX_REGISTRY_PROPERTY_PREFIX + "Class"` to just the prefix, which ends with a dot (`MCR.Solr.IndexRegistry.`). `MCRInstanceName.of()` then appends `.Class`, so the resolved key becomes `MCR.Solr.IndexRegistry..Class` (double dot) instead of `MCR.Solr.IndexRegistry.Class`. `obtainRegistry()` -> `orElseThrow()` will fail.

https://github.com/MyCoRe-Org/mycore/blob/a7b15d8b7a675ed0db8ed3d8a8b85f8d2e04ac6a/mycore-solr/src/main/java/org/mycore/solr/MCRSolrConstants.java#L32-L34

**2. Solr `MCRSolrPathDocumentFactory` uses the wrong property constant**

The removed `CLASS_PROPERTY` (`MCR.Solr.SolrInputDocument.Path.Factory`) was replaced with `SOLR_CONFIG_PREFIX` (`MCR.Solr.`), so the lookup resolves `MCR.Solr..Class` instead of `MCR.Solr.SolrInputDocument.Path.Factory.Class`. Both factory methods will throw. This breaks file indexing.

https://github.com/MyCoRe-Org/mycore/blob/a7b15d8b7a675ed0db8ed3d8a8b85f8d2e04ac6a/mycore-solr/src/main/java/org/mycore/solr/index/file/MCRSolrPathDocumentFactory.java#L51-L58

**3. neo4j: `MCRMetaXML` parser is no longer registered**

The parser code now reads sub-properties under prefix `MCR.Neo4J.Parser.`, and all entries were renamed accordingly, except this one which still uses the old `ParserClass` segment. `startsWith("MCR.Neo4J.Parser.")` does not match `MCR.Neo4J.ParserClass.MCRMetaXML.Class`, so the parser is silently dropped. The commented-out lines 39/40 have the same issue.

https://github.com/MyCoRe-Org/mycore/blob/a7b15d8b7a675ed0db8ed3d8a8b85f8d2e04ac6a/mycore-neo4j/src/main/resources/components/neo4j/config/mycore.properties#L35-L37

**4. JavaDoc example: `.Class` on the value side**

`MCR.PI.Generator.myGenerator=org.mycore.pi.urn.MCRGenericPIGenerator.Class` puts `.Class` on the class-name value. It belongs on the key: `MCR.PI.Generator.myGenerator.Class=org.mycore.pi.urn.MCRGenericPIGenerator`.

https://github.com/MyCoRe-Org/mycore/blob/a7b15d8b7a675ed0db8ed3d8a8b85f8d2e04ac6a/mycore-pi/src/main/java/org/mycore/pi/MCRGenericPIGenerator.java#L59-L61

**5. Malformed HTML in JavaDoc**

`The </code>class.</code> part has to be set` - the first tag should be an opening `<code>`.

https://github.com/MyCoRe-Org/mycore/blob/a7b15d8b7a675ed0db8ed3d8a8b85f8d2e04ac6a/mycore-base/src/main/java/org/mycore/common/xml/MCRURIResolver.java#L1249-L1251

#### Additional notes

- This is a breaking change (public `MCRInstanceName.Suffix` enum removed, `ofClass`/`ofClassName` signatures changed, public constants/methods renamed). Per our conventions, breaking changes should be marked in the commit message - both commit messages currently lack a marker.
- `MCRInstanceConfiguration` JavaDoc on `ofName(...)`, `nested(...)` and `nestedMap(...)` still describes the removed `.class` / empty-key behavior and should be updated.
- The literal `"Class"` / `".Class"` is hardcoded in several new places (`MCRInstanceConfiguration`, `MCRConfiguration2`) while `MCRInstanceName.CLASS_SUFFIX` already exists for the same concept - same point that came up in #2922.
- `MCRInstanceConfiguration` constructor now does `properties.remove("Class")`, which mutates `fullProperties` when caller passes the same map for both args (as the public `ofClass`/`ofClassName` API allows), breaking the documented "unmodified reference" invariant. Not hit by production code paths today, but the API contract is unsafe.

🤖 Generated with [Claude Code](https://claude.ai/code)

sebhofmann:
Separate from the review above: the `deprecated.properties` files in this PR use regex-style entries (e.g. `MCR.PI.Service.(\S+)=MCR.PI.Service.$1.Class`) to document property-family renames. The current deprecation mechanism (`MCRConfigurationBase.checkForDeprecatedProperties`) only does exact-key matching, so these regex entries serve as documentation but are not enforced at runtime. There is currently no way in MyCoRe to automate migration checks for dynamic property families.

I'd like to keep this out of the review scope for now, but we should definitely discuss the approach (extending the deprecation engine to be pattern-aware vs. a separate doc-only format) at the next MIR meeting on Tuesday.

toKrause:
> Separate from the review above: the `deprecated.properties` files in this PR use regex-style entries (e.g. `MCR.PI.Service.(\S+)=MCR.PI.Service.$1.Class`) to document property-family renames. The current deprecation mechanism (`MCRConfigurationBase.checkForDeprecatedProperties`) only does exact-key matching, so these regex entries serve as documentation but are not enforced at runtime. There is currently no way in MyCoRe to automate migration checks for dynamic property families.
> 
> I'd like to keep this out of the review scope for now, but we should definitely discuss the approach (extending the deprecation engine to be pattern-aware vs. a separate doc-only format) at the next MIR meeting on Tuesday.

I agree. The current implementation is only suitable for simple one-to-one replacements. Real-world situation have proven to be more complex. I added them anyways, in order to have at least some form of documentation in the code base.

However, we need to keep in mind, that the deprecation status of a property is checked every time that property is accessed via `MCRConfiguration2`-methods, so looping through a list of patterns to match might come with a significant performance penalty.

toKrause:
> * `MCRInstanceConfiguration` constructor now does `properties.remove("Class")`, which mutates `fullProperties` when caller passes the same map for both args (as the public `ofClass`/`ofClassName` API allows), breaking the documented "unmodified reference" invariant. Not hit by production code paths today, but the API contract is unsafe.

I've made changes, so the the internal `properties`-map is always a newly created map and the `fullProperties`-map is always the supplied map (which is never modified by `MCRInstanceConfiguration`, but is purposefully not a, since this is typically the rather large map of all properties)

sebhofmann (dismissed):

sebhofmann (approved):

sebhofmann:
Found while testing this branch together with the corresponding MIR PR:

```
Exception in thread "main" org.mycore.common.config.MCRConfigurationException: Configuration property MCR.Cronjob.Jobs.DNBURNRegistrationCheck is not set.
        at org.mycore.common.config.MCRConfiguration2.getInstanceOfOrThrow(MCRConfiguration2.java:164)
        at org.mycore.mcr.cronjob.MCRCronjobManager.toJob(MCRCronjobManager.java:139)
```

One cronjob property was missed during the rename - it is still in the old suffix-less form:

https://github.com/MyCoRe-Org/mycore/blob/72c3705847f508ed7637ddb6d8f6d49aa69e3e7f/mycore-pi/src/main/resources/components/pi/config/mycore.properties#L8-L9

It should be `MCR.Cronjob.Jobs.DNBURNRegistrationCheck.Class=org.mycore.pi.urn.MCRDNBURNRegistrationCheckCronjob`. All other cronjob entries (jsessions, jobqueue) already use `.Class`.

Side note (probably worth discussing separately, not blocking): `MCRConfiguration2.getInstantiatablePropertyKeys` still treats a single-segment key without a dot as instantiatable:

```java
if (!s.contains(".")) {
    result = true;
} else {
    result = s.endsWith(".Class") && !s.substring(0, s.length() - ".Class".length()).contains(".");
}
```

That is the suffix-less form this PR removes, so the broken property above is still enumerated and only fails later in `toJob` -> `getInstanceOfOrThrow` (which appends `.Class` and finds nothing). Enumeration and resolution are inconsistent here.

PR #2929, opened 2026-04-29
MCR-3695 MCRMetadataManager: per-object locking for write methods

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3695)

## Summary

Introduces a per-`MCRObjectID` lock primitive on `MCRMetadataManager` so that all write methods (`update`/`create`/`delete` × `MCRObject`/`MCRDerivate`) are atomic, and exposes the lock as an opt-in `MCRMetadataManager.lock(MCRObjectID)` API for callers that need to protect their own read-modify-write spans across multiple manager calls.

The optimistic `checkModificationDates` check was a TOCTOU because it lived outside any lock; it now sits inside the locked critical section and is finally reliable.

`MCRPIService.register()` and `MCRPIService.updateFlag()` use the new caller-side `lock(id)` to fix the PI flag race that motivated this work — concurrent flag mutations no longer silently lose entries.

## Design highlights

- `MCRMetadataManager.lock(MCRObjectID)` returns an `AutoCloseable` `MCRObjectLock` (try-with-resources).
- Backed by `ConcurrentHashMap<MCRObjectID, LockEntry>` + reference counting; entries dropped at refcount 0 (no map growth).
- `ReentrantLock` per entry — EventHandlers can re-enter `update()` on the same id without deadlock.
- `tryLock` with configurable timeout; throws `MCRPersistenceException` on timeout — caller can retry, surface, or abort.

## Configuration

| Property | Default | Meaning |
|----------|---------|---------|
| `MCR.Metadata.Manager.LockTimeoutSeconds` | `60` | tryLock timeout before `MCRPersistenceException` |

## Scope

- Single-JVM only. Multi-node clustering would require distributed locking; out of scope.
- No public API breakage. Existing callers gain serialisation transparently. New failure modes for concurrent writers: lock timeout or now-reliable stale `modifyDate` — both `MCRPersistenceException`, which the API already declared.

## Ticket & Documentation

- [ ] For UI changes: before & after screenshots are attached. _(N/A — server-side change)_
- [x] New features or migrations are documented. _(Javadoc on `lock()` + property in `mycore.properties`)_
  - [ ] Yes: Is a migration required? If yes, describe it. _(No migration; transparent for existing callers.)_
  - [ ] Breaking change is marked in the **commit message**. _(No breaking change.)_

## Bugfix-Specific Checks
_(N/A — improvement, not bugfix)_

## Testing
- [x] I have tested the changes locally (`CI=true mvn clean install` green).
- [ ] Were existing tests modified? _(No — only new tests added.)_

## MCR Conventions & Metadata
- [x] [MCR naming conventions](https://www.mycore.de/documentation/developer/conventions/) are followed.
  - [ ] Old API is deprecated or a migration is documented. _(N/A — only additions, no removal/deprecation.)_
  - [x] If not, no action needed. _(API addition only.)_

## Multi-Repo Considerations
  - [ ] If yes, is it already created? _(No MIR change required — purely internal serialisation.)_

## Test plan

- New tests:
  - `MCRMetadataManagerLockTest` — 5 unit tests on the lock primitive: reentrancy, cross-id parallelism, timeout, refcount cleanup, mutual exclusion.
  - `MCRMetadataManagerConcurrentUpdateTest` — 2 deterministic integration tests against a real metadata store. Both fail without the auto-lock:
    - `updateBlocksWhileExternalLockHeld` — caller holds `lock(id)` on a separate thread; main-thread `update(id)` must block until release.
    - `forcedRaceProducesExactlyOneWinner` — `CyclicBarrier` forces both threads to retrieve at the same `modifyDate` snapshot before either persists; with the lock, one wins and the other is rejected as stale; without the lock, lost update.
- Verified: `CI=true mvn clean install` green on this branch.

### DIFF
--- a/mycore-base/src/main/java/org/mycore/datamodel/metadata/MCRMetadataManager.java
+++ b/mycore-base/src/main/java/org/mycore/datamodel/metadata/MCRMetadataManager.java
@@ -989,12 +989,18 @@ public final class MCRMetadataManager {
 
         private final MCRObjectID id;
 
+        private boolean closed;
+
         MCRObjectLock(MCRObjectID id) {
             this.id = id;
         }
 
         @Override
         public void close() {
+            if (closed) {
+                return;
+            }
+            closed = true;
             LOCKS.compute(id, (k, v) -> {
                 v.lock.unlock();
                 return --v.refCount == 0 ? null : v;
--- a/mycore-base/src/main/java/org/mycore/datamodel/metadata/MCRMetadataManager.java
+++ b/mycore-base/src/main/java/org/mycore/datamodel/metadata/MCRMetadataManager.java
@@ -33,6 +33,10 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.Callable;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.ReentrantLock;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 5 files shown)


mycore-base/src/main/java/org/mycore/datamodel/metadata/MCRMetadataManager.java
```diff
@@ -909,4 +936,75 @@ public static MCRObjectID getObjectId(MCRObjectID derivateID) {
...
+        });
+        boolean acquired = false;
+        try {
+            acquired = entry.lock.tryLock(getLockTimeoutSeconds(), TimeUnit.SECONDS);
+            if (!acquired) {
+                throw new MCRPersistenceException("Lock timeout for " + id);
+            }
+            return new MCRObjectLock(id);
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
```
toKrause, line 966: Should we do this in the ~50 other places that catch an `InteruptedException` too? In a separate PR.

toKrause (approved):

mycore-base/src/main/java/org/mycore/datamodel/metadata/MCRMetadataManager.java
sebhofmann: The Recommended behavior is to rethrow InterruptedException or `Thread.currentThread().interrupt();` and throw a suitable Exception for the Api.
If its the thread root, then nothing needs to be done.

PR #2934, opened 2026-04-30
MCR-3697 texteditor always loads expanded object

Mewel:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3697).

## Ticket & Documentation

## Bugfix-Specific Checks

## Testing

## MCR Conventions & Metadata

## Multi-Repo Considerations

### DIFF
--- a/mycore-restapi/src/main/java/org/mycore/restapi/v2/MCRRestObjects.java
+++ b/mycore-restapi/src/main/java/org/mycore/restapi/v2/MCRRestObjects.java
@@ -410,8 +410,8 @@ public class MCRRestObjects {
     @Operation(
         summary = "Returns MCRObject with the given " + PARAM_MCRID + ".",
         tags = MCRRestUtils.TAG_MYCORE_OBJECT)
-    public Response getObject(@Parameter(example = "mir_mods_00004711") @PathParam(PARAM_MCRID) MCRObjectID id)
-        throws IOException {
+    public Response getObject(@Parameter(example = "mir_mods_00004711") @PathParam(PARAM_MCRID) MCRObjectID id,
+        @QueryParam("expanded") Boolean expandedParm) throws IOException {
         long modified;
         try {
             modified = MCRXMLMetadataManager.getInstance().getLastModified(id);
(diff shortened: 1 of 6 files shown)


mycore-restapi/src/main/java/org/mycore/restapi/v2/MCRRestObjects.java
```diff
@@ -410,8 +410,8 @@ public Response createObject(String xml) {
     @Operation(
         summary = "Returns MCRObject with the given " + PARAM_MCRID + ".",
         tags = MCRRestUtils.TAG_MYCORE_OBJECT)
-    public Response getObject(@Parameter(example = "mir_mods_00004711") @PathParam(PARAM_MCRID) MCRObjectID id)
-        throws IOException {
+    public Response getObject(@Parameter(example = "mir_mods_00004711") @PathParam(PARAM_MCRID) MCRObjectID id,
+        @QueryParam("expanded") Boolean expandedParm) throws IOException {
```
rsteph-de: in REST v2 classifications we use a [Mediatype parameter](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/MIME_types)  (see MCRDetailLevel) for different granularities of the response.

This is not my preferred choice, but we should keep the endpoints consistent.

rsteph-de (changes requested):

mycore-restapi/src/main/java/org/mycore/restapi/v2/MCRRestObjects.java
Mewel: What would you suggest for the expanded version? Using the _detail_ level or adding a special one?
Right now I didn't change the response, expanded is true by default. Which probably means detail level = _normal_. In this case _summary_ would be the unexpanded object.

I'm not happy with this, so we should discuss it. The REST API needs a way to distinguish between both versions.

sebhofmann: I would expect the `NORMAL` level to be the non expanded and `DETAILED` to be the expanded variant, but this would break the api compatibility. Using `NORMAL`  as expanded and `SUMMARY` for the non expanded would be backward compatible.

sebhofmann: expandedParm -> expandedParam

sebhofmann (approved):

PR #2945, opened 2026-05-06
MCR-3700 record externally supplied IDs in MCRFileBaseCacheObjectIDGenerator

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3700)

## Problem

`MCRFileBaseCacheObjectIDGenerator` (introduced in MCR-3046) only updates its per-base-id cache file from inside `getNextFreeId(...)`. `MCRMetadataManager.create(MCRObject|MCRDerivate)` skips the generator when the caller supplies a fixed (non-zero) ID, so the cache stays behind the actual store. Subsequent `getNextFreeId(...)` calls then return IDs that collide with already-stored objects (`MCRPersistenceException("... already exists")`).

The default `MCRDefaultObjectIDGenerator` is unaffected because it reconciles against `MCRXMLMetadataManager.getHighestStoredID(...)`. The cache-based generator deliberately avoids that scan (expensive on OCFL because of hashed paths) and therefore cannot self-heal.

## Fix

- New optional sub-interface `MCRTrackingObjectIDGenerator` extends `MCRObjectIDGenerator` with `recordID(MCRObjectID id)`. Existing third-party `MCRObjectIDGenerator` implementations are unaffected.
- `MCRFileBaseCacheObjectIDGenerator` implements the new interface. `recordID` acquires the existing per-base-id write lock + file lock and writes the ID into the cache file iff it is greater than the currently cached value (otherwise no-op).
- `MCRMetadataManager.create(MCRObject)` and `create(MCRDerivate)` call a new private helper `recordIDIfTracking(...)` after the ID has been determined; the helper uses `instanceof MCRTrackingObjectIDGenerator` and is a no-op for generators that do not opt in.
- Two new unit tests: `recordIDAdvancesCacheWhenIdIsHigher`, `recordIDIgnoresLowerOrEqualId`.

## Repository healing

Existing repositories with corrupted caches can be repaired with the existing CLI command `generate id cache`, which rebuilds the cache files from the stored objects.

## Forward merge

This PR targets `2023.06.x` (where the bug originated). After merge it must be merged forward to `2024.06.x`, `2025.06.x`, `2025.12.x` and `main`.

## Test plan

- [x] `mvn -pl mycore-base -am test` (608/608 green incl. the two new tests)
- [ ] `CI=true mvn clean install` (full multi-module build)
- [ ] manual verification on an OCFL deployment: create object with fixed ID, then create object with id `0`, confirm no collision

### DIFF
--- a/mycore-base/src/main/java/org/mycore/datamodel/common/MCRFileBaseCacheObjectIDGenerator.java
+++ b/mycore-base/src/main/java/org/mycore/datamodel/common/MCRFileBaseCacheObjectIDGenerator.java
@@ -47,7 +47,7 @@ import org.mycore.datamodel.metadata.MCRObjectID;
  * given base id. The cache file is located in the data directory of MyCoRe and is named "id_cache" and contains one
  * file for each base id. The file contains the last generated id as a string.
  */
-public class MCRFileBaseCacheObjectIDGenerator implements MCRObjectIDGenerator {
+public class MCRFileBaseCacheObjectIDGenerator implements MCRTrackingObjectIDGenerator {
 
     private static final Logger LOGGER = LogManager.getLogger();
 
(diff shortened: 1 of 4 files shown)


rsteph-de (approved):
OK,

In a new PR we schould cleanup the code in `main` and extract a `readIDFromFileChannel()` method.

PR #2949, opened 2026-05-13
MCR-3704 StackOverflowError in MCRScopedSession

Mewel:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3704).

## Ticket & Documentation

## Bugfix-Specific Checks

## Testing

## MCR Conventions & Metadata

## Multi-Repo Considerations

### DIFF
--- a/mycore-base/src/main/java/org/mycore/common/MCRScopedSession.java
+++ b/mycore-base/src/main/java/org/mycore/common/MCRScopedSession.java
@@ -78,7 +78,7 @@ public final class MCRScopedSession extends MCRSession {
      * @return the result of the action
      */
     public <T> T doAs(ScopedValues scopeValues, Supplier<T> action) {
-        Objects.requireNonNull(scopeValues).map.put(SCOPED_HINT, scopeValues);
+        Objects.requireNonNull(scopeValues).map.put(SCOPED_HINT, Boolean.TRUE);
         scopedValues.set(scopeValues);
         try {
             return action.get();
--- a/mycore-base/src/test/java/org/mycore/common/MCRScopedSessionTest.java
+++ b/mycore-base/src/test/java/org/mycore/common/MCRScopedSessionTest.java
@@ -24,11 +24,18 @@ import org.junit.Test;
 
 public class MCRScopedSessionTest extends MCRTestCase {
 
+    /**
+     * Verifies that MCR-3704 does not reappear when scoped values are rendered as text.
+     */
     @Test
     public void doAsDoesNotCreateRecursiveScopedValuesStringRepresentation() {
         MCRScopedSession session = new MCRScopedSession();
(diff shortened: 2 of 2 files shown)


sebhofmann:
## Code Review

### Fix looks correct

The one-line change in `MCRScopedSession.java:81` (storing `Boolean.TRUE` instead of the `ScopedValues` record itself) correctly fixes the root cause: the record's auto-generated `toString()` was recursing because the record's own `map` field contained the record. The fix is minimal and safe — the only production consumer of `SCOPED_HINT` is `MCRAccessCacheManager.java:55`, which only checks the value for `null`, so changing the stored value type has no impact.

### Main concern: the test does not actually reproduce the bug

`MCRScopedSessionTest#doAsDoesNotCreateRecursiveScopedValuesStringRepresentation` only asserts that the marker equals `Boolean.TRUE`. It never invokes `toString()` on anything, so the StackOverflowError the PR claims to prevent is never triggered.

Consequences:
- If someone reverts to `map.put(SCOPED_HINT, scopeValues)`, the test would fail — but with an assertion mismatch, not the actual `StackOverflowError`, so the failure message would not point to the real bug.
- If someone changes the marker to a different non-recursive value (e.g. a `String`), the test would fail even though no overflow would occur.
- The test is essentially asserting an implementation detail rather than the behavior that was broken.

**Suggested improvement** — add an assertion that actually exercises the original failure mode:

```java
session.doAs(values, () -> {
    // Would have thrown StackOverflowError before the fix
    values.toString();
    session.getMapEntries().toString();
});
```

That literally reproduces the crash on the buggy code and guards against any regression that re-introduces a self-referential structure, regardless of marker type.

### Additional coverage gaps (optional)

- The `Runnable` overload of `doAs` is not covered (separate entry point, even though it delegates internally).
- Nested `doAs` calls are not tested. `clearScopedValues()` unconditionally removes the `ThreadLocal`, which could clobber an outer scope.
- Exception path is not tested — does `clearScopedValues()` run when `action` throws? The `finally` block suggests yes, but it is currently unverified.

### Minor

A short Javadoc on the test method referencing MCR-3704 would help future readers understand what regression this guards against.

### Strengths

- License header, package, naming, and English comments all follow project conventions.
- Test name clearly describes intent.
- Extending `MCRTestCase` is consistent with the rest of the test suite.
- Fix is minimal and targeted — no unrelated refactoring.

### Recommendation

Strengthen the test to invoke `toString()` inside the `doAs` block so it reproduces the original `StackOverflowError` on the buggy code. After that, this is ready to merge.

sebhofmann (approved):
Thanks for the quick turnaround! The added `values.toString()` and `session.getMapEntries().toString()` calls inside `doAs` now reproduce the original StackOverflowError on the buggy code, which is exactly the regression guard the test needed. The Javadoc reference to MCR-3704 is a nice touch too. LGTM.

PR #2974, opened 2026-06-09
MCR-3662 Change of servstate is not recognized as metadata change

Possommi:
[MCR-3662](https://mycore.atlassian.net/browse/MCR-3662).

## Ticket & Documentation

## Bugfix-Specific Checks

## Testing

## MCR Conventions & Metadata

## Multi-Repo Considerations

[MCR-3662]: https://mycore.atlassian.net/browse/MCR-3662?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mycore-orcid2/src/main/java/org/mycore/orcid2/MCRORCIDUtils.java
+++ b/mycore-orcid2/src/main/java/org/mycore/orcid2/MCRORCIDUtils.java
@@ -78,10 +78,11 @@ public class MCRORCIDUtils {
      * Checks if MCRObjects' state is ready to publish.
      *
      * @param object the MCRObject
-     * @return true if state is ready to publish or there is not state
+     * @return true if state is ready to publish or there is not a state present
      */
     public static boolean checkPublishState(MCRObject object) {
-        return getStateValue(object).map(MCRORCIDUtils::checkPublishState).orElse(true);
+        String stateValue = getStateValue(object);
+        return stateValue == null ? true : checkPublishState(stateValue);
     }
 
     /**
@@ -207,7 +208,14 @@ public class MCRORCIDUtils {
         return new MCRIdentifier(element.getAttributeValue("type"), element.getTextTrim());
     }
 
-    public static Optional<String> getStateValue(MCRObject object) {
-        return Optional.ofNullable(object.getService().getState()).map(MCRCategoryID::getId);
+    /**
+     * Retrieves the state value of the given {@link MCRObject}.
+     *
+     * @param object the {@link MCRObject} whose state value is to be retrieved
+     * @return the category id value of the object's state as a {@code String}, or {@code null} if no state is available
+     */
+    public static String getStateValue(MCRObject object) {
+        Optional<String> v = Optional.ofNullable(object.getService().getState()).map(MCRCategoryID::getId);
+        return v.isPresent() ? v.get() : null;
     }
 }
--- a/mycore-orcid2/src/main/java/org/mycore/orcid2/MCRORCIDUtils.java
+++ b/mycore-orcid2/src/main/java/org/mycore/orcid2/MCRORCIDUtils.java
@@ -215,7 +215,6 @@ public class MCRORCIDUtils {
      * @return the category id value of the object's state as a {@code String}, or {@code null} if no state is available
      */
     public static String getStateValue(MCRObject object) {
-        Optional<String> v = Optional.ofNullable(object.getService().getState()).map(MCRCategoryID::getId);
-        return v.isPresent() ? v.get() : null;
+        return Optional.ofNullable(object.getService().getState()).map(MCRCategoryID::getId).orElse(null);
     }
 }
--- a/mycore-orcid2/src/main/java/org/mycore/orcid2/MCRORCIDUtils.java
+++ b/mycore-orcid2/src/main/java/org/mycore/orcid2/MCRORCIDUtils.java
@@ -207,7 +207,7 @@ public class MCRORCIDUtils {
         return new MCRIdentifier(element.getAttributeValue("type"), element.getTextTrim());
     }
 
-    private static Optional<String> getStateValue(MCRObject object) {
+    public static Optional<String> getStateValue(MCRObject object) {
         return Optional.ofNullable(object.getService().getState()).map(MCRCategoryID::getId);
     }
 }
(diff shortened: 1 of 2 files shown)


sebhofmann (approved):

mycore-orcid2/src/main/java/org/mycore/orcid2/work/MCRORCIDWorkEventHandler.java
```diff
@@ -146,9 +152,15 @@ private void handlePublication(MCRObject object) {
...
-                LOGGER.info("Metadata does not changed. Skipping {}...", objectID);
+
+            boolean changedStateToPublished = changedToPublished(MCRORCIDUtils.getStateValue(object),
+                MCRORCIDUtils.getStateValue(outdatedObject));
+
+            boolean changedMetadata = MCRXMLHelper.deepEqual(new MCRMODSWrapper(object).getMODS(),
+                new MCRMODSWrapper(outdatedObject).getMODS());
+
+            if (!changedMetadata && !changedStateToPublished) {
+                LOGGER.info("Metadata did not change. Skipping {}...", objectID);
```
golsch: The log message should be revised.

mycore-orcid2/src/main/java/org/mycore/orcid2/MCRORCIDUtils.java
```diff
@@ -207,7 +207,7 @@ private static MCRIdentifier getIdentfierFromElement(Element element) {
         return new MCRIdentifier(element.getAttributeValue("type"), element.getTextTrim());
     }
 
-    private static Optional<String> getStateValue(MCRObject object) {
+    public static Optional<String> getStateValue(MCRObject object) {
```
golsch: Return null instead of an optional value.

mycore-orcid2/src/main/java/org/mycore/orcid2/work/MCRORCIDWorkEventHandler.java
```diff
@@ -193,6 +205,17 @@ private void handlePublication(MCRObject object) {
...
+        if (!newState.isPresent() || !oldState.isPresent()) {
+            return false;
+        }
+
+        String s1 = newState.get();
+        String s2 = oldState.get();
+
+        return PUBLISH_STATES.contains(s1) && !s1.equals(s2);
+    }
+
```
golsch: The publish state is already checked earlier in L148. Here, only need to check whether the states differ.

golsch (changes requested):

mycore-orcid2/src/main/java/org/mycore/orcid2/MCRORCIDUtils.java
Possommi: Changing the return type will require additional api changes. I only changed the visibility of the method to avoid code duplication.

mycore-orcid2/src/main/java/org/mycore/orcid2/work/MCRORCIDWorkEventHandler.java
Possommi: This method checks two states and can be used without any context (e.g. a state value has been checked earlier). I think this is just more flexible

mycore-orcid2/src/main/java/org/mycore/orcid2/MCRORCIDUtils.java
golsch: This is only used internally in a single place. The Optional is unnecessary here and makes the comparison more complicated rather than simpler.

mycore-orcid2/src/main/java/org/mycore/orcid2/work/MCRORCIDWorkEventHandler.java
golsch: As mentioned, `if (!MCRORCIDUtils.checkPublishState(object)) {` already performs this check. The additional check `PUBLISH_STATES.contains(s1)` is unnecessary and also inconsistent with the previous one, since the property is documented in mycore.properties as may having `*` semantics.

mycore-orcid2/src/main/java/org/mycore/orcid2/MCRORCIDUtils.java
Possommi: It is also used in [public static boolean checkPublishState(MCRObject object)](https://github.com/MyCoRe-Org/mycore/blob/7e2ec71a37ffa691cb380390b9c4a7d8661ead34/mycore-orcid2/src/main/java/org/mycore/orcid2/MCRORCIDUtils.java#L83). This method must be altered aswell then.

mycore-orcid2/src/main/java/org/mycore/orcid2/work/MCRORCIDWorkEventHandler.java
Possommi: I changed semantics of the method and changed its name.

mycore-orcid2/src/main/java/org/mycore/orcid2/MCRORCIDUtils.java
golsch: Yes, that's what I mean. All you need is an `Optional#ofNullable`. That way, you won't have to handle the `Optional` in `statesDiffer` anymore and `Objects#equals` is sufficient because the check performed earlier already confirms that the state is publishable.

mycore-orcid2/src/main/java/org/mycore/orcid2/work/MCRORCIDWorkEventHandler.java
golsch: `if (!newState.isPresent() || !oldState.isPresent()) {`
This still does not fix the desired behavior: if the state changes from null to publish, nothing happens.

mycore-orcid2/src/main/java/org/mycore/orcid2/MCRORCIDUtils.java
Possommi: I do not question the perks that this may have. I did not want to change so much just for a bugfix. Anyways, I applied the requested changes.

golsch (approved):

PR #2982, opened 2026-06-12
MCR-3722 Improve support for entangled PIs

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3722).

## Ticket & Documentation

## Testing

## MCR Conventions & Metadata

## Multi-Repo Considerations

### DIFF
--- a/mycore-pi/src/main/java/org/mycore/pi/MCRPICreationEventHandler.java
+++ b/mycore-pi/src/main/java/org/mycore/pi/MCRPICreationEventHandler.java
@@ -19,10 +19,14 @@
 package org.mycore.pi;
 
 import java.util.List;
+import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.ExecutionException;
-import java.util.function.Predicate;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.mycore.access.MCRAccessException;
 import org.mycore.common.MCRException;
 import org.mycore.common.events.MCREvent;
@@ -31,6 +35,8 @@ import org.mycore.datamodel.metadata.MCRObject;
 import org.mycore.pi.exceptions.MCRPersistentIdentifierException;
 
 public class MCRPICreationEventHandler extends MCREventHandlerBase {
+
+    private static final Logger LOGGER = LogManager.getLogger();
 
     @Override
     protected void handleObjectCreated(MCREvent evt, MCRObject obj) {
@@ -43,26 +49,60 @@ public class MCRPICreationEventHandler extends MCREventHandlerBase {
     }
 
     private void processPIServices(MCRObject obj) {
-        List<MCRPIRegistrationInfo> registered = MCRPIManager.getInstance().getRegistered(obj);
 
-        final List<String> services = registered.stream().map(MCRPIRegistrationInfo::getService)
-            .collect(Collectors.toList());
+        // list of service IDs for PI service for which an PI is already registered
+        Set<String> servicesWithRegisteredIdentifier = MCRPIManager
+            .getInstance()
+            .getRegistered(obj)
+            .stream()
+            .map(MCRPIRegistrationInfo::getService)
+            .collect(Collectors.toSet());
 
-        List<MCRPIService<MCRPersistentIdentifier>> listOfServicesWithCreatablePIs = MCRPIServiceManager
-            .getInstance().getAutoCreationList().stream()
-            .filter(Predicate.not(s -> services.contains(s.getServiceID())))
-            .filter(Predicate.not(s -> MCRPIService.hasFlag(obj, "", s)))
-            .filter(s -> s.getCreationPredicate().test(obj))
-            .collect(Collectors.toList());
+        // map of auto-creating PI services for which no PI is registered yet
+        Map<String, MCRPIService<MCRPersistentIdentifier>> autoCreatingServices = MCRPIServiceManager
+            .getInstance()
+            .getAutoCreationList()
+            .stream()
+            .filter(service -> !servicesWithRegisteredIdentifier.contains(service.getServiceID()))
+            .filter(service -> !MCRPIService.hasFlag(obj, "", service))
+            .collect(Collectors.toMap(MCRPIService::getServiceID, Function.identity()));
 
-        listOfServicesWithCreatablePIs
-            .forEach((serviceToRegister) -> {
+        boolean mayRegisterAdditionalIdentifiers = true;
+        while (mayRegisterAdditionalIdentifiers && !autoCreatingServices.isEmpty()) {
+
+            // list of auto-creating PI services that will register a PI in this iteration
+            List<MCRPIService<MCRPersistentIdentifier>> matchingAutoCreatingServices = autoCreatingServices
+                .values()
+                .stream()
+                .filter(service -> service.getCreationPredicate().test(obj))
+                .toList();
+
+            // additional PIs will be registered in this iteration, so the conditions will change
+            // and additional predicates may evaluate to true in the next iteration
+            mayRegisterAdditionalIdentifiers = !matchingAutoCreatingServices.isEmpty();
+
+            for (MCRPIService<MCRPersistentIdentifier> service : matchingAutoCreatingServices) {
                 try {
-                    serviceToRegister.register(obj, "", false);
+
+                    // register new PI, succeeds or throws exception
+                    MCRPersistentIdentifier identifier = service.register(obj, "", false);
+
+                    if (LOGGER.isInfoEnabled()) {
+                        LOGGER.info("Created new identifier {} for object  {} using auto-creating PI service {}",
+                            identifier, obj.getId(), service.getServiceID());
+                    }
+
+                    // each auto-creating PI service ony gets one shot at registering a PI
+                    autoCreatingServices.remove(service.getServiceID());
+
                 } catch (MCRAccessException | MCRPersistentIdentifierException | ExecutionException
                     | InterruptedException e) {
                     throw new MCRException("Error while register pi for object " + obj.getId().toString(), e);
                 }
-            });
+            }
+
+        }
+
     }
+
 }
--- a/mycore-pi/src/main/java/org/mycore/pi/MCRPICreationEventHandler.java
+++ b/mycore-pi/src/main/java/org/mycore/pi/MCRPICreationEventHandler.java
@@ -20,7 +20,6 @@ package org.mycore.pi;
 
 import java.util.List;
 import java.util.Map;
-import java.util.Set;
 import java.util.concurrent.ExecutionException;
 import java.util.function.Function;
 import java.util.stream.Collectors;
@@ -50,20 +49,11 @@ public class MCRPICreationEventHandler extends MCREventHandlerBase {
 
     private void processPIServices(MCRObject obj) {
 
-        // list of service IDs for PI service for which an PI is already registered
-        Set<String> servicesWithRegisteredIdentifier = MCRPIManager
-            .getInstance()
-            .getRegistered(obj)
-            .stream()
-            .map(MCRPIRegistrationInfo::getService)
-            .collect(Collectors.toSet());
-
         // map of auto-creating PI services for which no PI is registered yet
         Map<String, MCRPIService<MCRPersistentIdentifier>> autoCreatingServices = MCRPIServiceManager
             .getInstance()
             .getAutoCreationList()
             .stream()
-            .filter(service -> !servicesWithRegisteredIdentifier.contains(service.getServiceID()))
             .filter(service -> !MCRPIService.hasFlag(obj, "", service))
             .collect(Collectors.toMap(MCRPIService::getServiceID, Function.identity()));
 
--- a/mycore-pi/src/main/java/org/mycore/pi/MCRPICreationEventHandler.java
+++ b/mycore-pi/src/main/java/org/mycore/pi/MCRPICreationEventHandler.java
@@ -82,7 +82,7 @@ public class MCRPICreationEventHandler extends MCREventHandlerBase {
                             identifier, obj.getId(), service.getServiceID());
                     }
 
-                    // each auto-creating PI service ony gets one shot at registering a PI
+                    // each auto-creating PI service only gets one shot at registering a PI
                     autoCreatingServices.remove(service.getServiceID());
 
                 } catch (MCRAccessException | MCRPersistentIdentifierException | ExecutionException
--- a/mycore-pi/src/main/java/org/mycore/pi/MCRPICreationEventHandler.java
+++ b/mycore-pi/src/main/java/org/mycore/pi/MCRPICreationEventHandler.java
@@ -78,7 +78,7 @@ public class MCRPICreationEventHandler extends MCREventHandlerBase {
                     MCRPersistentIdentifier identifier = service.register(obj, "", false);
 
                     if (LOGGER.isInfoEnabled()) {
-                        LOGGER.info("Created new identifier {} for object  {} using auto-creating PI service {}",
+                        LOGGER.info("Created new identifier {} for object {} using auto-creating PI service {}",
                             identifier, obj.getId(), service.getServiceID());
                     }
 
--- a/mycore-pi/src/main/java/org/mycore/pi/util/MCROtherPIValueExtractor.java
+++ b/mycore-pi/src/main/java/org/mycore/pi/util/MCROtherPIValueExtractor.java
@@ -1,0 +1,82 @@
+/*
+ * This file is part of ***  M y C o R e  ***
+ * See https://www.mycore.de/ for details.
+ *
+ * MyCoRe is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * MyCoRe is distributed in the hope that it will be useful,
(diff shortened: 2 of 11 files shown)


sebhofmann (commented):
Thanks for tackling this, the loop in `MCRPICreationEventHandler` is the right core idea. Two things I would like to see changed in this PR, plus an idea for later.

1. **Fold the extraction into `MCRGenericPIGenerator` and read it from the object, not from the database.** `MCROtherPIDOIGenerator` and `MCROtherPIURNGenerator` are thin wrappers that differ only in the PI type they produce, and we agreed that `MCRGenericPIGenerator` is the generator we point people to. What they do cannot be replaced by an `XPath.n` expression though: identifiers in the metadata are not necessarily minted by a PI service, they can come from an import or from the editor form, and `MCRPIXPathMetadataService` writes into the same field without recording which service produced the value. So XPath cannot tell "the URN our DNBURN service minted" apart from "some URN someone typed in", and would silently derive a DOI suffix from a foreign identifier. That part of your PR is solving a real problem.

   I would not read it from the `MCRPI` table either. `register()` runs its work through `REGISTER_POOL` in an `MCRFixedUserCallable`, so it commits in a separate session and transaction, and the table is not reliably current from the perspective of the calling code. The `MyCoRe-PI` flags on the object are the better source: `addFlagToObject` attaches them to the very instance that travels through the loop, they carry the service ID, and they are persisted in the object's `service` section anyway. `MCRPIService.getFlags(MCRBase, String, MCRPIService<?>)` already does the filtering, it just needs an overload taking the service ID as a `String` so the generator does not have to instantiate the service.

   On the configuration syntax: the type selector is redundant, a service has exactly one type (`MCRPIService` takes it in the constructor and `getType()` is final), so the service ID alone identifies the source. Suggestion:

   ```
   MCR.PI.Generator.DNBURNDOI.Class=org.mycore.pi.MCRGenericPIGenerator
   MCR.PI.Generator.DNBURNDOI.Type=doi
   MCR.PI.Generator.DNBURNDOI.GeneralPattern=10.1234/$Dependency.DNBURN.2
   MCR.PI.Generator.DNBURNDOI.Dependency.DNBURN=([a-z0-9:]+)-([0-9]+)
   ```

   `$Dependency.<ServiceID>.<n>` resolves to capture group `n` of the configured expression, so several groups of one identifier can be used in different places of the pattern. Without an index, `$Dependency.<ServiceID>` would resolve to the full identifier, which means the common case needs no expression at all. Missing flag or non matching expression should fail loudly with an `MCRPersistentIdentifierException`. The validation in `MCROtherPIValueExtractor` can stay, it just needs to allow more than one group and take its input from the flags.

2. **Keep the loop, but harden it.** Add an iteration cap bounded by the number of auto creating services, since termination currently relies on every `register` implementation setting the PI flag. Add a WARN for services whose creation predicate is still `false` after the last pass, otherwise a broken configuration fails silently. Also, `services` is computed once before the loop and never refreshed, it only works because the flag filter happens to catch the same case.

Optional follow up, no need to touch it here: it might be worth modelling dependencies as their own concept, separate from creation predicates. A predicate answers "may this PI be created", a dependency answers "what has to exist first", and expressing the second through the first is what forces the brute force loop in the first place. A dedicated `MCR.PI.Service.Datacite.Dependencies=DNBURN` checked by the event handler alongside the predicate would give it a topological order, cycle detection and a loud failure on a missing service. It would also line up with the generator syntax above, where `Dependency.DNBURN=...` would then only attach an expression to an already declared dependency. Just an idea for a separate ticket.

Independent of all this, `MCR.PI.Generator.DateDOI.Class` in `mycore.properties` is unrelated to the ticket and should move to its own.

toKrause:
Taking the PI information from the service flags, rather than the database is done in [MCR-3731](https://github.com/MyCoRe-Org/mycore/pull/2993). The ticket also gives a real-world example where this is necessary. Both ideas are important enough that they deserve their own PR, so I'd rather not combine these things into a single PR. This PR should only add the new capability.

Replacing (almost) all custom generators with `MCRGenericPIGenerator` should be its own PR. A ticket already exists (MCR-3763). Best course of action is to add this small new functionality and clean-up the existing generators (MCR-3723) and then, once we know all the things that `MCRGenericPIGenerator` should be capable of doing, extend or redesign that class (MCR-3763) and create corresponding up-to-date documentation (and a migration guide). That might be a more extensive task than we might think at this moment, so it certainly warrants its separate ticket (and we shouldn't stress ourselves with that task right now, if need be, that can be done after the next release).

I've created a ticket for the follow-up idea (MCR-3799).

If `Service` and `Type` are depended, then why does `MCRPIManager#getCreatedIdentifiers` need both parameters? that is the reason why `MCROtherPIValueExtractor` (and therefore the two new generators) currently require both values.

I will make the loop in `MCRPICreationEventHandler` more robust.




[MCR-3731]: https://mycore.atlassian.net/browse/MCR-3731?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

sebhofmann:
1. We agreed in the dev call that everything should go into `MCRGenericPIGenerator` and that this is the variant we communicate as the preferred one. That is the defined goal, and MCR-3763 is the ticket that gets us there. Adding two new specialized generators now works against that goal. They have to be removed again under MCR-3763, so the same capability gets built twice and I redo work that could be done in the right place right away. Please put it into `MCRGenericPIGenerator` instead.

2. The fact that #2993 already implements the lookup via the service flags is not a reason to implement this feature against the database here. That would knowingly build on the path MCR-3731 replaces, and it would have to be reworked immediately afterwards. If the overlap between the two PRs is the concern, then the order should be the other way round: merge #2993 first and let this PR build on the flags. You defined that order explicitly across your PRs, so it can just as well be defined the other way round.

3. I also do not follow the argument that we first need to find out what `MCRGenericPIGenerator` should be capable of. It is already type agnostic: it takes the PI type as a property and resolves the parser from the registry, so one class covers DOIs, URNs and every other registered type. What is missing there are individual placeholders, and MCR-3763 already names them, `$CreationDate` and `$UUID`. Adding one is a small and local change. The capability in this PR is exactly of that kind, so putting it directly into `MCRGenericPIGenerator` is a matter of minutes, instead of two new classes that then get base classes and tests in #2983, are touched again in #3040 and are deleted under MCR-3763. The same holds for the URN and DOI symmetry in #2983: `MCRCurrentDateDOIGenerator`, `MCRCreateDateDNBURNGenerator` and `MCRCurrentDateDNBURNGenerator` are new counterparts of existing implementations, and MCR-3763 lists exactly those pairs for removal, so they are built in order to be deleted again. And these are large changes to review, #2983 is +2422/-781 across 45 files and #3040 is +3874/-787 across 64 files, so that review effort lands on me as well.

4. On `MCRPIManager#getCreatedIdentifiers` needing both parameters: that is a low level query over the `MCRPI` table, where type and service are two independent columns, so the redundancy sits at the call site rather than in the model. `MCRPIService#isCreated` simply passes its own type and its own service ID. Every concrete service fixes its type in a no-arg constructor (`super(MCRDNBURN.TYPE)`, `super(TYPE)`, `super("handle")`) and `getType()` is final, so a service has exactly one type, and it can be derived via `MCRPIServiceManager#getRegistrationService(id).getType()`. Requiring both in a configuration property only makes an inconsistent pair configurable that then silently matches nothing. Once the lookup moves to the service flags it is moot anyway, `matches()` compares service ID and `additional` only.

Thanks for taking on the loop hardening, that part is settled from my side.

toKrause:
We should discuss this in the next web conference.

Some (random) thoughts:

---

What happened was:

- I created this PR (adding generators that work with existing predicates and using the same approach as the existing predicated) and noticed that some of the existing generator classes are in need of clean-up / modernization
- I created a PR that implemented that clean-up / modernization (MCR-3723)
- I noticed an issue with the interactions between the new generators and PI registration in certain circumstances (when PI registration is triggered from a WebCLI command, MCR-3731)
- I created a PR that fixes that bug, building upon the cleaned-up / modernized code
- In a later web conference, we decided that we should extend the capabilities of the generic generator and should (in a future PR) replace existing implementations with the generic implementation.

This is a normal incremental approach towards improvement (and I would prefer to keep it that way). I'm fully okay with creating classes that are deprecated or removed in the near future, if that is a side-effect of this approach.

When we gain additional information/knowledge, we usually don't go back and change the work done in the past. That's why the release branches are protected. The only difference to what usually happens is that in this case some of the work done in the past (i.e., this PR) hasn't been merged yet (because development speed was faster then integration speed). I'm open to small adjustments to the current PR, but I'd rather not redo and reorder work I have already done.

The PRs in question and the corresponding  tickets reflects the incremental approach. For example: MCR-3731 is a ticket for bug that only was noticeable once the new features in this PR are used. So yes, we could merge MCR-3731 first, but this would mangle the chain-of -thought that lead to these incremental improvements, making it harder to understand the ticket history in the future.

---

With regard to future work on the generic generator:

Adding support for `UUID` is indeed simple.

Adding support for `CreationDate`  and replacing `MCRDNBURNGenerator` (generator based on current date)  comes with some conceptual overhead:
-  Should `CreationDate` and `CurrentDate` share the same date format, of should each have it's own?
-  `MCRFLURNGenerator` limits generation to one identifier per second (b/c the date-format has that granularity). That functionality must somehow be included int the generic pattern.

Adding support for the suggested `$Dependency.DNBURN.2`  is not as straight forward as adding another simple replacement. It comes also comes with some conceptual overhead (design of the replacement patterns and the corresponding configuration properties). 

The configuration mechanism is reworked in MCR-3723 and moved to the annotation based configuration API. New capabilities should be added after MCR-3723, otherwise I would need to include support for the suggested configuration properties with the current approach to configuration just to throw that code away when I rebase MCR-3723.

If `MCRGenericPIGenerator` is going to be the default generator, we shouldn't just add additional patterns/replacements. `MCRGenericPIGenerator` should gain some extensibility, so support for additional patterns/replacements can be added via configuration.  `MCRGenericPIGenerator` should get a configurable list of "replacement providers" (better name needed) that each provide a pattern and a generator for the replacement. The existing patterns/replacements should be moved toward that. Something like

```java
@MCRSentinel
@MCRInstanceMap(name = REPLACEMENT_PROVIDERS, valueClass=ReplacementProvider.class, required = false)
public Map<String, ReplacementProvider> translations;
```

With properties like

```
MCR.PI.Generator.OBJECTDOI.Class=org.mycore.pi.MCRGenericPIGenerator
MCR.PI.Generator.OBJECTDOI.Type=doi
MCR.PI.Generator.OBJECTDOI.GeneralPattern=10.1234/$number-$date
MCR.PI.Generator.OBJECTDOI.ReplacementProviders.number.Class=org.mycore.pi.provider.MCRObjectNumberProvider
MCR.PI.Generator.OBJECTDOI.ReplacementProviders.date.Class=org.mycore.pi.provider.MCRCreationDateProvider
MCR.PI.Generator.OBJECTDOI.ReplacementProviders.date.Formatter.Class=org.mycore.common.date.MCRSimpleDateFormatter
MCR.PI.Generator.OBJECTDOI.ReplacementProviders.date.Formatter.Format=yyyy-MM-dd
```

or

```
MCR.PI.Generator.UUIDDOI.Class=org.mycore.pi.MCRGenericPIGenerator
MCR.PI.Generator.UUIDDOI.Type=doi
MCR.PI.Generator.UUIDDOI.GeneralPattern=10.1234/$uuid
MCR.PI.Generator.UUIDDOI.ReplacementProviders.uuid.Class=org.mycore.pi.provider.MCRUUIDReplacementProvider
```

or

```
MCR.PI.Generator.DNBURNDOI.Class=org.mycore.pi.MCRGenericPIGenerator
MCR.PI.Generator.DNBURNDOI.Type=doi
MCR.PI.Generator.DNBURNDOI.GeneralPattern=10.1234/$urnSuffix
MCR.PI.Generator.DNBURNDOI.ReplacementProviders.urnSuffix.Class=org.mycore.pi.MCROtherPIProvider
MCR.PI.Generator.DNBURNDOI.ReplacementProviders.urnSuffix.Service=DNBURN
MCR.PI.Generator.DNBURNDOI.ReplacementProviders.urnSuffix.Pattern=[a-z0-9:]+-(.+)
```

This approach would also alleviate some of the above mentioned concerns (one vs. multiple formatted: each replacement provider that formats dates would have its own formatter and pattern; configuration of "complex" placeholder names: each `MCROtherPIProvider` has its own name, service and single-capture-group-pattern). 

_The set of replacement provider implementations could also be configured globally, not on a per-generator-basis._

---

If you want, I could also take over responsibility for MCR-3763 (replacing all generators with the generic generator) from you.

---

With regards to the diff sizes. The number of insertions and deletions reported by GitHub are incorrect.

Correct numbers:
- MCR-3723: 44 files, 2278 insertions, 850 deletions
- MCR-3731: 6 files, 69 insertions, 24 deletions
- MCR-3760: 32 files, 1578 insertions, 132 deletions

In both "large" cases: A lot of the insertions are due to license headers, boilerplate and test classes.

I get that reviewing code is also effort and takes time, but I think the situation is less dramatic than the numbers provided by GitHub would suggest..

---

With regards to `Type` and `Service` in the configuration: The new generators are inspired by the existing predicates. Both, `MCRPIOtherPICreatedPredicate` and `MCRPIOtherPIRegisteredPredicate`, have configuration properties `Type` and `Service`, so consequently, if we roll MCR-3731 into this PR, this PR should also fix that, so this PR, which was only intended to add a very simple generator strategy, would than have to

- add that strategy,
- switch the existing predicated over to use flags instead of the database
- fix the configuration options of the existing generators

I'd rather make another follow-up PR.

---

With regards to hardening the loop:

Each auto-creating service should only create at most identifier (once its predicate matches), or none at all, if the predicate never matches. 

I think the following approach should work:

1. Compile list of services for which an identifier exists.
2. Compile list of auto-creating services for which no identifier exists.
3. For each service in that list: If the predicate matches, generate identifier and remove service from list.
4. If any identifier was created, go back to 3.

After some iterations, we either run out of services or no service created an identifier.

If a predicate matches and the service fails to create an identifier, an exception is thrown.

toKrause:
I reordered the PRs, MCR-3731 (switching from database lookup to flag-lookup) now comes first.

Question: If the service flags are the source-of-truth for the PI state, do we need 

```java
.filter(service -> !servicesWithRegisteredIdentifier.contains(service.getServiceID()))
.filter(service -> !MCRPIService.hasFlag(obj, "", service))
```

in `MCRPICreationEventHandler`, or will 

```java
.filter(service -> !MCRPIService.hasFlag(obj, "", service))
```

suffice?

toKrause:
I squashed MCR-3731 and MCR-3800 into this PR.

mycore-pi/src/main/java/org/mycore/pi/util/MCROtherPIValueExtractor.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    public String extractValue(MCRBase base) {
+
+        MCRPIServiceManager serviceManager = MCRPIServiceManager.getInstance();
+        MCRPIService<MCRPersistentIdentifier> service = serviceManager.getRegistrationService(this.service);
+        List<MCRPI> pis = MCRPIService.getFlags(base, "", service);
+
+        if (pis.isEmpty()) {
+            throw new MCRException("No identifier found for " + base + " and service " + service);
```
sebhofmann: probably `this.service` wanted

mycore-pi/src/main/java/org/mycore/pi/util/MCROtherPIValueExtractor.java
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        String identifier = pis.getFirst().getIdentifier();
+        Matcher matcher = pattern.matcher(identifier);
+
+        if (!matcher.find()) {
+            throw new MCRException("Identifier " + identifier + ", found for " + base + " and service "
+                + service + ", doesn't match pattern " + pattern.pattern());
```
sebhofmann: probably `this.service` wanted

mycore-pi/src/main/java/org/mycore/pi/util/MCROtherPIValueExtractor.java
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        String value = matcher.group(1);
+
+        if (value.isEmpty()) {
+            throw new MCRException("Identifier " + identifier + ", found for " + base + " and service "
+                + service + ", contains empty value for pattern " + pattern.pattern());
```
sebhofmann: probably `this.service` wanted

mycore-pi/src/main/java/org/mycore/pi/MCRPICreationEventHandler.java
```diff
@@ -43,26 +48,51 @@ protected void handleObjectUpdated(MCREvent evt, MCRObject obj) {
...
-            .forEach((serviceToRegister) -> {
+            for (MCRPIService<MCRPersistentIdentifier> service : matchingAutoCreatingServices) {
                 try {
-                    serviceToRegister.register(obj, "", false);
+
+                    // register new PI, succeeds or throws exception
+                    MCRPersistentIdentifier identifier = service.register(obj, "", false);
+
+                    if (LOGGER.isInfoEnabled()) {
+                        LOGGER.info("Created new identifier {} for object  {} using auto-creating PI service {}",
```
sebhofmann: double space between `object` and `{}`

sebhofmann (changes requested):

toKrause:
Fixed issues and rebased to current main.

sebhofmann (approved):

PR #2983, opened 2026-06-14
MCR-3723 Modernize and harmonize PI generator configuration, add unit tests

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3723).

## Ticket & Documentation

- [X] Does this change affect existing applications, data, or configurations?

## Testing
- [X] Were existing tests modified?
  - [X] Yes: explain the changes for reviewers.

## MCR Conventions & Metadata
- [X] If the **public API** has changed:
- [X] Javadoc is written for non-self-explanatory classes/methods (Clean Code).
- [X] All configuration options are documented in Javadoc and `mycore.properties`.

## Multi-Repo Considerations
- [X] Is an equivalent PR in `MIR` required?
  - [X] If yes, is it already created?

**To be used in conjunction with https://github.com/MyCoRe-Org/mir/pull/1390.**


toKrause:
@sebhofmann As requested, I moved the data-formatting-stuff into a separate PR: https://github.com/MyCoRe-Org/mycore/pull/3040

mycore-pi/src/main/java/org/mycore/pi/MCRGenericPIGenerator.java
```diff
@@ -300,7 +323,8 @@ private String applyCount(String resultingPI) {
             LOGGER.info("Count is {}", count);
             final String pattern = IntStream.range(0, Math.abs(countPrecision)).mapToObj((i) -> "0")
```
sebhofmann: This method should use `formatCount`.

mycore-pi/src/main/java/org/mycore/pi/doi/MCRCurrentDateDOIGenerator.java
```diff
@@ -0,0 +1,95 @@
...
+
+    private String lastSuffix;
+
+    public MCRCurrentDateDOIGenerator(MCRDOIParser parser, String prefix) {
+        super(parser);
+        this.prefix = Objects.requireNonNull(prefix, "Prefix must not be null");
+    }
+
+    @Override
+    protected String buildDOI(MCRBase base, String additional) {
```
sebhofmann: Should be synchronized

mycore-pi/src/main/java/org/mycore/pi/urn/MCRCreateDateDNBURNGenerator.java
```diff
@@ -0,0 +1,155 @@
...
+ * Example:
+ * <pre><code>
+ * [...].Class=org.mycore.pi.urn.MCRCreateDateDNBURNGenerator
+ * [...].DateFormat=yyyy-MM-dd
+ * [...].Prefix=10.1234
+ * [...].CountPrecision=6
+ * </code></pre>
+ */
+@MCRConfigurationProxy(proxyClass = MCRCreateDateDNBURNGenerator.Factory.class)
+public class MCRCreateDateDNBURNGenerator extends MCRDNBURNGeneratorBase {
```
sebhofmann: I don't think we need that. The `MCRGenericPIGenerator` can do that.

mycore-pi/src/main/java/org/mycore/pi/doi/MCRCreateDateDOIGenerator.java
```diff
@@ -18,96 +18,127 @@
...
+ * [...].Class=org.mycore.pi.doi.MCRCreateDateDOIGenerator
+ * [...].DateFormat=yyyy-MM-dd
+ * [...].Prefix=10.1234
+ * [...].CountPrecision=6
+ * </code></pre>
+ */
+@MCRConfigurationProxy(proxyClass = MCRCreateDateDOIGenerator.Factory.class)
+public class MCRCreateDateDOIGenerator extends MCRDOIGeneratorBase {
+
+    public static final String DEFAULT_DATE_FORMAT = "yyyyMMdd-HHmmss";
```
sebhofmann: I think we should just deprecate that thing, because `MCRGenericPIGenerator` can do that.

mycore-pi/src/main/java/org/mycore/pi/doi/MCRCurrentDateDOIGenerator.java
```diff
@@ -0,0 +1,95 @@
...
+ * specify the prefix.
+ * </ul>
+ * Example:
+ * <pre><code>
+ * [...].Class=org.mycore.pi.doi.MCRCurrentDateDOIGenerator
+ * [...].Prefix=10.1234
+ * </code></pre>
+ */
+@MCRConfigurationProxy(proxyClass = MCRCurrentDateDOIGenerator.Factory.class)
+public class MCRCurrentDateDOIGenerator extends MCRDOIGeneratorBase {
```
sebhofmann: Is there really anyone using it?

sebhofmann (changes requested):

mycore-pi/src/main/java/org/mycore/pi/urn/MCRCreateDateDNBURNGenerator.java
toKrause: Yes, simply added for symmetry, because `MCRCreateDateDOIGenerator` exists.

mycore-pi/src/main/java/org/mycore/pi/doi/MCRCurrentDateDOIGenerator.java
```diff
@@ -0,0 +1,95 @@
...
+
+import org.mycore.common.config.annotation.MCRConfigurationProxy;
+import org.mycore.common.config.annotation.MCRProperty;
+import org.mycore.datamodel.metadata.MCRBase;
+import org.mycore.pi.MCRPIGenerator;
+import org.mycore.pi.urn.MCRDNBURN;
+import org.mycore.pi.util.MCRFLDateScrambler;
+
+/**
+ * {@link MCRCurrentDateDOIGenerator} is a {@link MCRPIGenerator} for {@link MCRDNBURN} identifiers
```
sebhofmann: Its for DOI not for URN

mycore-pi/src/main/java/org/mycore/pi/doi/MCRCreateDateDOIGenerator.java
toKrause: True, I will make this a small agenda item in today's web-call. Either
- keep `MCRCreateDateDOIGenerator` and add `MCRCreateDateDNBURNGenerator` or
- deprecate  `MCRCreateDateDOIGenerator` and drop `MCRCreateDateDNBURNGenerator`

mycore-pi/src/main/java/org/mycore/pi/doi/MCRCurrentDateDOIGenerator.java
toKrause: good point

mycore-pi/src/main/java/org/mycore/pi/MCRGenericPIGenerator.java
```diff
@@ -48,224 +46,258 @@
...
-
-    static final String PLACE_HOLDER_CURRENT_DATE = "$CurrentDate";
+@MCRConfigurationProxy(proxyClass = MCRGenericPIGenerator.Factory.class)
+public class MCRGenericPIGenerator implements MCRPIGenerator<MCRPersistentIdentifier> {
 
-    static final String PLACE_HOLDER_OBJECT_DATE = "$ObjectDate";
+    public static final String DEFAULT_DATE_FORMAT = "ddMMyyyy";
 
-    static final String PLACE_HOLDER_OBJECT_TYPE = "$ObjectType";
+    public static final Locale DFAULT_DATE_LOCALE = Locale.ROOT;
```
sebhofmann: TYPO

mycore-pi/src/main/java/org/mycore/pi/doi/MCRCurrentDateDOIGenerator.java
toKrause: I don't know, simply added for symmetry, because `MCRCurrentDateDNBURNGenerator` (formally kown as `MCRFLURNGenerator`) exists.

mycore-pi/src/main/java/org/mycore/pi/MCRGenericPIGenerator.java
toKrause: fixed

mycore-pi/src/main/java/org/mycore/pi/doi/MCRCurrentDateDOIGenerator.java
toKrause: fixed

toKrause:
With regards to removing or deprecating specialized generators:

- `MCRCreateDateDOIGenerator` / `MCRCreateDateDNBURNGenerator` can not be replaced with `MCRGenericPIGenerator` because the generic generator only has a placeholder for the current date, not the creation date.
- `MCRCurrentDateDOIGenerator` / `MCRCurrentDateDNBURNGenerator` can not be replaced with `MCRGenericPIGenerator` due to the vastly differing date formatting strategies; this would require [MCR-3670](https://github.com/MyCoRe-Org/mycore/pull/3040), which is a follow-up to this PR.

Therefore, I suggest the following course of action:
- merge this PR as is (keeps the code base backwards compatible and consistent)
- merge MCR-3760 (keeps the code base backwards compatible and makes it more flexible)
- create PR for [MCR-3763](https://mycore.atlassian.net/browse/MCR-3763) that enhances the generic generator and deprecates the specialized generators

[MCR-3670]: https://mycore.atlassian.net/browse/MCR-3670?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MCR-3763]: https://mycore.atlassian.net/browse/MCR-3763?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

mycore-pi/src/main/java/org/mycore/pi/doi/MCRCreateDateDOIGenerator.java
toKrause: Not true actually, `MCRGenericPIGenerator` only supports the current date, not the object creation date.

mycore-pi/src/test/java/org/mycore/pi/MCRGenericPIGeneratorTest.java
```diff
@@ -44,33 +50,76 @@
 })
 public class MCRGenericPIGeneratorTest {
 
-    public static final int CURRENT_YEAR = Calendar.getInstance().get(Calendar.YEAR);
+    public static final String DATE_FORMAT = "mmDDyyyy";
```
sebhofmann:
```java
public static final String DATE_FORMAT = "mmDDyyyy";
```

In `SimpleDateFormat`, `mm` is minutes and `DD` is day-of-year, so this pattern is
neither month nor day-of-month. For 2026-08-17 14:03:09 it yields:

```
mmDDyyyy  -> 032292026     minute 03, day 229 of year, 2026
MMddyyyy  -> 08172026      month 08, day 17, 2026
```

If month/day/year is intended, the case is inverted; it should be `MMddyyyy`.
The current pattern also produces nine characters instead of eight.

The test still passes because it builds the expected value with the same pattern it
passes to the generator, so it would also accept a genuinely broken date formatting.
Consider asserting against a literal expected string instead.

Note: the default in `mycore.properties`
(`MCR.Default.PI.Generator.Generic.DateFormat=ddMMyyyy`) is correct; only the test
constant is affected. `DATE_FORMAT` is new on this branch, `main` used an inline
`new SimpleDateFormat("yyyy", Locale.ROOT)`.

mycore-pi/src/main/java/org/mycore/pi/urn/MCRUUIDDNBURNGenerator.java
```diff
@@ -0,0 +1,78 @@
...
+ * </ul>
+ * Example:
+ * <pre><code>
+ * [...].Class=org.mycore.pi.urn.MCRUUIDDNBURNGenerator
+ * [...].Namespace=urn:nbn:de:gbv:xyz
+ * [...].Delimiter=-
+ * </code></pre>
+ */
+@MCRConfigurationProxy(proxyClass = MCRUUIDDNBURNGenerator.Factory.class)
+public class MCRUUIDDNBURNGenerator extends MCRDNBURNGeneratorBase {
```
sebhofmann: `MCRUUIDDNBURNGenerator` is a new class that carries no behaviour the existing
`MCRUUIDURNGenerator` did not already have. It exists only to mirror the naming of
`MCRUUIDDOIGenerator`. Per the agreement in the call, no new generator classes should
be introduced by mirroring URN, DOI or PURL.

What is left over is a pure duplicate: `MCRUUIDURNGenerator` now extends
`MCRUUIDDNBURNGenerator`, adds no method, and repeats the identical `Factory` with the
same two `@MCRProperty` fields (`Namespace`, `Delimiter`). The deprecation also points
the wrong way, deprecating the established name that installations already have in
`MCR.PI.Generator.*.Class` in favour of a name that did not exist until this branch.

Suggestion: drop `MCRUUIDDNBURNGenerator` and modernize `MCRUUIDURNGenerator` in place,
i.e. let it extend `MCRDNBURNGeneratorBase` and carry the `Factory` itself. No
deprecation is then needed, because the class name does not change.

Places to adjust when removing it:

- `mycore-pi/src/test/java/org/mycore/pi/MCRPIUtils.java:35` and `:61`
- `mycore-pi/src/test/java/org/mycore/pi/urn/rest/MCRURNGranularRESTServiceTest.java:61` and `:85`
- rename `MCRUUIDDNBURNGeneratorTest` back to `MCRUUIDURNGeneratorTest`

mycore-pi/src/main/java/org/mycore/pi/doi/MCRCreateDateDOIGenerator.java
```diff
@@ -18,96 +18,121 @@
...
+
+        @MCRProperty(name = PREFIX_KEY, defaultName = "MCR.DOI.Prefix")
+        public String prefix;
+
+        @MCRProperty(name = COUNT_PRECISION_KEY, defaultName = DEFAULT_PROPERTY_PREFIX + COUNT_PRECISION_KEY)
+        public String countPrecision = "-1";
+
+        @Override
+        public MCRCreateDateDOIGenerator get() {
+            return new MCRCreateDateDOIGenerator(new MCRDOIParser(),dateFormat, prefix,
```
sebhofmann: Missing space after the comma, the Eclipse formatter has not been run over this:

```java
return new MCRCreateDateDOIGenerator(new MCRDOIParser(),dateFormat, prefix,
```

Trailing whitespace, reported by `git diff main...HEAD --check`:

- `mycore-pi/src/main/java/org/mycore/pi/doi/MCRMapObjectIDDOIGenerator.java:43`
- `mycore-pi/src/main/java/org/mycore/pi/urn/MCRFLURNGenerator.java:51`

Duplicated package prefix in the Javadoc example at
`mycore-pi/src/main/java/org/mycore/pi/MCRGenericPIGenerator.java:125`:

```
[...].Class=org.mycore.org.mycore.pi.MCRGenericPIGenerator
```

should be `org.mycore.pi.MCRGenericPIGenerator`.

mycore-pi/src/test/java/org/mycore/pi/doi/MCRMapObjectIDDOIGeneratorTest.java
```diff
@@ -20,60 +20,77 @@
...
+
+    @Test
+    public void generate() throws MCRPersistentIdentifierException {
+
+        MCRObject object = new MCRObject();
+        object.setSchema("http://www.w3.org/2001/XMLSchema");
+        object.setId(MCRObjectID.getInstance("my_test_00000123"));
+
+        Map<String, String> prefixMap = Map.of("my_test", PREFIX);
+        MCRMapObjectIDDOIGenerator generator = new MCRMapObjectIDDOIGenerator(new MCRDOIParser(), prefixMap);
```
sebhofmann, line 53: On `main`, three test classes instantiated generators through the configuration and
therefore exercised the property wiring:

- `MCRMapObjectIDDOIGeneratorTest:51`, with `MCR.PI.Generator.MapObjectIDDOI.Prefix.junit_test`
  and `.Prefix.my_test`
- `MCRIDPURLGeneratorTest:61` and `:66`, with `.BaseURLTemplate` in two variants
- `MCRDNBURNGeneratorTest:62`, with `.Namespace`, covering the stripping of the
  `urn:nbn:de:` prefix

On this branch all of them construct the generator directly and pass the values as
constructor arguments. The only remaining `MCRConfiguration2.getInstanceOfOrThrow` call
in `mycore-pi/src/test` is `MCRPIManagerTest:181` for the mock generator, and that one
lost its properties together with `MCRMockIdentifierGenerator.TEST_PROPERTY`.

That leaves the actual subject of this ticket without any coverage: none of the new
`Factory` classes, no `@MCRProperty` name, no `defaultName` fallback and no
`@MCRPropertyMap` is touched by a test. A typo in a property name, a wrong
`defaultName` or a missing default in `mycore.properties` would not fail the build.

Suggestion: keep the direct constructor tests for the generation logic, but add at
least one configuration driven test per generator that goes through
`MCRConfiguration2.getInstanceOfOrThrow(...)` with `@MCRTestProperty`.

sebhofmann (commented):
Review of the generator changes. Four inline remarks, none of them blocking on their own.

The most substantial one is the loss of configuration coverage in the tests: the property wiring that this ticket rewrites is no longer exercised anywhere.

Separately, regarding the agreement from the call that no new generator classes should be added by mirroring URN, DOI or PURL: `MCRCurrentDateDOIGenerator`, `MCRCreateDateDNBURNGenerator` and `MCRCurrentDateDNBURNGenerator` are gone, which covers most of it. `MCRUUIDDNBURNGenerator` still looks like a pure mirror, see the inline comment. The two `MCROtherPI*` generators I would keep: they share their logic in `MCROtherPIValueExtractor`, and DOI-from-URN and URN-from-DOI are two real configurations rather than one feature duplicated.

mycore-pi/src/test/java/org/mycore/pi/MCRGenericPIGeneratorTest.java
toKrause: I fixed the date formatting (using the configured default date pattern). Since the test pattern uses `$CurrentDate`, we can't write the test against a literal expected string.

mycore-pi/src/main/java/org/mycore/pi/urn/MCRUUIDDNBURNGenerator.java
toKrause: `MCRUUIDDNBURNGenerator`  would not have been a _new_ generator. `MCRUUIDURNGenerator` was merely renamed into `MCRUUIDDNBURNGenerator` as part of the PR (in a backwards compatible name).

Since `MCRUUIDURNGenerator` is on the chopping board anyway, I undid this improvement (on the understanding that `MCRUUIDURNGenerator` will be made redundant by improvements to the generic generator during the development cycle for the next release).

mycore-pi/src/main/java/org/mycore/pi/doi/MCRCreateDateDOIGenerator.java
toKrause: Fixed.

mycore-pi/src/test/java/org/mycore/pi/doi/MCRMapObjectIDDOIGeneratorTest.java
toKrause: I'll add additional tests following the style introduced with all the `MCRResource[Provider|Locator|Filter|Selector]*`-classes.

mycore-pi/src/main/java/org/mycore/pi/util/MCRPIGeneratorUtils.java
```diff
@@ -0,0 +1,153 @@
...
+         * @param type   the type of persistent identifier.
+         * @param pattern a regex pattern which can be used to extract count value in existing identifiers.
+         *                The first capturing group captures the count.
+         *                Example: <code>[0-9]+-mods-2017-([0-9][0-9][0-9][0-9])-[0-9]</code>
+         *                will match <code>31-mods-2017-0003-3</code> and the returned count should be <code>4</code>
+         *                (<code>3+1</code>).
+         *
+         * @return the next count
+         */
+        int getCount(String pattern, String type);
```
sebhofmann: The parameter order of this method is swapped relative to its own Javadoc (`@param type` first, `@param pattern` second), and two of the three call sites follow the Javadoc rather than the signature:

- `MCRCountingDNBURNGeneratorBase:57`: `counter.getCount(MCRDNBURN.TYPE, pattern)`
- `MCRCreateDateDOIGenerator:98`: `counter.getCount(MCRDigitalObjectIdentifier.TYPE, Pattern.quote(prefixWithDate) + countPattern)`
- `MCRGenericPIGenerator:299`: `counter.getCount(counterPattern, type)` (matches the signature)

For the first two, `CachingDatabaseCounter.getCount` ends up calling `MCRPIManager.getInstance().getList(<regex>, -1, -1)` with the regex as the identifier type. That query never matches anything, so the counter is initialized to 0 on every restart instead of continuing after the highest existing count. On `main` both generators queried the database with `MCRDNBURN.TYPE` / `MCRDigitalObjectIdentifier.TYPE`, so this is a regression that can produce duplicate URNs and DOIs after a restart.

As a second effect, the cache key becomes the constant `"dnbUrn"` / `"doi"` instead of the pattern, so all patterns of one type share a single counter.

The new tests do not catch this, because `MCRMockCounter#getCount` ignores both arguments.

Suggestion: change the signature to `getCount(String type, String pattern)`, so that it matches the Javadoc and both call sites above, and adjust `MCRGenericPIGenerator:299` accordingly.

mycore-pi/src/main/java/org/mycore/pi/util/MCRPIGeneratorUtils.java
```diff
@@ -0,0 +1,153 @@
...
+
+    }
+
+    /**
+     * {@link Counter} that gets the count for a specific type and pattern and increase the internal counter.
+     * If there is no internal counter, it will look into the Database and detect the highest count with the pattern.
+     */
+    public static final class CachingDatabaseCounter implements Counter {
+
+        private static final Map<String, AtomicInteger> PATTERN_COUNT_MAP = new HashMap<>();
```
sebhofmann: `PATTERN_COUNT_MAP` is `static`, so it is shared by every `CachingDatabaseCounter` instance, whereas on `main` each generator class had its own map. Combined with a key that contains only the pattern and not the type, two generators of different types that happen to use the same pattern now share one counter.

Once the argument order is fixed, the key should be derived from both type and pattern.

mycore-pi/src/main/java/org/mycore/pi/MCRGenericPIGenerator.java
```diff
@@ -279,29 +287,21 @@ public MCRPersistentIdentifier generate(MCRBase mcrBase, String additional)
...
-            final int count = getCount(counterPattern);
+            final int count = counter.getCount(counterPattern, type);
             LOGGER.info("Count is {}", count);
             final String pattern = IntStream.range(0, Math.abs(countPrecision)).mapToObj((i) -> "0")
                 .collect(Collectors.joining(""));
-            DecimalFormat decimalFormat = new DecimalFormat(pattern, DecimalFormatSymbols.getInstance(Locale.ROOT));
+            DecimalFormat decimalFormat =
+                new DecimalFormat(pattern, DecimalFormatSymbols.getInstance(Locale.ROOT));
             final String countAsString = countPrecision != -1 ? decimalFormat.format(count) : String.valueOf(count);
             result = resultingPI.replace(PLACE_HOLDER_COUNT, countAsString);
```
sebhofmann: This block still builds the count string itself, so https://github.com/MyCoRe-Org/mycore/pull/2983#discussion_r3577712089 is still open on the current head (`2c18cf95`). Opening a new thread here, because the old one is anchored to a single line and does not show the relevant code.

`MCRPIGeneratorUtils.formatCount` already encapsulates exactly this logic, including the check that throws an `MCRPersistentIdentifierException` when the count needs more digits than the configured precision. The inline variant here silently produces an identifier that is longer than configured instead.

`MCRCreateDateDOIGenerator:100` already uses `formatCount`, so the two code paths are inconsistent as well.

toKrause: that got lost among the other comments, fixed

mycore-pi/src/main/java/org/mycore/pi/util/MCRPIGeneratorUtils.java
toKrause: In the current implementation, we could theoretically have two different generators with patterns that happen to be equal that could generate the same PI:

```java
    @Test
    @MCRTestConfiguration(properties = { @MCRTestProperty(key = "MCR.DOI.Prefix", string = "10.1234") })
    public void test() throws MCRPersistentIdentifierException {

        MCRObject object1 = new MCRObject();
        object1.setSchema("http://www.w3.org/2001/XMLSchema");
        object1.setId(MCRObjectID.getInstance("my_test_00000123"));

        MCRCreateDateDOIGenerator generator1 = new MCRCreateDateDOIGenerator();

        String doi1 = generator1.generate(object1, "").asString();
        System.out.println(doi1);

        MCRObject object2 = new MCRObject();
        object2.setSchema("http://www.w3.org/2001/XMLSchema");
        object2.setId(MCRObjectID.getInstance("my_test_00000234"));

        MCRGenericPIGenerator generator2 = new MCRGenericPIGenerator(
            MCRConfiguration2.getStringOrThrow("MCR.DOI.Prefix") + "/$ObjectDate-$Count",
            new SimpleDateFormat("yyyyMMdd-HHmmss", Locale.ROOT),
            null, null, -1, MCRDigitalObjectIdentifier.TYPE);

        String doi2 = generator2.generate(object2, "").asString();
        System.out.println(doi2);

        assertNotEquals(doi1, doi2); // fails

    }
```

Having a shared (static) cache for then shared database across all `CachingDatabaseCounter` would, in theory, solve this very theoretical problem. However, with the current implementations this protection doesn't actually kick in, because `MCRCreateDateDOIGenerator` now protects the non-counter-part of the pattern with `Pattern#quote` (resulting pattern: `\Q10.1234/19082026-\E([0-9][0-9][0-9])`), while there is no such protection in the generic generator (resulting pattern: `10.1234/19082026-([0-9][0-9][0-9])`). Since the resulting patterns are not equal, the map has one count for each pattern.

Since the pattern covers the entire PI, I don't (currently) see a problem with the type-independent cache.

toKrause: Therefore, I'll return to the non-shared, caching,

toKrause: Correction, I've added proper protection to the generic generator, made the cache-sharing more explicit and added a corresponding test.

mycore-pi/src/main/java/org/mycore/pi/util/MCRPIGeneratorUtils.java
toKrause: Good catch, fixed. I thought I used IntelliJs "Change signature"-feature, but maybe I didn't, or the refactoring wasn't applied entirely :thinking:

mycore-pi/src/main/java/org/mycore/pi/util/MCRPIGeneratorUtils.java
```diff
@@ -0,0 +1,154 @@
...
+     */
+    public static final class CachingDatabaseCounter implements Counter {
+
+        private final Map<String, AtomicInteger> patternCountMap = new HashMap<>();
+
+        @Override
+        public synchronized int getCount(String type, String pattern) {
+            return this.patternCountMap
+                .computeIfAbsent(pattern, p -> readCountFromDatabase(type, p))
+                .getAndIncrement();
```
sebhofmann, line 121: The map is no longer `static`, but with `SHARED_COUNTER` there is exactly one instance for all generators, so the cache is still global. That is intentional now and covered by `MCRSharedDatabaseCounterTest`, no objection to that.

What is left is that the key is the pattern alone. `type` only takes effect on a cache miss, in `readCountFromDatabase`. Whichever type registers a pattern first therefore decides which type the database was scanned for, and every later type inherits that entry.

In practice DOI and URN patterns differ in their prefix, so this is unlikely to be hit, but since the counter instance is now deliberately shared across types, I would rather not rely on that. Deriving the key from type and pattern keeps the intended sharing within one type and rules out mixing between types.

mycore-pi/src/main/java/org/mycore/pi/util/MCRPIGeneratorUtils.java
```diff
@@ -0,0 +1,154 @@
...
+        /**
+         * Returns the next counter value to be used for a specific type and pattern. The returned value
+         * needs to be unique. Returned values should be increasing.
+         * @param type    the type of persistent identifier.
+         * @param pattern a regex pattern which can be used to extract count value in existing identifiers.
+         *                The first capturing group captures the count.
+         *                Example: <code>[0-9]+-mods-2017-([0-9][0-9][0-9][0-9])-[0-9]</code>
+         *                will match <code>31-mods-2017-0003-3</code> and the returned count should be <code>4</code>
+         *                (<code>3+1</code>). All non-variable parts of the pattern should be protected with
+         *                {@link Pattern#quote(String)}, in order to make pattens comparable.
```
sebhofmann: Typo: `pattens` -> `patterns`.

mycore-pi/src/main/java/org/mycore/pi/MCRGenericPIGenerator.java
```diff
@@ -276,151 +279,70 @@ public MCRPersistentIdentifier generate(MCRBase mcrBase, String additional)
...
+                    counterPattern.append(Pattern.quote(resultingPI.substring(start, index)));
+                }
+                counterPattern.append(countPattern);
+                start = index + PLACE_HOLDER_COUNT.length();
+                index = resultingPI.indexOf(PLACE_HOLDER_COUNT, index + 1);
+            }
+            if (start != resultingPI.length()) {
+                counterPattern.append(Pattern.quote(resultingPI.substring(start)));
+            }
+
```
sebhofmann: Line 301 consists of four spaces, `git diff main...HEAD --check` reports it as trailing whitespace. The two spots from my earlier review are still there as well:

- `mycore-pi/src/main/java/org/mycore/pi/doi/MCRMapObjectIDDOIGenerator.java:43`
- `mycore-pi/src/main/java/org/mycore/pi/urn/MCRFLURNGenerator.java:51`

Running the Eclipse formatter over the changed files takes care of all three.

Minor, on line 296: `indexOf(PLACE_HOLDER_COUNT, index + 1)` gives the same result as `indexOf(PLACE_HOLDER_COUNT, start)` only because `$Count` cannot overlap itself. Searching from `start` expresses the intent directly and does not depend on that property.

sebhofmann (commented):
Thanks, the parameter order and `formatCount` are fixed, I checked every call site of `Counter#getCount` and they are consistent now. I also walked through the new quoting loop for `$Count` at the start, at the end, multiple occurrences and with trailing text, that part looks correct.

A few things are left, mostly cosmetic, except for the counter cache key.

mycore-pi/src/main/java/org/mycore/pi/urn/MCRCountingDNBURNGeneratorBase.java
```diff
@@ -0,0 +1,58 @@
...
+     * look into the Database and detect the highest count with the pattern.
+     *
+     * @param pattern a regex pattern which will be used to detect the highest count. The first group is the count.
+     *                e.G. [0-9]+-mods-2017-([0-9][0-9][0-9][0-9])-[0-9] will match 31-mods-2017-0003-3 and the returned
+     *                count will be 4 (3+1).
+     * @return the next count
+     */
+    public final synchronized int getCount(String pattern) {
+        return counter.getCount(MCRDNBURN.TYPE, pattern);
+    }
```
sebhofmann, line 59: Two things in this block:

Line 38: `{@link MCRCountingDNBURNGeneratorBase(MCRPIGeneratorUtils.Counter, String, String)}` is not a valid Javadoc reference, a constructor link needs the `#` form, i.e. `{@link MCRCountingDNBURNGeneratorBase#MCRCountingDNBURNGeneratorBase(MCRPIGeneratorUtils.Counter, String, String)}`.

Lines 45-53: the Javadoc does not mention the quoting requirement that was just added to `Counter#getCount`. Since this class now writes into the same `SHARED_COUNTER`, a subclass in an application that passes its pattern unquoted gets a second, independent counter for the very same identifier space. Either document it here as well, or quote the fixed parts in this class.

mycore-pi/src/test/java/org/mycore/pi/MCRSharedDatabaseCounterTest.java
```diff
@@ -0,0 +1,101 @@
...
+    public static final String PREFIX = "10.1234";
+
+    /**
+     * Tests that two otherwise independent PI generators that are configured to produce
+     * similar PIs and are using the shared {@link CachingDatabaseCounter}
+     * do indeed produce different PIs that only differ in the applied count value.
+     */
+    @Test
+    @MCRTestConfiguration(properties = {
+        @MCRTestProperty(key = "MCR.DOI.Prefix", string = "10.1234")
```
sebhofmann: `MCR.DOI.Prefix` is never read in this test. Both generators are constructed directly, so the prefix comes from the `PREFIX` constant, and the `defaultName` fallback that consumes this property only runs when the generator is created through the `Factory` via `MCRConfiguration2`. `MCRDOIParser` does not read it either.

Either drop the line, or create `generator1` through `MCRConfiguration2.getInstanceOfOrThrow` so the property is actually used. The latter would also cover the `defaultName` fallback, which is otherwise untested.

mycore-pi/src/main/java/org/mycore/pi/util/MCRPIGeneratorUtils.java
toKrause: fixed

mycore-pi/src/main/java/org/mycore/pi/MCRGenericPIGenerator.java
toKrause: the Eclipse formatter plugin in IntelliJ does not remove trailing spaces in comments - or some other configuration in IntelliJ prevents the formatter from removing such space. fixed manually

toKrause: also improved the `indexOf` call as suggested

mycore-pi/src/main/java/org/mycore/pi/urn/MCRCountingDNBURNGeneratorBase.java
toKrause: link worked fine in IntelliJ but may be non-standard. I think `{@link #MCRCountingDNBURNGeneratorBase(MCRPIGeneratorUtils.Counter, String, String)}` w/o the redundant `MCRCountingDNBURNGeneratorBase` should also work

mycore-pi/src/test/java/org/mycore/pi/MCRSharedDatabaseCounterTest.java
toKrause: dropped the `MCRTestProperty`, so the test case is more concise. I don't think we need to test every possible configuration path, otherwise we would  have to add many tests, and the annotation based configuration API already has extensive tests

mycore-pi/src/main/java/org/mycore/pi/MCRGenericPIGenerator.java
```diff
@@ -276,151 +279,72 @@ public MCRPersistentIdentifier generate(MCRBase mcrBase, String additional)
...
+                    counterPattern.append(Pattern.quote(resultingPI.substring(start, index)));
+                }
+                counterPattern.append(countPattern);
+                start = index + PLACE_HOLDER_COUNT.length();
+                index = resultingPI.indexOf(PLACE_HOLDER_COUNT, start);
+            }
+            if (start != resultingPI.length()) {
+                counterPattern.append(Pattern.quote(resultingPI.substring(start)));
+            }
+
```
sebhofmann: This one survived, it just moved from 301 to 303 because of the two new comment lines. What looks like a blank line here consists of four spaces:

```
$ git diff origin/main...54e8e312bb --check
mycore-pi/src/main/java/org/mycore/pi/MCRGenericPIGenerator.java:303: trailing whitespace.
+    
```

The two Javadoc spots are gone, thanks. This one is inside a method body rather than in a comment, so the IntelliJ behaviour you described should not apply here.

sebhofmann (commented):
The `Key` record settles the counter cache question, and the `indexOf`, Javadoc link, quoting note, typo and test property changes all look good to me.

Three small leftovers.

mycore-pi/src/main/java/org/mycore/pi/MCRGenericPIGenerator.java
```diff
@@ -276,151 +279,72 @@ public MCRPersistentIdentifier generate(MCRBase mcrBase, String additional)
...
-                regexpStr = "("
-                    + IntStream.range(0, countPrecision).mapToObj((i) -> "[0-9]").collect(Collectors.joining(""))
-                    + ")";
-            }
 
-            String counterPattern = resultingPI.replace(PLACE_HOLDER_COUNT, regexpStr);
-            if (getType().equals("dnbUrn")) {
-                counterPattern = counterPattern + "[0-9]";
+            // replace all occurrences of the count placeholder with the count pattern, and quote
+            // all other parts, os they are interpreted as literals in the generated pattern
```
sebhofmann: Typo in the new comment: `os they are interpreted` -> `so they are interpreted`.

mycore-pi/src/main/java/org/mycore/pi/util/MCRPIGeneratorUtils.java
```diff
@@ -0,0 +1,157 @@
...
+     * If there is no internal counter, it will look into the Database and detect the highest count with the pattern.
+     */
+    public static final class CachingDatabaseCounter implements Counter {
+
+        private final Map<Key, AtomicInteger> typePatternCountMap = new HashMap<>();
+
+        @Override
+        public synchronized int getCount(String type, String pattern) {
+            return this.typePatternCountMap
+                .computeIfAbsent(new Key(type, pattern), k -> readCountFromDatabase(type, k.pattern))
```
sebhofmann: `k.pattern` reads the private record field directly instead of going through the accessor. It compiles because both classes are in the same nest, but it is inconsistent with `type`, which comes from the enclosing scope.

Either `readCountFromDatabase(k.type(), k.pattern())` or simply `readCountFromDatabase(type, pattern)`.

sebhofmann (approved):

PR #2988, opened 2026-06-17
MCR-3726 fix MCRBuildXMLURIResolver dropping _rootName_ and parameters

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3726).

`MCRBuildXMLURIResolver` parsed the full `href` (including the `buildxml:` scheme prefix) instead of the query part. As a result the first parameter key became e.g. `buildxml:_rootName_` instead of `_rootName_`, so `_rootName_` was no longer recognized and the built XML was wrong (only the bogus first child was returned). Regression from the URI resolver refactoring (MCR-3688). In MIR this broke the metadata import build chain, so DOI/PPN imports lost all metadata (MIR-1600).

Fix: parse `key` (scheme stripped) instead of `href`.

## Ticket & Documentation
- [x] The `fixed-version` is correctly set in the ticket and matches the PR's target branch. **Note: target branch is `2026.06.x`, not `main`.**

## Bugfix-Specific Checks

## Testing
- [x] The feature behaves as described in the ticket (DOI import works again).
- [ ] Were existing tests modified? No, a new regression test was added.

## MCR Conventions & Metadata

## Multi-Repo Considerations
- [x] Is an equivalent PR in `MIR` required? No — the bug is in MyCoRe; MIR is fixed transitively.

### DIFF
--- a/mycore-base/src/main/java/org/mycore/common/xml/MCRBuildXMLURIResolver.java
+++ b/mycore-base/src/main/java/org/mycore/common/xml/MCRBuildXMLURIResolver.java
@@ -74,7 +74,7 @@ public class MCRBuildXMLURIResolver implements URIResolver {
         String key = href.substring(href.indexOf(':') + 1);
         LOGGER.debug("Building xml from {}", key);
 
-        Map<String, String> params = MCRURIResolverHelper.parseQueryParameters(href);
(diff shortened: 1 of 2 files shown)


kkrebs (approved):
tested locally with MIR -> DOI and PPN import works again. Thanks!

PR #3013, opened 2026-06-25
MCR-3670 Harmonize error messages in case of missing .Class properties

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3670). Follow-Up.


rsteph-de (approved):
tested locally - works for me!

PR #3017, opened 2026-06-26
MCR-3747 RestApiv1 vue-i18n-messages

rsteph-de:

## Ticket & Documentation

- [x] For UI changes: before & after screenshots are attached.

## Bugfix-Specific Checks
- [ ] Affected version is listed in the ticket.
- [ ] This PR truly fixes **only** the reported bug.

## Testing

## MCR Conventions & Metadata

## Multi-Repo Considerations

### DIFF
--- a/mycore-restapi/src/main/java/org/mycore/restapi/v1/MCRRestAPIMessages.java
+++ b/mycore-restapi/src/main/java/org/mycore/restapi/v1/MCRRestAPIMessages.java
@@ -73,6 +73,8 @@ public class MCRRestAPIMessages {
 
     public static final String FORMAT_XML = "xml";
 
+    //"MCRApiDraft" - the vue-i18n output is experimental.
+    //It might change or be removed in the future
     public static final String FORMAT_VUE_I18N = "vue-i18n";
 
     public static final String FORMAT_PROPERTY = "property";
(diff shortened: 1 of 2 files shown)


golsch:
My opinion: The API should provide locale messages as plain framework agnostic JSON and the client should transform them into the required format.
Furthermore it does not make sense to make adjustments to the deprecated v1 API. Additionally there is a JWT resource for Lang.

rsteph-de:
Thanks for your feedback, some of your points I considered during implementation...
- **plain framework agnostic JSON + client side transformation**
  - We decided for Vue.js is as our primary frontend framework,
     so it makes sense to support the Vue.js favoured message format
  - I wanted to keep the client side simple
  - Otherwise we would have to provide Javascript util functions
- **deprecated v1 API**
  - there is actually nobody working on migrating the messages api endpoint
  - (objects and classification endpoints should go first)
 - **JWT resource for lang [did you mean RSC?]**
   - I thought about it - but I favoured the REST-API from a perspective of "repository as backend"
     - specialized frontend application (with own URL and different server) should use the repository as backend 
        (retrieve objects, execute Solr searches and ... get centrally stored messsages and their translations)
    - I see our RSC Endpoints as Endpoints for the repository frontend itself but not for "external clients".
      (different requirements for CORS, ...)

golsch:
Vue is currently our primary frontend, but I wouldn't want to commit to that being the case forever.
That's why I already started collecting shared functionality in [@mycore/js-common ](https://github.com/MyCoRe-Org/js-common) for cases like this. We could then bundle the framework-specific glue code in something like `vue-common`.

That's true, but at some point we need to stop maintaining v1 altogether. Otherwise we’ll never move forward. I don't think adding an endpoint for messages is all that complex.

Oops, I meant rsc. At the moment, that's our standard approach in MIR, for example. Then we should standardize on a single approach. But in general, I agree with you regarding the REST API.

toKrause (approved):
looks good, tested locally, works fine

PR #3023, opened 2026-06-30
MCR-3745 CLI Commands should use default XSL Transformer

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3745).

## Problem

The CLI command `xslt {0} with file {1}` (`MCRObjectCommands.xslt`) created its
`TransformerFactory` via `TransformerFactory.newInstance()`. With both Xalan and
Saxon on the classpath the JAXP service lookup picked Xalan instead of the
configured MyCoRe default factory.

## Fix

Instantiate the factory from `MCRXSLTransformer.DEFAULT_FACTORY_CLASS`
(resolved from `MCR.LayoutService.TransformerFactoryClass`), so the CLI uses the
same default XSL transformer as the rest of MyCoRe. To reuse the existing
single source of truth, `MCRXSLTransformer.DEFAULT_FACTORY_CLASS` was widened
from `protected` to `public`.

## Ticket & Documentation
- [ ] The `fixed-version` is correctly set in the ticket and matches the PR's target branch (`2026.06.x`).

## Bugfix-Specific Checks
- [ ] A relevant test was added (if feasible). Not feasible without both Xalan+Saxon on the classpath and an integration setup for the metadata manager.

## Testing
- [x] The feature behaves as described in the ticket. Verified in MIR: Saxon is now used.

## MCR Conventions & Metadata
  - [x] `MCRXSLTransformer.DEFAULT_FACTORY_CLASS` visibility widened from `protected` to `public` (non-breaking).

## Multi-Repo Considerations
- [ ] Is an equivalent PR in `MIR` required? No, MIR configuration already selects Saxon.

### DIFF
--- a/mycore-base/src/main/java/org/mycore/frontend/cli/MCRCommandUtils.java
+++ b/mycore-base/src/main/java/org/mycore/frontend/cli/MCRCommandUtils.java
@@ -37,8 +37,10 @@ import org.jdom2.filter.Filters;
 import org.jdom2.transform.JDOMSource;
 import org.jdom2.xpath.XPathExpression;
 import org.jdom2.xpath.XPathFactory;
+import org.mycore.common.MCRClassTools;
 import org.mycore.common.MCRConstants;
 import org.mycore.common.MCRUsageException;
+import org.mycore.common.content.transformer.MCRXSLTransformer;
 import org.mycore.common.xsl.MCRXSLResourceHelper;
 import org.mycore.common.xsl.uriresolver.MCRURIResolver;
 import org.mycore.datamodel.common.MCRXMLMetadataManager;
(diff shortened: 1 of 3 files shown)


mycore-base/src/main/java/org/mycore/frontend/cli/MCRCommandUtils.java
```diff
@@ -198,7 +200,8 @@ public static Transformer getTransformer(String style, String defaultStyle, Map<
 
         try {
             if (element != null) {
-                TransformerFactory transformerFactory = TransformerFactory.newInstance();
+                TransformerFactory transformerFactory = TransformerFactory
```
rsteph-de, line 203: Maybe we should create a static method for this: e.g. `createDefaultTransformerFactory` in MCRXSLTransformer

rsteph-de (approved):

PR #3024, opened 2026-06-30
MCR-3749 xed:output does not show HTML content

rsteph-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3749).

## Ticket & Documentation

## Bugfix-Specific Checks

## Testing

## MCR Conventions & Metadata

## Multi-Repo Considerations
- [-] Is an equivalent PR in `MIR` required?

### DIFF
--- a/mycore-xeditor/src/main/java/org/mycore/frontend/xeditor/transformer/MCROutputTransformerHelper.java
+++ b/mycore-xeditor/src/main/java/org/mycore/frontend/xeditor/transformer/MCROutputTransformerHelper.java
@@ -19,7 +19,6 @@
 package org.mycore.frontend.xeditor.transformer;
 
 import java.io.IOException;
-import java.io.StringReader;
 import java.util.Arrays;
 import java.util.List;
 
@@ -28,7 +27,8 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.jdom2.Element;
 import org.jdom2.JDOMException;
-import org.jdom2.input.SAXBuilder;
+import org.mycore.common.content.MCRStringContent;
+import org.mycore.common.xml.MCRXMLParserFactory;
 import org.mycore.services.i18n.MCRTranslation;
 
 /**
@@ -43,8 +43,6 @@ public class MCROutputTransformerHelper extends MCRTransformerHelperBase {
     private static final String ATTR_VALUE = "value";
     private static final String ATTR_I18N = "i18n";
     private static final String ATTR_ESCAPE_XML = "escape-xml";
-
-    private SAXBuilder saxBuilder = new SAXBuilder();
 
     @Override
     List<String> getSupportedMethods() {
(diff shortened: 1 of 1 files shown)


fluetze (approved):

mycore-xeditor/src/main/java/org/mycore/frontend/xeditor/transformer/MCROutputTransformerHelper.java
```diff
@@ -31,8 +38,13 @@
  */
 public class MCROutputTransformerHelper extends MCRTransformerHelperBase {
 
+    private static final Logger LOGGER = LogManager.getLogger();
+
     private static final String ATTR_VALUE = "value";
     private static final String ATTR_I18N = "i18n";
+    private static final String ATTR_ESCAPE_XML = "escape-xml";
+
+    private SAXBuilder saxBuilder = new SAXBuilder();
```
sebhofmann: A local variable would be saver since SAXBuilder is not threadsafe.  Instead of using a raw SAXBuilder, you could use the hardened MyCoRe Parser for consistency:
```
MCRXMLParserFactory.getNonValidatingParser().parseXML(new MCRStringContent("<text>" + output + "</text>"))
```

mycore-xeditor/src/main/java/org/mycore/frontend/xeditor/transformer/MCROutputTransformerHelper.java
```diff
@@ -43,8 +55,21 @@ List<String> getSupportedMethods() {
...
         String value = call.getAttributeValueOrDefault(ATTR_VALUE, null);
         String i18n = call.getAttributeValueOrDefault(ATTR_I18N, null);
+        boolean escapeXML = "true".equalsIgnoreCase(call.getAttributeValueOrDefault(ATTR_ESCAPE_XML, "true"));
         String output = output(value, i18n);
-        call.getReturnElement().setText(output);
+        if (escapeXML) {
+            call.getReturnElement().setText(output);
+        } else {
+            try {
+                Element eText = saxBuilder.build(new StringReader("<text>" + output + "</text>")).getRootElement();
```
sebhofmann: From security perspective the intended use case of the ticket is ok, but the `output` method does more. 

The output element has 3 parameters:

1. `i18n` - which is a translation key
2. `value` - which is a xpath which will be resolved from the document, and passed to the translation method or if no parameter `i18n` is set it will be used as the output
3. `escape-xml` - which is the new boolean parameter 

Assuming the new `escape-xml` set to false.

- If  `i18n`  is set and `value` is not set, the output comes solely from the message properties, so this case is fine.
- Every other case, the output can come from the document and then renders directly to raw html, which can pose an XSS threat

fluetze: The output may also come from configuration properties, and maybe also from request parameters, using the $variable syntax - which is also a valid, accepted xpath expression.

With this change, it is the responsibility of the xed form developer to be aware of the risks.
The escape-xml option is useful but dangerous, but only if combined with xpaths hat process user input.

rsteph-de: These are all valid concerns, I add some hints to the documentation after the PR- merge.

sebhofmann (approved):

PR #3039, opened 2026-07-06
MCR-3757 use xml:lang instead of lang() for classification label sele…

michael-becker:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3757).

## Ticket & Documentation

## Bugfix-Specific Checks

## Testing
  - [x] Yes: added a category to the classification for causing the error

## MCR Conventions & Metadata

## Multi-Repo Considerations

### DIFF
--- a/mycore-base/src/main/resources/xslt/functions/classification.xsl
+++ b/mycore-base/src/main/resources/xslt/functions/classification.xsl
@@ -47,14 +47,14 @@
       <xsl:when test="$class[@classid and @categid]">
         <xsl:sequence select="mcrclass:label($lang, document(concat('classification:metadata:0:children:',$class/@classid,':',$class/@categid))//category)" />
       </xsl:when>
-      <xsl:when test="string-length($lang) > 0 and $class/label[lang($lang)]">
-        <xsl:sequence select="$class/label[lang($lang)]" />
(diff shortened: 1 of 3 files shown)


toKrause (dismissed):

toKrause:
Should the new behavior only be used for language-tags starting with `x-'`? For normal languages, the old behavior might be better.

rsteph-de:
> Should the new behavior only be used for language-tags starting with `x-'`? For normal languages, the old behavior might be better.
for my use cases the `@xml:lang=$x` would work - here I am not dependent on the magic of the [`lang()`] (https://www.data2type.de/en/xml-xslt-xslfo/xslt/xslt-xpath-function-reference/alphabetical-xslt-and-xpath-reference/lang) function:
 - inheritance of the value of the `@xml:lang` Attribute to child elements (without `@xml:lang`)
 - strip country codes, e.g. `de-AT`&rarr; `de`

michael-becker:
> Should the new behavior only be used for language-tags starting with `x-'`? For normal languages, the old behavior might be better.

I would argue to change it for all languages. Since the function returns a single element, and lang() can return a list of elements, we can always run in an exception, e.g. if we have en_uk and en_us statements in a classification. Leaving it with lang, we would have to change the signature of the function and this might break applications elsewhere.

toKrause (approved):

yagee-de (approved):

PR #3040, opened 2026-07-07
MCR-3760 Externalize date formatting from PI generators

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3760).

## Ticket & Documentation

## Testing
- [X] Were existing tests modified?
  - [X] Yes: explain the changes for reviewers.

## MCR Conventions & Metadata
  - [x] Old API is deprecated or a migration is documented.
- [X] Javadoc is written for non-self-explanatory classes/methods (Clean Code).
- [X] All configuration options are documented in Javadoc and `mycore.properties`.
- [X] No default properties are hardcoded — all set via `mycore.properties`.

## Multi-Repo Considerations


mycore-base/src/main/java/org/mycore/common/date/MCRDateTimeStyler.java
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    public MCRDateTimeStyler(FormatStyle dateStyle, FormatStyle timeStyle, Locale locale) {
+        this(dateStyle, timeStyle, locale, ZoneId.systemDefault());
+    }
+
+    @SuppressForbidden
+    public MCRDateTimeStyler(FormatStyle dateStyle, FormatStyle timeStyle, Locale locale, ZoneId zoneId) {
+        Objects.requireNonNull(dateStyle, "Date style must not be null");
+        Objects.requireNonNull(dateStyle, "Time style must not be null");
```
sebhofmann: **Blocking, copy/paste bug.** This validates `dateStyle` a second time, so `timeStyle` is never null checked:

```suggestion
        Objects.requireNonNull(timeStyle, "Time style must not be null");
```

Without the fix, `new MCRDateTimeStyler(FormatStyle.SHORT, null, locale, zoneId)` passes both guards and blows up inside `DateTimeFormatter.ofLocalizedDateTime` with `NullPointerException: timeStyle`. Via configuration, a `TimeStyle` that fails to resolve yields an opaque JDK stack trace instead of the intended message.

mycore-pi/src/main/resources/components/pi/config/deprecated.properties
```diff
@@ -11,3 +11,5 @@ MCR.PI.Generator.(\S+)=MCR.PI.Generator.$1.Class
 MCR.PI.MetadataService.(\S+)=MCR.PI.MetadataService.$1.Class
 
 MCR.URIResolver.ModuleResolver.pi=MCR.URIResolver.ModuleResolver.pi.Class
+
+MCR.PI.Generator.TypeYearCountURN.DateFormat=(MCR.PI.Generator.TypeYearCountURN.DateFormatter.Class=org.mycore.common.date.MCRSimpleDateFormatter and MCR.PI.Generator.TypeYearCountURN.DateFormatter.Format)
```
sebhofmann: **Blocking, two problems.**

1. `DateFormat` is not deprecated yet. `MCRGenericPIGenerator.Factory` still declares `@MCRProperty(name = DATE_FORMAT_KEY, ...)`, `getDateFormatter()` still falls back to it (`MCRGenericPIGenerator:366`), the class javadoc still documents it, and `MCR.Default.PI.Generator.Generic.DateFormat=ddMMyyyy` still ships. `MCRConfigurationBase.checkForDeprecatedProperties` throws `MCRConfigurationException` at startup, so an installation that copied the shipped example and set `DateFormat` would no longer boot.

2. The generator ID is hard coded. Every other entry in this file uses the `(\S+)` capture, so as written only `TypeYearCountURN` is matched and every other generator ID, the normal case, gets no notice at all. If the entry is meant to stay, it should read:

```
MCR.PI.Generator.(\S+).DateFormat=(MCR.PI.Generator.$1.DateFormatter.Class=org.mycore.common.date.MCRSimpleDateFormatter and MCR.PI.Generator.$1.DateFormatter.Format)
```

Given point 1, I would rather drop the entry from this PR and add it in the release that actually removes `DateFormat`.

mycore-pi/src/main/java/org/mycore/pi/MCRGenericPIGenerator.java
```diff
@@ -118,11 +124,19 @@
  * Example:
  * <pre><code>
  * [...].Class=org.mycore.pi.MCRGenericPIGenerator
- * [...].GeneralPattern=urn:nbn:de:gbv:$CurrentDate-$1-$2-$ObjectType-$ObjectProject-$ObjectNumber-$Count-
- * [...].DateFormat=yyyy-MM-dd
+ * [...].GeneralPattern=10.1234/$ObjectType-$ObjectProject-$ObjectDate-$ObjectNumber
+ * [...].DateFormatt=yyyy-MM-dd
```
sebhofmann: **Blocking, typo:** `DateFormatt` with a double `t`.

```suggestion
 * [...].DateFormat=yyyy-MM-dd
```

`DateFormatt` is an unknown suffix, so nothing is reported. `DateFormat` then falls back to `MCR.Default.PI.Generator.Generic.DateFormat=ddMMyyyy` and every generated identifier carries `05112025` instead of `2025-11-05`. Silently wrong, and persistent identifiers are permanent.

mycore-base/src/test/java/org/mycore/common/date/MCRMockDateFormatter.java
```diff
@@ -0,0 +1,37 @@
+package org.mycore.common.date;
```
sebhofmann, line 19: **Blocking, missing GPL header.** This is the only new Java file in the PR without one, and it is public API of the `mycore-base` test jar that `mycore-pi` consumes, so it would ship unlicensed. `mvn clean install -Pauto-update` adds it.

mycore-base/src/main/java/org/mycore/common/date/MCRDateFormatterUtils.java
```diff
@@ -0,0 +1,49 @@
...
+
+    private MCRDateFormatterUtils() {
+    }
+
+    public static Locale getLocale(String locale) {
+        return switch (locale) {
+            case null -> Locale.getDefault();
+            case "ROOT" -> Locale.ROOT;
+            case "DEFAULT" -> Locale.getDefault();
+            case String languageTag -> Locale.forLanguageTag(languageTag);
```
sebhofmann: `Locale.forLanguageTag` accepts only BCP 47 tags. For the underscore form it returns `Locale.ROOT` without raising anything, and `de_DE` is exactly what the javadoc of every new formatter and styler gives as the example value.

So a user who copies `[...].Locale=de_DE` gets ROOT output ("January 1, 2000" instead of "1. Januar 2000") with nothing logged. Either change all examples to `de-DE`, or accept the `language_COUNTRY` form here. Worth noting that `MCRISO8601Date.getLocale` has the same trap and its default `MCR.Metadata.SimpleDateFormat.Locales` is literally `de_DE`, so the underscore form is what users of this codebase will reach for.

mycore-base/src/main/java/org/mycore/common/date/MCRDateStyler.java
```diff
@@ -0,0 +1,114 @@
...
+ * special values <code>DEFAULT</code> and <code>ROOT</code> can be used to set
+ * {@link Locale#getDefault()} and {@link Locale#ROOT} respectively).
+ * <li> The property suffix {@link MCRDateStyler#TIME_ZONE_KEY}can be used to
+ * specify the timezone (optional, defaults to {@link ZoneId#systemDefault()};
+ * special value <code>DEFAULT</code> can be used to set {@link ZoneId#systemDefault()}).
+ * </ul>
+ * Example:
+ * <pre><code>
+ * [...].Class=org.mycore.common.date.MCRDateStyler
+ * [...].Format=yyyy-MM-dd'T'HH:mm
```
sebhofmann: This example property does not exist. The class requires `DateStyle` (see `DATE_STYLE_KEY` below), and there is no `Format` property at all.

```suggestion
 * [...].DateStyle=MEDIUM
```

As documented, copying the block yields an `MCRConfigurationException` for the missing required `DateStyle`; a user who maps the value onto `DateStyle` instead gets `IllegalArgumentException` from `FormatStyle.valueOf("yyyy-MM-dd'T'HH:mm")`.

Separately, the class javadoc links `ofLocalizedDateTime(FormatStyle)` while the code calls `ofLocalizedDate`.

mycore-base/src/main/java/org/mycore/common/date/MCRDateTimeStyler.java
```diff
@@ -0,0 +1,123 @@
...
+ * special values <code>DEFAULT</code> and <code>ROOT</code> can be used to set
+ * {@link Locale#getDefault()} and {@link Locale#ROOT} respectively).
+ * <li> The property suffix {@link MCRDateTimeStyler#TIME_ZONE_KEY}can be used to
+ * specify the timezone (optional, defaults to {@link ZoneId#systemDefault()};
+ * special value <code>DEFAULT</code> can be used to set {@link ZoneId#systemDefault()}).
+ * </ul>
+ * Example:
+ * <pre><code>
+ * [...].Class=org.mycore.common.date.MCRDateTimeStyler
+ * [...].Format=yyyy-MM-dd'T'HH:mm
```
sebhofmann: Same as in `MCRDateStyler`: `Format` does not exist. This class needs both `DateStyle` and `TimeStyle`.

```suggestion
 * [...].DateStyle=MEDIUM
 * [...].TimeStyle=SHORT
```

mycore-pi/src/main/java/org/mycore/pi/doi/MCRCreateDateDOIGenerator.java
```diff
@@ -59,7 +54,7 @@
  * </code></pre>
  */
 @MCRConfigurationProxy(proxyClass = MCRCreateDateDOIGenerator.Factory.class)
-public class MCRCreateDateDOIGenerator extends MCRDOIGeneratorBase {
+public class MCRCreateDateDOIGenerator extends MCRGenericPIGenerator {
```
sebhofmann, line 57: Reparenting from `MCRDOIGeneratorBase` to `MCRGenericPIGenerator` changes three things silently:

1. **Parser resolution.** The old code held a hard-wired `new MCRDOIParser()`. `MCRGenericPIGenerator.generate` now calls `MCRPIManager.getInstance().getParserForType(type)`, which is `getParserInstance(typeParserMap.get(type))` and dereferences the result immediately. An application that has not loaded `MCR.PI.Parsers.doi` gets a bare NPE where the old code always worked. A guard with a message naming the missing property would help.

2. **Type widening.** The class is no longer an `MCRPIGenerator<MCRDigitalObjectIdentifier>` but effectively an `MCRPIGenerator<MCRPersistentIdentifier>`, so a misconfigured `MCR.PI.Parsers.doi` now surfaces as a `ClassCastException` in `MCRPIService.getNewIdentifier` rather than failing here. The class javadoc above still claims `MCRPIGenerator` for `MCRDigitalObjectIdentifier` identifiers.

3. **Instantiation cost.** A fresh parser is reflectively created on every `generate()` call instead of being held in a field.

Point 2 also makes commit `77772a7` source breaking, which should be marked with `!` and checked against MIR.

mycore-base/src/main/java/org/mycore/common/date/MCRSimpleDateFormatter.java
```diff
@@ -0,0 +1,116 @@
...
+
+    public MCRSimpleDateFormatter(String format, Locale locale) {
+        this(format, locale, ZoneId.systemDefault());
+    }
+
+    public MCRSimpleDateFormatter(String format, Locale locale, ZoneId zoneId) {
+        Objects.requireNonNull(format, "Format must not be null");
+        Objects.requireNonNull(locale, "Locale must not be null");
+        Objects.requireNonNull(zoneId, "Zone ID must not be null");
+        formatHolder = ThreadLocal.withInitial(getFormatSupplier(format, locale, TimeZone.getTimeZone(zoneId)));
```
sebhofmann: Two concerns with the `ThreadLocal` approach:

1. **Leak.** Configured formatters are cached configuration instances that live for the whole application lifetime, and nothing ever calls `remove()`. In a servlet container every request pool thread that formats a date permanently retains a `SimpleDateFormat` in its `ThreadLocalMap`, and on webapp redeploy the still referenced `ThreadLocal` pins the old classloader. This repo already ships `classloader-leak-prevention-core` to fight exactly that.

2. **Late validation.** The pattern is compiled on first use per thread, so `[...].Format=yyyy-MM-qq` passes startup validation and only throws `IllegalArgumentException` on the first `format()` call on each thread. The sibling `MCRDateTimeFormatter` fails fast in its constructor.

Since `MCRDateTimeFormatter` is thread safe and precompiled, is a separate `SimpleDateFormat` based implementation needed at all? If it is only there for legacy pattern compatibility, that is worth a javadoc note.

mycore-base/src/main/java/org/mycore/common/date/MCRISO8601DateFormatter.java
```diff
@@ -0,0 +1,109 @@
...
+        this.timeZone = Objects.requireNonNull(zoneId, "Zone ID must not be null").getId();
+    }
+
+    @Override
+    public String format(Date date) {
+
+        MCRISO8601Date isoDate = new MCRISO8601Date();
+        isoDate.setDate(date);
+
+        return isoDate.format(format, locale, timeZone);
```
sebhofmann: `MCRISO8601Date.format` catches every formatting exception and returns `null`, so this method can return `null` from an interface contract that has no nullability annotation.

Downstream, `MCRGenericPIGenerator.generate` does `resultingPI.replace(PLACE_HOLDER_OBJECT_DATE, dateFormatter.format(objectCreateDate))`, which then throws a context free NPE inside `String.replace` instead of an `MCRPersistentIdentifierException` naming the bad pattern. Please either propagate the failure or document the `null` return.

Also inherited from `MCRISO8601Date`: any pattern containing `G` goes through `df.format(dt).replace("-", "")`, so `new MCRISO8601DateFormatter("yyyy-MM-dd G")` returns `"20250101 n. Chr."`. That legacy quirk is now part of a general purpose formatting API and should at least be mentioned in the javadoc.

mycore-base/src/main/java/org/mycore/common/date/MCRISO8601DateFormatter.java
```diff
@@ -0,0 +1,109 @@
...
+public final class MCRISO8601DateFormatter extends MCRDateFormatterBase {
+
+    private final String format;
+
+    private final Locale locale;
+
+    private final String timeZone;
+
+    public MCRISO8601DateFormatter(String format) {
+        this(format, Locale.ROOT, ZoneId.systemDefault());
```
sebhofmann: This defaults to `Locale.ROOT`, but the class javadoc says the locale "defaults to `Locale#getDefault()`", the `Factory` honours that via `getLocale(null)`, and every sibling formatter's one argument constructor uses `Locale.getDefault()`.

Result: `new MCRISO8601DateFormatter("dd MMMM yyyy")` gives "01 January 2000" while `new MCRSimpleDateFormatter("dd MMMM yyyy")` on the same JVM gives "01 Januar 2000". `MCRISO8601DateFormatterTest` does not catch it because it only asserts numeric patterns.

mycore-base/src/main/java/org/mycore/common/date/MCRISO8601DateFormatter.java
```diff
@@ -0,0 +1,109 @@
...
+        MCRISO8601Date isoDate = new MCRISO8601Date();
+        isoDate.setDate(date);
+
+        return isoDate.format(format, locale, timeZone);
+
+    }
+
+    public static final class Factory implements Supplier<MCRISO8601DateFormatter> {
+
+        @MCRProperty(name = "Format")
```
sebhofmann: This is the only new `Factory` using raw string literals. `MCRSimpleDateFormatter`, `MCRDateTimeFormatter`, `MCRDateStyler` and `MCRDateTimeStyler` all expose `FORMAT_KEY` / `LOCALE_KEY` / `TIME_ZONE_KEY` constants.

Consequence: the class javadoc has to link `{@link MCRSimpleDateFormatter#FORMAT_KEY}` and friends, pointing readers at another class's API, and a rename here would not be caught by the compiler. Suggest declaring the constants on this class as well.

mycore-pi/src/main/java/org/mycore/pi/MCRGenericPIGenerator.java
```diff
@@ -341,10 +358,14 @@ public static class Factory implements Supplier<MCRGenericPIGenerator> {
 
         @Override
         public MCRGenericPIGenerator get() {
-            return new MCRGenericPIGenerator(MCRPIGeneratorUtils.SHARED_COUNTER, generalPattern, dateFormat,
+            return new MCRGenericPIGenerator(MCRPIGeneratorUtils.SHARED_COUNTER, generalPattern, getDateFormatter(),
                 projectIdMappings, typeIdMappings, Integer.parseInt(countPrecision), type, xPaths);
         }
 
+        private MCRDateFormatter getDateFormatter() {
+            return dateFormatter != null ? dateFormatter : new MCRSimpleDateFormatter(dateFormat, Locale.ROOT);
```
sebhofmann, line 366: This backward compatibility branch, the `DateFormat` to `MCRSimpleDateFormatter` fallback, now has no test coverage. `configurationDateFormatterPrecedence` only asserts that `DateFormatter` wins, and every other test injects `MCRMockDateFormatter`. Worth a test that sets only `DateFormat` and asserts the resulting identifier.

mycore-pi/src/test/java/org/mycore/pi/MCRGenericPIGeneratorTest.java
```diff
@@ -164,11 +232,41 @@ public void configuration() throws MCRPersistentIdentifierException {
...
+    @Test
+    @MCRTestConfiguration(properties = {
+        @MCRTestProperty(key = "Test.Class", classNameOf = MCRGenericPIGenerator.class),
+        @MCRTestProperty(key = "Test.GeneralPattern", string = "10.1234/$ObjectDate-$ObjectNumber"),
+        @MCRTestProperty(key = "Test.DateFormat", string = DATE_FORMAT),
+        @MCRTestProperty(key = "Test.DateFormatter.Class", classNameOf = MCRMockDateFormatter.class),
+        @MCRTestProperty(key = "Test.DateFormatter.Value", string = MOCK_DATE),
+        @MCRTestProperty(key = "Test.Type", string = MCRDigitalObjectIdentifier.TYPE)
+    })
+    public void configurationDateFormatterPrecedence() throws MCRPersistentIdentifierException {
```
sebhofmann, line 286: The rewrite removed the only test in which a real date formatter actually reached a date placeholder. The old `testGenerate` asserted against a live `SimpleDateFormat(DATE_FORMAT).format(new Date())`; now `testGenerate1`, `testGenerate2`, `configuration1` and `configuration2` all feed the constant `MCRMockDateFormatter`, and `generateMultiple` passes a real `MCRSimpleDateFormatter` but with a pattern (`10.1234/$ObjectType-$Count`) that contains no date placeholder, so it is never invoked.

Minor, in `testGenerate1` the `MCRMockCounter(42)` and `countPrecision=3` are dead setup because the pattern has no `$Count`.

mycore-base/src/test/java/org/mycore/common/date/MCRDateTimeFormatterTest.java
```diff
@@ -0,0 +1,98 @@
...
+
+    private static Date toDate(int year, int month, int day, int hour, int minute) {
+        return Date.from(toInstant(year, month, day, hour, minute));
+    }
+
+    @ParameterizedTest
+    @MethodSource("provideDates")
+    public void formatDate(Date date, String format, String expected) {
+
+        String formattedDate = new MCRDateTimeFormatter(format).format(date);
```
sebhofmann: The one argument constructor uses `Locale.getDefault()`, but the assertions expect ASCII digits. Run the build with a locale whose numbering system is non Latin (for example `-Duser.language=hi -Duser.country=IN -Duser.extensions=u-nu-deva`) and `DateTimeFormatter` emits Devanagari digits, so `assertEquals("20000101000000", ...)` fails. Passing an explicit `Locale.ROOT` pins the test. Same exposure in `MCRSimpleDateFormatterTest:64` and `:92`.

mycore-pi/src/main/java/org/mycore/pi/urn/MCRFLURNGenerator.java
```diff
@@ -68,17 +68,17 @@ public MCRFLURNGenerator(String namespace, String delimiter) {
     protected synchronized String buildNISS(MCRBase base, String additional) {
 
         Date date = new Date((System.currentTimeMillis() / 1000) * 1000);
-        String niss = MCRFLDateScrambler.scrambleDate(date);
+        String niss = new MCRFLDateScrambler().format(date);
```
sebhofmann: `MCRFLDateScrambler` is stateless, so a `private static final` instance would avoid allocating one per call.

More importantly, the collision handling right below has three issues: the 500 ms `Thread.sleep` happens inside the `synchronized` method and blocks every other URN generating thread; the `InterruptedException` is swallowed without restoring the flag via `Thread.currentThread().interrupt()`, so a shutdown signal is lost; and the retry is unbounded recursion rather than a loop, which under sustained contention can overflow the stack. A `while` loop plus interrupt restoration, ideally with the wait outside the monitor, would be safer.

Pre-existing behaviour, so fine to defer, but the file is touched by this PR anyway.

sebhofmann (changes requested):
[Link to jira](https://mycore.atlassian.net/browse/MCR-3760)

Reviewed locally at `ea46cbf`. The direction is good: pulling date formatting behind an `MCRDateFormatter` abstraction is a clear improvement over the hard-wired `SimpleDateFormat`, and the per-implementation `Factory` classes fit the existing configuration scheme. A number of issues should be resolved before merge, details inline.

## Blocking

1. **`MCRDateTimeStyler:89`** copy/paste bug: the second `requireNonNull` validates `dateStyle` again, so `timeStyle` is never checked.
2. **`deprecated.properties:15`** deprecates a property that is still fully supported, and hard codes one generator ID instead of the `(\S+)` capture every other entry uses. Installations that set `DateFormat` would fail to boot.
3. **`MCRGenericPIGenerator:128`** javadoc example reads `DateFormatt`. Copying it silently falls back to the default pattern and produces wrong, permanent identifiers.
4. **`MCRMockDateFormatter`** has no GPL header, and it ships as part of the `mycore-base` test jar that `mycore-pi` consumes.

## Should be addressed

5. **`MCRDateFormatterUtils:37`** `Locale.forLanguageTag` silently returns `Locale.ROOT` for the `de_DE` form that every new javadoc gives as the example.
6. **`MCRDateStyler:58` / `MCRDateTimeStyler:60`** configuration examples document a `Format` property that does not exist.
7. **`MCRCreateDateDOIGenerator:57`** the reparenting swaps a hard-wired parser for a config resolved one without a null guard, and widens the identifier type.
8. **`MCRSimpleDateFormatter:82`** `ThreadLocal<SimpleDateFormat>` in a long lived configured singleton, never removed, plus late pattern validation.
9. **`MCRISO8601DateFormatter:87`** can return `null` from `format`, and inherits the "strip all hyphens when the pattern contains `G`" quirk undocumented.
10. **`MCRFLURNGenerator:71`** sleeps 500 ms inside `synchronized`, swallows the interrupt, then recurses unbounded.

## Nits

`MCRISO8601DateFormatter:68` (locale default inconsistent with its own javadoc and all siblings), `MCRISO8601DateFormatter:93` (raw string literals instead of key constants), `MCRGenericPIGenerator:366` and `MCRGenericPIGeneratorTest:249` (the backward compatibility path lost its test coverage), `MCRDateTimeFormatterTest:64` and `MCRSimpleDateFormatterTest:64/92` (assertions depend on the machine default locale).

## Commit conventions

`77772a7 MCR-3760 reimplement MCRCreateDateDOIGenerator as subclass of MCRGenericPIGenerator` removes the public `MCRCreateDateDOIGenerator(MCRDOIParser, Counter, String, String, int)` constructor, changes the supertype and widens the generated identifier type. That is source breaking for MIR and overlays, but the commit carries no `!` marker, unlike its sibling `d434c66`. Please also check whether MIR needs a companion ticket and PR.

`6b26218 scout rule: fix typo` has no `MCR-xxxx` prefix.

mycore-pi/src/main/java/org/mycore/pi/urn/MCRFLURNGenerator.java
toKrause: I'll add the static constant. The other issue should be addressed in a specialized PR.

toKrause: I've created, https://mycore.atlassian.net/browse/MCR-3827, so this issue is tracked.

mycore-pi/src/test/java/org/mycore/pi/MCRGenericPIGeneratorTest.java
toKrause: I've added a test that uses the pattern from configuration values, addressing this point as well as the one above (testing that the `DateFormat` configuration still works).

mycore-pi/src/main/java/org/mycore/pi/doi/MCRCreateDateDOIGenerator.java
toKrause: Points 1 and 3 should probably be addressed in `MCRGenericPIGenerator`, not just for `MCRCreateDateDOIGenerator`. Since `MCRCreateDateDOIGenerator` fixates type `doi`, it can only create DOIs, assuming a sensible configuration where `MCR.PI.Parsers.doi` point to a parser for DOIs. In case of a misconfiguration, I assume, many other things will break as well.

toKrause: Due to type erasure, we can not check reliably, if `MCR.PI.Parsers.doi` points to a `MCRPIParser<MCRDigitalObjectIdentifier>`.

sebhofmann (commented):
Re-reviewed at `409a3299`. Thanks, the blocking points are all resolved: the `timeStyle` null check, the `deprecated.properties` entry, the `DateFormatt` typo and the missing GPL header on `MCRMockDateFormatter`. The `de_DE` handling in `MCRDateFormatterUtils` with its new test, the `ThreadLocal` removal in `MCRSimpleDateFormatter` plus fail fast validation, the non null contract in `MCRISO8601DateFormatter`, the key constants, the static `MCRFLDateScrambler`, the pinned test locales and the new `configurationDateFormat` test all look good to me.

Two small things that came in with `409a3299 MCR-3760 fix minor issues`, details inline.

mycore-base/src/test/java/org/mycore/common/config/examples/MCRExample01.java
```diff
@@ -0,0 +1,44 @@
+package org.mycore.common.config.examples;
```
sebhofmann: This file arrived with `409a3299 MCR-3760 fix minor issues` and looks unintentional:

1. It has no GPL header, and it is now the only Java file in the PR without one. `mvn clean install -Pauto-update` adds it.
2. It is unrelated to MCR-3760. It exercises `MCRConfiguration2.getInstanceOfOrThrow` with a local `Foo` class and touches nothing from the date formatting work.

Was it committed by accident? If it is meant to stay, it should probably go in on its own ticket.

mycore-base/src/main/java/org/mycore/common/date/MCRISO8601DateFormatter.java
```diff
@@ -0,0 +1,127 @@
...
+    @Override
+    public String format(Date date) {
+
+        MCRISO8601Date isoDate = new MCRISO8601Date();
+        isoDate.setDate(date);
+
+        String formattedDate = isoDate.format(format, locale, timeZone);
+
+        if (formattedDate == null) {
+            throw new MCRConfigurationException("Failed to convert date " + date + "with format " + format);
```
sebhofmann: Missing space before `with`, so the message reads `Failed to convert date Mon Sep 08 ...with format yyyy-MM-dd`.

```suggestion
            throw new MCRConfigurationException("Failed to convert date " + date + " with format " + format);
```

toKrause:
fixed minor issues and rebased to current main

sebhofmann (approved):

PR #3043, opened 2026-07-14
MCR-3762 errors in solr initialization are NOT 'successful'

rsteph-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3762).

## Ticket & Documentation

## Bugfix-Specific Checks

## Testing
- [ ] Were existing tests modified? - NO

## MCR Conventions & Metadata

## Multi-Repo Considerations
- [] Is an equivalent PR in `MIR` required? NO

### DIFF
--- a/mycore-solr/src/main/java/org/mycore/solr/schema/MCRSolrSchemaReloader.java
+++ b/mycore-solr/src/main/java/org/mycore/solr/schema/MCRSolrSchemaReloader.java
@@ -50,6 +50,7 @@ import com.google.gson.JsonArray;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 import com.google.gson.JsonParser;
+import com.google.gson.JsonSyntaxException;
 
 /**
  * This class provides methods to reload a SOLR schema using the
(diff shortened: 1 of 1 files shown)


kkrebs (approved):

sebhofmann:
## Local review

The functional change is limited to `MCRSolrSchemaReloader.executeSchemaRequest`. The approach is correct: Solr returns an HTTP body containing an `error` key even when the request is streamed, so parsing the body and logging via `LOGGER.error` on `error` fulfills the ticket. A few points below.

### 1. [Medium] NPE / IllegalStateException swallows the very error that should be logged

```java
String msg = json.getAsJsonObject().get("error").getAsJsonObject().get("msg").getAsString();
```

This assumes `error` is always an object with a non-null string `msg`. If Solr shapes the error differently (no `msg`, or `error` as an array/string), `getAsJsonObject()` throws `IllegalStateException` or `get("msg")` returns `null` and `getAsString()` throws NPE. That exception is then caught by the new `catch (Exception e)` and logged as `"Could not execute the following Solr config command"` with `command` (not `response`), so the actual Solr error message is lost. This defeats the purpose of the ticket for those cases.

Suggestion: null-guard `msg` and fall back to logging the full `response` when it is missing.

### 2. [Low-Medium] `catch (IOException)` widened to `catch (Exception)` is too broad

Previously runtime exceptions propagated; now everything is swallowed (e.g. `JsonSyntaxException` on a malformed response) and logged as `"Could not execute the following Solr config command"`. That message is misleading: the command *was* executed, only parsing the response failed. There is no `IllegalCatch` in checkstyle so this does not break the build, but the broad catch hides bugs. Consider catching `IOException | RuntimeException` explicitly, or adjusting the message.

### 3. [Please verify] The error branch only runs if solrj returns the body on HTTP 400

With `InputStreamResponseParser` solrj generally does not validate the status, so the body comes through and this works. But if the solrj version in use throws `RemoteSolrException` on a non-2xx response, the `msg` extraction never runs and you fall through to the generic error log. Worth a quick test against a real Solr with a broken schema command to confirm the extracted `msg` actually appears.

### 4. [Nit] Trailing whitespace in the new comment block

Some of the new comment lines (`msg` / `errors` / `code:400`) have trailing spaces, and the success-response sample in the comment has an unbalanced brace. The checkstyle `RegexpSingleline` rule for trailing spaces is set to `severity=ignore`, so this does not break the build; the Eclipse formatter should strip the trailing whitespace anyway.

### Looks good

- Imports are clean, `IOException` is still used elsewhere (no unused import).
- Catch order is valid (`SolrServerException` before `Exception`).
- The sample success/error responses in the comment are useful documentation.

**Main recommendation:** address finding 1 (null-guard plus log `response` instead of `command` in the fallback), otherwise on an error the change can swallow exactly the information it is meant to surface.

rsteph-de:
> 1) Suggestion: null-guard msg and fall back to logging the full response when it is missing.

Done! - If solr changes the returned JSON in the future, we could have handled the exception

> 2) widened catch IOException

Done! back to IOException and added JsonSyntaxException / left error message unchanged

> 3) [Please verify] The error branch only runs if solrj returns the body on HTTP 400

Ignored! - the AI answer makes assumptions I cannot confirm - 
  we always log the response and can solve the problem later if it occurs 

> 4) Trailing whitespace in the new comment block

Done! - manually, our Eclipse formatter has ignored those whitespace in comments

sebhofmann (approved):
An additional check for `get("error").isJsonObject()` would be good. But not a blocker.

rsteph-de:
> An additional check for `get("error").isJsonObject()` would be good. But not a blocker.

This problem is handled by catching the JsonSyntaxException. 
If Solr should change it return format in the future, we will have to look at this part of code anyways.

PR #3054, opened 2026-07-20
MCR-3773 Fix XEditor validation failures sharing the same XPath

yagee-de:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3773).

## Ticket & Documentation

## Bugfix-Specific Checks

## Testing

## MCR Conventions & Metadata

## Multi-Repo Considerations

### DIFF
--- a/mycore-xeditor/pom.xml
+++ b/mycore-xeditor/pom.xml
@@ -74,5 +74,10 @@
       <groupId>xalan</groupId>
       <artifactId>xalan</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-api</artifactId>
+      <scope>test</scope>
+    </dependency>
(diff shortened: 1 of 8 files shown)


fluetze (changes requested):
In general, the behavior of validation can be changed that way. It is a design choice to continue validation if a preceding rule already failed for the given xpath - as it was a design choice before to skip following validations, so that conditions can be used as guard clauses.

I am concerned that this change may break validation code with existing editor forms. What has to be documented? If a simple validation is done first (say required plus regex check) and a compilcated validation afterwards (say external method call), the second validation can not rely on the simple checks already done - it now always gets a value already identified as invalid, right? So the java method's implementation might need more code now to guard against bad values?

At least I'd like to see one or more unit tests for the combination of validation rules that may rely on each other, e.g. "required" validation plus regex plus condition on a number field...

PR #3063, opened 2026-07-21
MCR-3781 Treat blank properties as absent properties

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3781).

## Ticket & Documentation

## Testing

## MCR Conventions & Metadata

## Multi-Repo Considerations

### DIFF
--- a/mycore-base/src/main/java/org/mycore/common/MCRConstants.java
+++ b/mycore-base/src/main/java/org/mycore/common/MCRConstants.java
@@ -26,7 +26,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import org.apache.logging.log4j.LogManager;
 import org.jdom2.Namespace;
 import org.jdom2.xpath.XPathFactory;
-import org.mycore.common.config.MCRConfiguration2;
+import org.mycore.common.config.MCRConfigurationBase;
 
 /**
  * This class replaces the deprecated MCRDefaults interface and provides some
@@ -138,7 +138,7 @@ public final class MCRConstants {
                 }
             }
         }
-        MCRConfiguration2.getSubPropertiesMap("MCR.Namespace.")
+        MCRConfigurationBase.getSubpropertiesMap("MCR.Namespace.")
             .forEach(MCRConstants::registerNamespace);
     }
 
--- a/mycore-base/src/main/java/org/mycore/common/config/MCRConfiguration2.java
+++ b/mycore-base/src/main/java/org/mycore/common/config/MCRConfiguration2.java
@@ -18,13 +18,13 @@
 
 package org.mycore.common.config;
 
-import java.util.Collections;
 import java.util.Map;
 import java.util.Optional;
 import java.util.UUID;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.function.Supplier;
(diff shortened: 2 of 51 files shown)


toKrause:
**Most relevant changes:**

In `MCRInstanceConfiguration`:
- use `MCRConfiguration2#getAllPropertiesMap` (which does not contain empty properties) instead of `MCRConfigurationBase#getAllPropertiesMap` (which may contain empty properties)


In `MCRPropertyListSource` / `MCRPropertyMapSource`:

- removed `hasRelevantProperty` (because a collection with empty short-form-property now must be treated as collection that is not configured, aka. "missing")
- always return an instance; stop returning `null` as indicator to distinguish a "missing" collection from a collection that is intentionally configured as an empty collection

In `MCRSourceBase`:

- line 63/71/75: `isMissingResult` (in case of string property: empty or missing property; in case of collection: no entries) can now always be used, no need to treat special `null` case that was removed from `MCRPropertyListSource` / `MCRPropertyMapSource`
- renamed `missingException` -> `missingResultException`
- renamed `nullResultReplacement` -> `missingResultReplacement`

Test classes:

- treat test cases for empty and absent configurations the same (like absent configuration)

toKrause:
Rebased to current main and removed some unintended changes that are now part of upcoming PRs.

sebhofmann:
Reviewed locally on top of `origin/main` (branch is up to date, 0 commits behind). The overall direction looks good to me: routing `MCRInstanceConfiguration` through `MCRConfiguration2.getAllPropertiesMap()` makes the blank handling uniform instead of per source, removes the `hasRelevantProperty` bookkeeping, and guarantees that no `getResult` implementation can return `null` anymore. That last part also removes a latent NPE, since `MCRSourceBase:58` already called `isMissingResult(result)` and `MCRPropertyListSource.isMissingResult` is `result.isEmpty()`, which would have blown up on `null` with debug logging enabled.

A few things I would like to discuss before approving.

**1. An empty `Class` property no longer disables an implicit instance.** This is the most significant behavioral change and it is not mentioned in the commit message. `MCRInstanceConfiguration.resolveValueClass` reads `properties.get("Class")`. Previously `MCR.Foo.Class=` was present and blank, so the method returned `null` and the instance was not instantiatable. Now the key is filtered out entirely, so the code falls through to `options.contains(Option.IMPLICIT) && Modifier.isFinal(superClass.getModifiers())` and instantiates the super class. This affects every `MCRConfiguration2.getInstanceOf` call, since those always pass `Options.IMPLICIT`. In practice it is rare, because the super class is usually an interface or abstract, but for those cases the well established "empty value disables it" convention no longer works. `MCRInstantiatorImplicitTest` asserts the new behavior explicitly (the line `shouldNotInstantiateNestedClass |= classProperty == ClassProperty.SET_EMPTY;` was removed), so it is clearly intentional. Is disabling now supposed to happen only via `@MCRSentinel`, and can we get that into the breaking change notes?

**2. Values are now trimmed.** `MCRConfiguration2.getAllPropertiesMap()` applies `entry.getValue().trim()`. The instantiator therefore receives trimmed values where it previously received raw ones. That is consistent with `getString()`, but it is a second breaking change in the same commit without a mention.

**3. The default name fallback now triggers on an empty collection.** `MCRSourceBase:63` changed from `result == null` to `isMissingResult(result)`. For `MCRPropertyListSource` and `MCRPropertyMapSource` this means a list that is configured but whose entries were all rejected by a sentinel is now considered missing, so `defaultName` kicks in. Previously it did not, because the result was a non null empty list. This fits the "empty equals absent" doctrine, but it is a distinct behavioral change worth calling out. No effect on the instance sources, whose `defaultName()` is a constant `""`.

**4. Two contracts depending on the entry point.** The blank rule lives solely in the pre filtered property map, not in the sources themselves. `MCRPropertySource.isMissingResult` is still `result == null`. Anyone calling `MCRInstanceConfiguration.ofName(X.class, name, ownMap)` with unfiltered properties keeps the old behavior. Inside the repo only tests do that, since production code uses the shorthands, but external modules may not. If the rule is meant to be an invariant of the instantiator, it should probably also be enforced in `isMissingResult`.

**5. Nit:** the remaining `!value.isEmpty()` guards in `MCRPropertyListSource:114` and `MCRPropertyMapSource:119` are dead code on the standard path, and on the raw path they let `"   "` through. `isBlank()` would be more consistent with the new contract.

**6. Nit:** `@SuppressWarnings("PMD.NPathComplexity")` on `MCRSourceBase.get()` may be obsolete after the simplification. Worth a check.

toKrause:
**1.** Ticket description mentions this explicitly.

**2.** I've included this in the ticket description.

**3.** I've  clarified this in the ticket description.

**4.** We could fix this by letting `MCRPropertySource` check 'value == null || value.isBlank()', but in that case we would also have to trim values. `MCRPropertySource` should not do this extra work, because it is called quiet often. Instead, I removed the offending entry points (so `MCRInstanceConfiguration` always uses properly prepared properties) and adopted the tests.

**5.** I removed the unnecessary checks.

**6.** Suppression is still necessary.

sebhofmann (approved):

PR #3067, opened 2026-07-22
MCR-3785 Add method that provides all non-absent properties

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3785).

## Ticket & Documentation

## Testing

## MCR Conventions & Metadata

## Multi-Repo Considerations

### DIFF
--- a/mycore-base/src/main/java/org/mycore/common/config/MCRConfiguration2.java
+++ b/mycore-base/src/main/java/org/mycore/common/config/MCRConfiguration2.java
@@ -18,6 +18,7 @@
 
 package org.mycore.common.config;
 
+import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
 import java.util.UUID;
@@ -107,7 +108,7 @@ public class MCRConfiguration2 {
         .softValues()
         .build();
 
-    private static final AtomicReference<Map<String, String>> PROPERTIES = new AtomicReference<>(null);
+    private static final AtomicReference<Map<String, String>> PROPERTIES = new AtomicReference<>(Map.of());
 
     static Map<SingletonKey, Object> instanceHolder = new MCRConcurrentHashMap<>();
 
@@ -127,12 +128,14 @@ public class MCRConfiguration2 {
      */
     public static Map<String, String> getAllPropertiesMap() {
         return PROPERTIES.updateAndGet(existingValue -> {
-            if (existingValue != null) {
-                return existingValue;
+            if (existingValue.isEmpty()) {
+                return MCRConfigurationBase.getAllPropertiesMap()
+                    .entrySet()
+                    .stream()
+                    .filter(entry -> !entry.getValue().isBlank())
+                    .collect(Collectors.toUnmodifiableMap(Map.Entry::getKey, entry -> entry.getValue().trim()));
             }
-            return MCRConfigurationBase.getAllPropertiesMap().entrySet().stream()
-                .filter(entry -> !entry.getValue().isBlank())
-                .collect(Collectors.toUnmodifiableMap(Map.Entry::getKey, entry -> entry.getValue().trim()));
+            return existingValue;
         });
     }
 
@@ -525,7 +528,9 @@ public class MCRConfiguration2 {
     }
 
     static void clearCaches() {
-        PROPERTIES.set(null);
+        // BEWARE: do NOT replace with constant value, i.e. do NOT replace with PROPERTIES.set(Map.of());
+        // AtomicReference#updateAndGet only works as expected, when each update sets a distinct value
+        PROPERTIES.set(new HashMap<>());
         CONFIGURATIONS.invalidateAll();
     }
 
--- a/mycore-base/src/main/java/org/mycore/common/config/MCRConfiguration2.java
+++ b/mycore-base/src/main/java/org/mycore/common/config/MCRConfiguration2.java
@@ -18,13 +18,13 @@
 
 package org.mycore.common.config;
 
-import java.util.Collections;
 import java.util.Map;
 import java.util.Optional;
 import java.util.UUID;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.function.Supplier;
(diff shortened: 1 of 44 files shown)


yagee-de:
MCRConfigurationBase gives a view of the underlying properties files and provided a getString(String) method that returns a Optional that is empty if the property is not present and an empty String if the property is present but has a blank value.

On top of MCRConfigurationBase is MCRConfiguration2 which provides convenience methods for mapping values and get maps of values with the restriction that empty values of properties are handled as unset properties. This is the contract of the class:

> All {@link Optional} values returned by this class are {@link Optional#empty() empty} if the property
> * is not set OR the trimmed value {@link String#isEmpty() is empty}. If you want to distinguish between
> * empty properties and unset properties use {@link MCRConfigurationBase#getString(String)} instead.
 
 This PR breaks with the contract and also introduce new API.

toKrause:
I'm adding a single method to `MCRConfiguration2` that provides all non-blank properties, i.e all the properties that result in an nonempty result when calling, for example, `MCRConfiguration2#getString`. How does that break any contract.

I'd argue that the existing method, `MCRConfiguration2#getPropertiesMap` is breaking that contract.

toKrause:
The new method creates an unmodifiable map with all non-blank properties the first time it is called and places that map in a thread-safe cache. The cache is cleared, when properties are changed. 

Is there an alternative to adding that method that allows retrieving a map of all non-blank properties somewhat efficiently?

toKrause:
When this change was discussed in the web conference earlier this year, someone suggested using [`Maps#filterValues` from Guava](https://guava.dev/releases/23.0/api/docs/com/google/common/collect/Maps.html#filterValues-java.util.Map-com.google.common.base.Predicate-) with the map returned by `MCRConfiguration2#getPropertiesMap`, but the resulting map is not thread safe and some operations are not very efficient.

toKrause:
@yagee-de Can you elaborate on why you think that the suggested new method, `getAllPropertiesMap`, breaks the classes contract? I'm stumped.

mycore-base/src/main/java/org/mycore/common/config/MCRConfiguration2.java
```diff
@@ -107,10 +107,43 @@ public class MCRConfiguration2 {
...
+    public static Map<String, String> getAllPropertiesMap() {
+        return PROPERTIES.updateAndGet(existingValue -> {
+            if (existingValue != null) {
+                return existingValue;
+            }
+            return MCRConfigurationBase.getAllPropertiesMap().entrySet().stream()
+                .filter(entry -> !entry.getValue().isBlank())
+                .collect(Collectors.toMap(Map.Entry::getKey, entry -> entry.getValue().trim()));
+        });
+    }
```
yagee-de, line 140: The returned map should be unmodifiable:
```java
Map<String, String> first = MCRConfiguration2.getAllPropertiesMap();

assertSame(first, MCRConfiguration2.getAllPropertiesMap()); // passes

first.put("Injected.Property", "value");

assertNotEquals(
      "value",
      MCRConfiguration2.getAllPropertiesMap().get("Injected.Property")
  ); // fails

```

mycore-base/src/main/java/org/mycore/common/config/MCRConfigurationBase.java
yagee-de, line 1: https://github.com/MyCoRe-Org/mycore/commit/7d9a135d8a71339b1fc26238e7a5c45dedc883fb introduced a bug that affects this change and maybe fixed together with this PR.
A super class should not call a subclass. In https://github.com/MyCoRe-Org/mycore/blame/9246d4d1548086b68e8bc5c86d21103289936462/mycore-base/src/main/java/org/mycore/common/config/MCRConfigurationBase.java#L262 a call to clear caches is conditional where it should be unconditional. On every change the cache should be cleared if the cached result is affected.
I would prefer a base clearCache() method that could be overridden in subclasses to get rid of the circular dependency.

yagee-de (changes requested):
Overall this is how we planned to do this, but caching introduces a few more bugs than https://github.com/MyCoRe-Org/mycore/pull/2922 already introduced and should be fixed.

mycore-base/src/main/java/org/mycore/common/config/MCRConfiguration2.java
toKrause: small oversight, good point, will fix

mycore-base/src/main/java/org/mycore/common/config/MCRConfigurationBase.java
toKrause: This is not a bug, just bad design (`MCRConfigurationBase` should not directly invalidate caches of classes that use data provided by `MCRConfigurationBase`). It was introduced by [MCR-3335](https://github.com/MyCoRe-Org/mycore/pull/2746) (which was suggested, approved and merged on a Wednesday, so outside of our normal procedures). The code was than merely modified in [MCR-3668](https://github.com/MyCoRe-Org/mycore/pull/2922).

If you think that this should be cleaned-up, please make a follow-up PR. I don't think this issue should creep into this PR, which only aims to add a single method and is already way larger than ever intended.

yagee-de: As I stated this PR makes it worse as a new cache is in town that provides stale data. It should be fixed therefor.
I can even update via set(String, String) and will not see the value in getAllProperties() after that if I am correct.

yagee-de: We will address any caching issue in a later PR

yagee-de (dismissed):

mycore-base/src/main/java/org/mycore/datamodel/metadata/MCRObjectID.java
```diff
@@ -42,6 +42,7 @@
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonFormat;
 import com.fasterxml.jackson.annotation.JsonValue;
+import org.mycore.common.config.MCRConfigurationBase;
```
yagee-de: Import order should be fixed

yagee-de (changes requested):
Overall, everything looks good to me. Just found an import order problem that should be easy to fix.

mycore-base/src/main/java/org/mycore/datamodel/metadata/MCRObjectID.java
toKrause: fixed

mycore-base/src/main/java/org/mycore/common/config/MCRConfiguration2.java
```diff
@@ -107,10 +107,43 @@ public class MCRConfiguration2 {
...
+     * only properties that are recognized by, for example, {@link #getString(String)}.
+     */
+    public static Map<String, String> getAllPropertiesMap() {
+        return PROPERTIES.updateAndGet(existingValue -> {
+            if (existingValue != null) {
+                return existingValue;
+            }
+            return MCRConfigurationBase.getAllPropertiesMap().entrySet().stream()
+                .filter(entry -> !entry.getValue().isBlank())
+                .collect(Collectors.toUnmodifiableMap(Map.Entry::getKey, entry -> entry.getValue().trim()));
```
yagee-de: Thread 1: existingValue == null --> computing starts
Thread 2: set() is called and adds or changes a value and clears cache
Thread 1: finishes work and adds stalled data to cache

existingValue should check that MCRConfigurationBase.getResolvedProperties() is the same instance

```java
private record Snapshot(MCRProperties source, Map<String, String> properties) { }
private static final AtomicReference<Snapshot> PROPERTIES = new AtomicReference<>();
[…]
public static Map<String, String> getAllPropertiesMap() {
    //source changes if MCRConfigurationBase.set() is called
    MCRProperties source = MCRConfigurationBase.getResolvedProperties();
    Snapshot snapshot = PROPERTIES.get();
    if (snapshot == null || snapshot.source() != source) {
        snapshot = new Snapshot(source, computeFrom(source));
        PROPERTIES.set(snapshot); // last writer wins, but is checked at next call
    }
    return snapshot.properties();
}

```

yagee-de (changes requested):

mycore-base/src/main/java/org/mycore/common/config/MCRConfiguration2.java
toKrause: good point. fixed, but a bit more elegantly (w/o re-implementing part of what `AtomicReference` already does, i.e. the object reference comparison).

yagee-de (approved):

PR #3068, opened 2026-07-23
MCR-3786 Add annotations for configured Class-values

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3786).

## Ticket & Documentation

## Testing

## MCR Conventions & Metadata

## Multi-Repo Considerations

### DIFF
--- a/mycore-base/src/main/java/org/mycore/common/config/annotation/MCRClassProperty.java
+++ b/mycore-base/src/main/java/org/mycore/common/config/annotation/MCRClassProperty.java
@@ -1,0 +1,74 @@
+/*
+ * This file is part of ***  M y C o R e  ***
+ * See https://www.mycore.de/ for details.
+ *
+ * MyCoRe is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * MyCoRe is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with MyCoRe.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.mycore.common.config.annotation;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Inherited;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.mycore.common.config.MCRConfigurationException;
+
+/**
+ * This annotation is used to mark fields or methods that should be set to or called with a value of type
+ * {@link Class} from the configuration properties.
+ * <p>
+ * The field or method needs to be public.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ ElementType.METHOD, ElementType.FIELD })
+@Inherited
+public @interface MCRClassProperty {
+
+    /**
+     * @return The name of property containing the string value.
+     */
+    String name();
+
+    /**
+     * @return The class or a superclass of the configured class.
+     */
+    Class<?> valueClass() default Object.class;
+
+    /**
+     * @return true if the property specified by {@link MCRClassProperty#name()} has to be present in the properties.
+     * {@link MCRConfigurationException} is thrown if the property is required but not present.
+     */
+    boolean required() default true;
+
+    /**
+     * @return true if the property name specified by {@link MCRClassProperty#name()} is absolute and not specific for
+     * this instance e.g. <code>MCR.Foo.Bar</code>.
+     */
+    boolean absolute() default false;
+
+    /**
+     * @return The name for a default property that should be used as a fallback, if no value is configured.
+     * The default property must be absolute, e.g. <code>MCR.Foo.Bar</code>.
+     */
+    String defaultName() default "";
+
+    /**
+     * @return The order in which the annotated fields or methods are processed. The higher the value, the later the
+     * field or method is processed. All fields are processed first, then all methods are processed.
+     */
+    int order() default 0;
+
+}
--- a/mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRClassPropertyExtractor.java
+++ b/mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRClassPropertyExtractor.java
@@ -1,0 +1,45 @@
+/*
+ * This file is part of ***  M y C o R e  ***
+ * See https://www.mycore.de/ for details.
+ *
+ * MyCoRe is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * MyCoRe is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with MyCoRe.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.mycore.common.config.instantiator.source;
+
+import org.mycore.common.MCRClassTools;
+
+class MCRClassPropertyExtractor implements MCRValueExtractor<Class<?>> {
+
+    private final Class<?> valueClass;
+
+    MCRClassPropertyExtractor(Class<?> valueClass) {
+        this.valueClass = valueClass;
+    }
+
+    @Override
+    public Class<?> toValue(MCRSourceContext context, String value) {
+        try {
+            Class<?> configuredClass = MCRClassTools.forName(value);
+            if (!valueClass.isAssignableFrom(configuredClass)) {
+                throw context.incompatibilityException(valueClass, configuredClass);
+            }
+            return configuredClass;
+        } catch (ClassNotFoundException exception) {
+            throw context.configurationException("has a class (" + value
+                + ") that could not be loaded", exception);
+        }
+    }
+
+}
--- a/mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRClassPropertySource.java
+++ b/mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRClassPropertySource.java
@@ -18,37 +18,31 @@
 
 package org.mycore.common.config.instantiator.source;
 
-import java.util.Map;
 import java.util.Set;
 
-import org.mycore.common.config.MCRConfigurationException;
-import org.mycore.common.config.annotation.MCRInstance;
-import org.mycore.common.config.annotation.MCRSentinel;
-import org.mycore.common.config.instantiator.MCRInstanceConfiguration;
+import org.mycore.common.config.annotation.MCRClassProperty;
 import org.mycore.common.config.instantiator.target.MCRTarget;
 
 /**
- * A {@link MCRInstanceSource} is a {@link MCRSource} that interprets a {@link MCRInstance}.
+ * A {@link MCRClassPropertySource} is a {@link MCRSource} that interprets a {@link MCRClassProperty}.
  */
-final class MCRInstanceSource extends MCRInstanceSourceBase<Object> {
+final class MCRClassPropertySource extends MCRValueSourceBase<Class<?>> {
 
-    private final MCRInstance annotation;
+    private final MCRClassProperty annotation;
 
-    private final MCRSentinel sentinel;
-
-    MCRInstanceSource(MCRInstance annotation, MCRAnnotationProvider annotationProvider) {
+    MCRClassPropertySource(MCRClassProperty annotation, MCRAnnotationProvider annotationProvider) {
+        super(annotationProvider, new MCRClassPropertyExtractor(annotation.valueClass()));
         this.annotation = annotation;
-        this.sentinel = annotationProvider.get(MCRSentinel.class);
     }
 
     @Override
     public Type type() {
-        return Type.INSTANCE;
+        return Type.PROPERTY;
     }
 
     @Override
-    public Class<MCRInstance> annotationClass() {
-        return MCRInstance.class;
+    public Class<MCRClassProperty> annotationClass() {
+        return MCRClassProperty.class;
     }
 
     @Override
@@ -63,12 +57,12 @@ final class MCRInstanceSource extends MCRInstanceSourceBase<Object> {
 
     @Override
     public Class<?> valueClass() {
-        return annotation.valueClass();
+        return Class.class;
     }
 
     @Override
     protected String description() {
-        return "instance";
+        return "class";
     }
 
     @Override
@@ -83,7 +77,7 @@ final class MCRInstanceSource extends MCRInstanceSourceBase<Object> {
 
     @Override
     protected boolean absolute() {
-        return false;
+        return annotation.absolute();
     }
 
     @Override
@@ -93,31 +87,7 @@ final class MCRInstanceSource extends MCRInstanceSourceBase<Object> {
 
     @Override
     protected String defaultName() {
-        return "";
-    }
-
-    @Override
-    protected Object getResult(MCRSourceContext context, MCRInstanceConfiguration<?> configuration,
-        Map<String, String> properties, String prefix) {
-
-        MCRInstanceConfiguration<?> nestedConfiguration = configuration.nested(annotation.valueClass(), name());
-        return createInstance(context, nestedConfiguration, sentinel);
-
-    }
-
-    @Override
-    protected boolean isMissingResult(Object result) {
-        return result == null;
-    }
-
-    @Override
-    protected MCRConfigurationException missingResultException(MCRSourceContext context) {
-        return context.missingException();
-    }
-
-    @Override
-    protected Object missingResultReplacement() {
-        return null;
+        return annotation.defaultName();
     }
 
 }
--- a/mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRPropertyExtractor.java
+++ b/mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRPropertyExtractor.java
@@ -1,0 +1,28 @@
+/*
+ * This file is part of ***  M y C o R e  ***
+ * See https://www.mycore.de/ for details.
+ *
+ * MyCoRe is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * MyCoRe is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with MyCoRe.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.mycore.common.config.instantiator.source;
+
+class MCRPropertyExtractor implements MCRValueExtractor<String> {
+
+    @Override
+    public String toValue(MCRSourceContext context, String value) {
+        return value;
+    }
+
+}
--- a/mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRSource.java
+++ b/mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRSource.java
@@ -28,6 +28,9 @@ import java.util.Set;
 import java.util.function.BiFunction;
 import java.util.function.Function;
 
+import org.mycore.common.config.annotation.MCRClassProperty;
+import org.mycore.common.config.annotation.MCRClassPropertyList;
+import org.mycore.common.config.annotation.MCRClassPropertyMap;
 import org.mycore.common.config.annotation.MCRInstance;
 import org.mycore.common.config.annotation.MCRInstanceList;
 import org.mycore.common.config.annotation.MCRInstanceMap;
@@ -70,6 +73,18 @@ public sealed interface MCRSource permits MCRSourceBase, MCRPostConstructionSour
         PROPERTY_LIST(
             Group.VALUE_INJECTION,
             new Mapper<>(MCRPropertyList.class, MCRPropertyListSource::new)),
+
+        CLASS_PROPERTY(
+            Group.VALUE_INJECTION,
+            new Mapper<>(MCRClassProperty.class, MCRClassPropertySource::new)),
+
+        CLASS_PROPERTY_MAP(
+            Group.VALUE_INJECTION,
+            new Mapper<>(MCRClassPropertyMap.class, MCRClassPropertyMapSource::new)),
+
+        CLASS_PROPERTY_LIST(
+            Group.VALUE_INJECTION,
+            new Mapper<>(MCRClassPropertyList.class, MCRClassPropertyListSource::new)),
 
         RAW_PROPERTIES(
             Group.VALUE_INJECTION,
--- a/mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRSourceContext.java
+++ b/mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRSourceContext.java
@@ -97,9 +97,8 @@ public final class MCRSourceContext {
         return new MCRConfigurationException(exceptionMessage(exceptionMessage), exception);
     }
 
-    public MCRConfigurationException incompatibilityException(
-        Class<?> annotationValueClass, Object instance) {
-        return configurationException("has a class (" + instance.getClass().getName() + "),"
+    public MCRConfigurationException incompatibilityException(Class<?> annotationValueClass, Class<?> actualClass) {
+        return configurationException("has a class (" + actualClass.getName() + ")"
             + " that is incompatible with the annotated value class (" + annotationValueClass.getName() + ")");
     }
 
--- a/mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRValueListSourceBase.java
+++ b/mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRValueListSourceBase.java
@@ -1,0 +1,113 @@
+/*
+ * This file is part of ***  M y C o R e  ***
+ * See https://www.mycore.de/ for details.
+ *
+ * MyCoRe is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * MyCoRe is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with MyCoRe.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.mycore.common.config.instantiator.source;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import org.mycore.common.config.MCRConfiguration2;
+import org.mycore.common.config.MCRConfigurationException;
+import org.mycore.common.config.annotation.MCRSentinel;
+import org.mycore.common.config.instantiator.MCRInstanceConfiguration;
+
+abstract sealed class MCRValueListSourceBase<Value> extends MCRSourceBase<List<Value>> permits
+    MCRClassPropertyListSource, MCRPropertyListSource {
+
+    private final MCRSentinel sentinel;
+
+    private final MCRValueExtractor<Value> extractor;
+
+    MCRValueListSourceBase(MCRAnnotationProvider annotationProvider, MCRValueExtractor<Value> extractor) {
+        this.sentinel = annotationProvider.get(MCRSentinel.class);
+        this.extractor = extractor;
+    }
+
+    @Override
+    protected final List<Value> getResult(MCRSourceContext context, MCRInstanceConfiguration<?> configuration,
+        Map<String, String> properties, String prefix) {
+
+        Map<String, String> listProperties = new HashMap<>();
+        String keyPrefix = prefix.isEmpty() ? prefix : prefix + ".";
+        int keyPrefixLength = keyPrefix.length();
+        properties.forEach((key, value) -> {
+            if (key.startsWith(keyPrefix) && !key.isEmpty()) {
+                int index = key.indexOf('.', keyPrefixLength);
+                if (index == -1) {
+                    if (!value.isBlank()) {
+                        listProperties.put(key.substring(keyPrefixLength), value);
+                    }
+                }
+            }
+        });
+
+        List<Value> headPropertyList = new ArrayList<>(0);
+        List<Value> tailPropertyList = new ArrayList<>(listProperties.size());
+
+        List<String> keyList = context.orderedKeys(listProperties);
+        for (String key : keyList) {
+            MCRSourceContext nestedContext = context.nested(key, "property list element");
+            if (!rejectedBySentinel(sentinel, nestedContext, properties, keyPrefix + key + ".")) {
+                if (key.charAt(0) == '-') {
+                    headPropertyList.add(extractor.toValue(nestedContext, listProperties.get(key)));
+                } else {
+                    tailPropertyList.add(extractor.toValue(nestedContext, listProperties.get(key)));
+                }
+            }
+        }
+
+        List<Value> shortFormList = new ArrayList<>();
+        String shortFormProperty = properties.get(prefix);
+        if (shortFormProperty != null) {
+            for (String shortFormValue : parseShortFormList(shortFormProperty)) {
+                shortFormList.add(extractor.toValue(context, shortFormValue));
+            }
+        }
+
+        int totalSize = headPropertyList.size() + shortFormList.size() + tailPropertyList.size();
+        List<Value> list = new ArrayList<>(totalSize);
+        list.addAll(headPropertyList);
+        list.addAll(shortFormList);
+        list.addAll(tailPropertyList);
+
+        return list;
+
+    }
+
+    private List<String> parseShortFormList(String value) {
+        return MCRConfiguration2.splitValue(value).toList();
+    }
+
+    @Override
+    protected final boolean isMissingResult(List<Value> result) {
+        return result.isEmpty();
+    }
+
+    @Override
+    protected final MCRConfigurationException missingResultException(MCRSourceContext context) {
+        return context.emptyException();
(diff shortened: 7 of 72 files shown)


toKrause:
- Adds annotation `MCRClassProperty`
- Adds support for `MCRClassProperty` in `MCRSource`
- Extracts generic base class `MCRValueSourceBase` from `MCRPropertySource`
	- `MCRValueSourceBase` uses `MCRValueExtractor` to extract generic values
	- `MCRPropertySource` uses `MCRPropertyExtractor` to extract simple string values
- Adds `MCRClassPropertySource` to process `MCRClassProperty` annotations
	- `MCRClassPropertySource` uses `MCRClassPropertyExtractor` to extract class values
- Adds test class `MCRInstantiatorClassPropertyTest`

Same as above for `MCRClassPropertyList` and `MCRClassPropertyMap`

- Adds test class `MCRInstantiatorClassPropertyBasicTest` to demonstrate basic usages
- Updates `MCRSimpleJobSelector` and `MCRRestV2App` to use new annotations

mycore-restapi/src/main/java/org/mycore/restapi/v2/MCRRestV2App.java
```diff
@@ -146,40 +143,17 @@ public record Binding(Class<?> implementationClass, Class<?> apiClass) {
...
             @Override
             public Binding get() {
-                Class<?> implementationClass = toClass(this.implementationClass, "From");
-                Class<?> apiClass = toClass(this.apiClass, "To");
-                if (!apiClass.isAssignableFrom(implementationClass)) {
-                    throw new IllegalArgumentException("Implementation " + this.implementationClass + " configured in "
-                        + property + ".From does not implement/extend " + this.apiClass + " configured in "
-                        + property + ".To");
-                }
                 return new Binding(implementationClass, apiClass);
```
sebhofmann, line 165: **Regression: the `From`/`To` compatibility check is gone.**

The removed code verified `apiClass.isAssignableFrom(implementationClass)` and threw an `IllegalArgumentException` naming the offending property. Nothing re-establishes that check: both fields use `@MCRClassProperty` without `valueClass`, and the default `Object.class` makes the assignability test in `MCRClassPropertyExtractor` always pass.

With `MCR.RestAPI.V2.Services.ObjectLockService.From` and `.To` swapped (or pointing at unrelated classes), `registerServices()` now calls `bind(api).to(impl)` happily and the failure surfaces much later as an opaque HK2 `MultiException` on the first request that injects the service, with no reference to the misconfigured property.

The check cannot be expressed with the current annotations because `valueClass` is a compile time constant while `To` is configured at runtime, so this either needs an explicit check back in `get()` or a way for one class property to constrain another.

mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRClassPropertySource.java
```diff
@@ -0,0 +1,93 @@
...
+    private final MCRClassProperty annotation;
+
+    MCRClassPropertySource(MCRClassProperty annotation, MCRAnnotationProvider annotationProvider) {
+        super(annotationProvider, new MCRClassPropertyExtractor(annotation.valueClass()));
+        this.annotation = annotation;
+    }
+
+    @Override
+    public Type type() {
+        return Type.PROPERTY;
```
sebhofmann: **Wrong enum constant.** This returns `Type.PROPERTY`, but `Type.CLASS_PROPERTY` was added in this PR. The same applies to `MCRClassPropertyListSource.type()` (`Type.PROPERTY_LIST`) and `MCRClassPropertyMapSource.type()` (`Type.PROPERTY_MAP`).

It is silent today because `Type.order()` only returns `group.ordinal()` and both constants live in `Group.VALUE_INJECTION`. But `MCRSource.type()` is public API, so as soon as `order()` gains per constant meaning or any caller switches on `source.type()`, a `@MCRClassProperty` field becomes indistinguishable from a `@MCRProperty` field, without a compile error.

mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRClassPropertyExtractor.java
```diff
@@ -0,0 +1,45 @@
...
+
+    @Override
+    public Class<?> toValue(MCRSourceContext context, String value) {
+        try {
+            Class<?> configuredClass = MCRClassTools.forName(value);
+            if (!valueClass.isAssignableFrom(configuredClass)) {
+                throw context.incompatibilityException(valueClass, configuredClass);
+            }
+            return configuredClass;
+        } catch (ClassNotFoundException exception) {
```
sebhofmann: **`LinkageError` escapes unwrapped.** `MCRClassTools.forName` delegates to `Class.forName(String)`, which also *initializes* the class. Only `ClassNotFoundException` is caught here, so `ExceptionInInitializerError`, `NoClassDefFoundError` and other `LinkageError`s propagate out of `MCRInstantiator` raw.

For `Foo.Value = com.example.BadInit` with a throwing static initializer, the operator gets no hint which property caused it. The old inline code in `MCRRestV2App` and `MCRSimpleJobSelector` had the same gap, but this new central extractor is the right place to close it.

Also note that `context.incompatibilityException(...)` is thrown inside the `try`, so it currently passes through the `catch` unharmed only because it is not a `ClassNotFoundException`. Worth keeping the loading and the check in separate scopes for clarity.

sebhofmann (commented):
## Local code review

Reviewed the full diff of this branch against `main`. The extraction of `MCRValueExtractor` and the shared `MCRValue*SourceBase` classes is a good simplification, and moving class loading out of the individual factories is clearly the right direction. Twelve findings below, one of which I consider a blocker.

### Blocker

**`MCRRestV2App.Binding.Factory` silently lost its type check.** See the inline comment. The `apiClass.isAssignableFrom(implementationClass)` validation was deleted and not replaced, and because `@MCRClassProperty.valueClass()` defaults to `Object.class`, nothing validates the configured classes any more. A misconfiguration that used to fail at startup with the property name in the message now fails on the first request with an HK2 `MultiException`.

The root cause is shared with the second point below, so fixing `valueClass()` first is probably the better order.

### Summary of the inline comments

| # | Location | Topic |
|---|---|---|
| 1 | `MCRRestV2App:154` | Lost `From`/`To` compatibility check (blocker) |
| 2 | `MCRClassPropertySource:40` | `type()` returns `PROPERTY` instead of `CLASS_PROPERTY` (also list and map) |
| 3 | `MCRClassPropertyExtractor:39` | `LinkageError` from class initialization escapes unwrapped |
| 4 | `MCRClassProperty:48` | `valueClass()` defaults to `Object.class`, inconsistent with `MCRInstance*` |
| 5 | `MCRSimpleJobSelector:156` | `required = true` versus blank shipped default (pre existing) |
| 6 | `MCRValueListSourceBase:64` | Hardcoded element description in the shared base |
| 7 | `MCRValueListSourceBase:78` | Short form elements extracted with the parent context |
| 8 | `MCRValueMapSourceBase:67` | Short form entries get a context for a non existent property |
| 9 | `MCRValueSourceBase:30` | Sentinel field duplicated six times |
| 10 | `MCRSource:77` | Nine new classes per value type (design question) |
| 11 | `MCRInstantiatorClassPropertyBasicTest:44` | Untested failure branches for list and map |

### Two findings outside the diff

**`MCRInstanceConfiguration.ofClassName` (line 91) was not included in the `MCRInstantiatorUtils.getClass` refactoring.** `resolveValueClass` in the same file was converted, `ofClassName` was not, and `MCRClassPropertyExtractor.toValue` adds a third independent copy of the load and wrap logic. The three paths throw three different things:

* `resolveValueClass`: `MCRConfigurationException("Missing class (X) configured in: Y")`
* `ofClassName`: `MCRException("Failed to load class X")`, with no property name at all
* `MCRClassPropertyExtractor`: `MCRConfigurationException("... has a class (X) that could not be loaded")`

Callers that catch `MCRConfigurationException` to report configuration problems miss the `ofClassName` path entirely, and the `LinkageError` gap from finding 3 would have to be fixed in three places.

**`MCRInstantiator` javadoc (lines 150 to 158) still lists the old annotation set.** The list claims *"Source implementations exist for all supported annotations"* but names only `MCRProperty`, `MCRRawProperties`, `MCRInstance`, `MCRInstanceMap`, `MCRInstanceList` and `MCRPostConstruction`. `MCRPropertyList` and `MCRPropertyMap` were already missing, and this PR adds three more, so the one place documenting the complete annotation set now omits five of eleven.

<sub>Review produced locally with Claude Code and verified against the working tree.</sub>

mycore-base/src/main/java/org/mycore/common/config/annotation/MCRClassProperty.java
```diff
@@ -0,0 +1,74 @@
...
+
+    /**
+     * @return The name of property containing the string value.
+     */
+    String name();
+
+    /**
+     * @return The class or a superclass of the configured class.
+     */
+    Class<?> valueClass() default Object.class;
```
sebhofmann: **`valueClass()` should not have a default.** `MCRInstance`, `MCRInstanceList` and `MCRInstanceMap` all declare `Class<?> valueClass();` without a default, which forces the developer to state the contract. Here (and in `MCRClassPropertyList` and `MCRClassPropertyMap`) the default is `Object.class`, and `Object.class.isAssignableFrom(anything)` is always true, so the type check silently degrades to no check at all.

This is not hypothetical: the only production consumer added in this PR, `MCRRestV2App.Binding.Factory`, uses `@MCRClassProperty(name = "From")` without `valueClass` and therefore validates nothing. Making the attribute mandatory, consistent with the `MCRInstance*` family, would have made that regression impossible.

mycore-jobqueue/src/main/java/org/mycore/services/queuedjob/MCRSimpleJobSelector.java
```diff
@@ -155,8 +153,9 @@ public enum Mode {
 
     public static class Factory implements Supplier<MCRSimpleJobSelector> {
 
-        @MCRPropertyList(name = ACTIONS_KEY, defaultName = "MCR.QueuedJob.Selectors.Default.Actions")
-        public List<String> actions;
+        @MCRClassPropertyList(name = ACTIONS_KEY, valueClass = MCRJobAction.class,
```
sebhofmann: **Pre existing, but worth fixing while the line is being rewritten:** `required` keeps its `true` default, while `components/jobqueue/config/mycore.properties:21` ships `MCR.QueuedJob.Selectors.Default.Actions=` (blank).

`MCRConfiguration2.getAllPropertiesMap()` filters blank values, so the property is absent, `MCRValueListSourceBase` returns an empty list, `isMissingResult` is true and `MCRSourceBase.get()` throws *"Default class list, configured in MCR.QueuedJob.Selectors.Default.Actions ... is empty"*. Yet `getPredicates` explicitly handles `actions.isEmpty()` and the constructor accepts an empty collection, so a selector configured with only `.Statuses` and `.AgeDays` is intended to be legal but is rejected. `statuses` on line 163 has the same problem via `Default.Statuses`.

mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRValueListSourceBase.java
```diff
@@ -0,0 +1,111 @@
...
+                }
+            }
+        });
+
+        List<Value> headPropertyList = new ArrayList<>(0);
+        List<Value> tailPropertyList = new ArrayList<>(listProperties.size());
+
+        List<String> keyList = context.orderedKeys(listProperties);
+        for (String key : keyList) {
+            MCRSourceContext nestedContext = context.nested(key, "property list element");
```
sebhofmann: **Hardcoded description in a now shared base class.** `"property list element"` here (and `"property map entry"` in `MCRValueMapSourceBase:67`) was correct while these bodies lived in `MCRPropertyListSource`, but the base is now shared with `MCRClassPropertyListSource`.

For a `@MCRClassPropertyList` field the outer failure reads *"Class list, configured in Foo.Values ..."* while the element failure for the very same field reads *"Property list element, configured in Foo.Values.10 ..."*. Deriving it from the concrete source, e.g. `description() + " element"`, would keep the two consistent.

mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRValueListSourceBase.java
```diff
@@ -0,0 +1,111 @@
...
+                    tailPropertyList.add(extractor.toValue(nestedContext, listProperties.get(key)));
+                }
+            }
+        }
+
+        List<Value> shortFormList = new ArrayList<>();
+        String shortFormProperty = properties.get(prefix);
+        if (shortFormProperty != null) {
+            for (String shortFormValue : parseShortFormList(shortFormProperty)) {
+                shortFormList.add(extractor.toValue(context, shortFormValue));
```
sebhofmann, line 75: **Short form elements lose their position.** The long form branch above builds a nested context per element, this one passes the parent `context`.

For `Foo.Values = a.b.Good, a.b.Missing, a.b.AlsoGood` the extractor throws *"Class list, configured in Foo.Values (and its sub-properties) ... has a class (a.b.Missing) that could not be loaded"*. The class name only makes it into the message by luck; the index does not, so for a long list the operator has to guess. This did not matter while `MCRPropertyListSource` could not fail during extraction, but it can now.

mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRValueMapSourceBase.java
```diff
@@ -0,0 +1,100 @@
...
+                if (index == -1) {
+                    mapProperties.put(key.substring(keyPrefixLength), value);
+                }
+            }
+        });
+
+        Map<String, Value> map = new HashMap<>();
+
+        for (String key : mapProperties.keySet()) {
+            MCRSourceContext nestedContext = context.nested(key, "property map entry");
```
sebhofmann: **Short form entries get a nested context for a property that does not exist.** `shortFormMap` is merged into `mapProperties` before the loop, so a short form entry ends up with the context `prefix + "." + key` and a sentinel lookup prefix of `keyPrefix + key + "."`.

For `Foo.Values = alpha:com.example.Missing` the failure reads *"Property map entry, configured in Foo.Values.alpha (and its sub-properties) ... has a class (com.example.Missing) that could not be loaded"*, but there is no `Foo.Values.alpha` in the configuration. The operator is pointed at a key that does not exist instead of at `Foo.Values`. The sentinel lookup had the same defect before, it was just invisible while string maps could not fail.

mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRValueSourceBase.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.Map;
+
+import org.mycore.common.config.MCRConfigurationException;
+import org.mycore.common.config.annotation.MCRSentinel;
+import org.mycore.common.config.instantiator.MCRInstanceConfiguration;
+
+abstract sealed class MCRValueSourceBase<Value> extends MCRSourceBase<Value> permits MCRClassPropertySource,
+    MCRPropertySource {
+
+    private final MCRSentinel sentinel;
```
sebhofmann: **The sentinel field is now duplicated six times.** `private final MCRSentinel sentinel` plus `annotationProvider.get(MCRSentinel.class)` appears here, in `MCRValueListSourceBase:34`, in `MCRValueMapSourceBase:33` and in the three `MCRInstance*Source` classes.

Its only consumer is `MCRSourceBase.rejectedBySentinel`, i.e. the state belongs to the class that uses it. Moving it into `MCRSourceBase` (constructor taking the `MCRAnnotationProvider`, `rejectedBySentinel(context, properties, prefix)` reading the field) would remove all six copies and the three argument signature that exists only to hand the field back to its owner.

mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRSource.java
```diff
@@ -71,6 +74,18 @@ enum Type {
             Group.VALUE_INJECTION,
             new Mapper<>(MCRPropertyList.class, MCRPropertyListSource::new)),
 
+        CLASS_PROPERTY(
```
sebhofmann, line 81: **Design question, not a blocker:** supporting one additional value type cost three annotations, three sources and three abstract base classes. `MCRClassProperty`, `MCRClassPropertyList` and `MCRClassPropertyMap` are copies of `MCRProperty`, `MCRPropertyList` and `MCRPropertyMap` plus a single `valueClass()` member, and the three `MCRClassProperty*Source` classes differ from the `MCRProperty*Source` classes only in the extractor handed to `super(...)`.

The `MCRValueExtractor` abstraction introduced here already isolates the only varying behaviour, which is the good part of this change. It is just not surfaced generically, so the next value type (`URI`, `Path`, `Duration`, enums) repeats the same nine class expansion. Worth considering whether `@MCRProperty` could select the extractor from the target's declared type, or carry an explicit `converter` attribute, instead of growing a parallel annotation family per type.

mycore-base/src/test/java/org/mycore/common/config/MCRInstantiatorClassPropertyBasicTest.java
```diff
@@ -0,0 +1,145 @@
...
+import org.mycore.common.config.instantiator.MCRInstanceConfiguration;
+import org.mycore.test.MyCoReTest;
+
+@MyCoReTest
+public class MCRInstantiatorClassPropertyBasicTest {
+
+    @Test
+    @MCRTestConfiguration(
+        properties = {
+            @MCRTestProperty(key = "Foo.Class", classNameOf = TestClass.class),
```
sebhofmann, line 46: **Coverage gaps.** `MCRClassPropertyExtractor.toValue` has two failure paths and the tests only reach one of them, for the scalar case:

1. No test sets a property to a non existent class name, so the `catch (ClassNotFoundException)` branch and its message are unverified.
2. No `@MCRClassPropertyList` or `@MCRClassPropertyMap` in the new tests declares `valueClass` (all 27 annotation uses across the three test classes rely on the `Object.class` default), so the assignability check can never fail there and the nested context messages ship unchecked.

Adding an incompatible element to a list and to a map would also pin down the message wording discussed in the `MCRValueListSourceBase` and `MCRValueMapSourceBase` comments.

mycore-base/src/main/java/org/mycore/common/config/annotation/MCRClassProperty.java
toKrause: Not the same situation. With `MCRInstance`, it makes no sense to inject a plain `Object`, since we want to interact with the injected values. Therefore, the expected class has to be configured.

With `MCRClassProperty`, injecting a generic `Class<?>` is a sensible default behavior. Configuring a `valueClass` is just an opt-in restriction on the allowed values; `Object.class` is a sensible (no-op) default value for this feature.

Compare this to `MCRProperty` with the upcoming conversion-feature (for `Integer`, `Enum`, ...) where configuring a `valueClass` enables an opt-in conversion and `String.class` is a sensible (no-op) default value for that feature.

mycore-jobqueue/src/main/java/org/mycore/services/queuedjob/MCRSimpleJobSelector.java
toKrause: Will fix in a separate PR.

mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRValueListSourceBase.java
toKrause: I don't understand what the problem is supposed to be here. If the short form property `Foo.Values` contains a problematic value, the error message correctly points to that property and names the problematic value.

mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRSource.java
toKrause: We only need six new classes per additional "type" (three annotations and one source per annotation, the latter can't be improved, because annotation classes can't implement an interface). However we do not need additional "types" per feature, only per structurally distinct "type". Conversion of simple properties (string to integer, string to enum value, ...) can and will be handled by the existing sources for `MCRProperty`, ...

mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRValueMapSourceBase.java
toKrause: Already fixed in upcoming PR  https://github.com/MyCoRe-Org/mycore/pull/3074

mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRValueListSourceBase.java
sebhofmann: Expanding on this, since the original wording was too terse.

### The two notations

A list property can be written in two ways:

```properties
# long form
Foo.Values.10 = a.b.Good
Foo.Values.20 = a.b.Missing
Foo.Values.30 = a.b.AlsoGood

# short form
Foo.Values = a.b.Good, a.b.Missing, a.b.AlsoGood
```

`getResult` handles them in two separate loops, and the loops treat the context differently.

**Long form** (line 60 after 9fc6332060):
```java
MCRSourceContext nestedContext = context.nested(key, context.description() + " element");
...
tailPropertyList.add(extractor.toValue(nestedContext, listProperties.get(key)));
```

**Short form** (line 74):
```java
shortFormList.add(extractor.toValue(context, shortFormValue));
```

The first gives every element its own context with `property = Foo.Values.20`. The second passes the parent context, whose `property` is just `Foo.Values`. Since the context is what builds the message in `MCRSourceContext.exceptionMessage`, the two branches produce different diagnostics for the same defect.

### The resulting messages

Long form:

> **Class list element**, configured in **Foo.Values.20** (and its sub-properties), for target field 'values' in configured class org.example.Foo has a class (a.b.Missing) that could not be loaded

Short form:

> **Class list**, configured in **Foo.Values** (and its sub-properties), for target field 'values' in configured class org.example.Foo has a class (a.b.Missing) that could not be loaded

The first points at the offending line, the second at the whole list. With three entries that is fine, with fifteen it is a search.

### What "by luck" was meant to say

In the short form case the context contributes no positional information at all. The operator still sees `a.b.Missing` only because `MCRClassPropertyExtractor` happens to inline the value into its own message text:

```java
throw context.configurationException("has a class (" + value + ") that could not be loaded", exception);
```

That is a property of this one extractor, not a contract of `MCRValueExtractor`. Both of today's failure paths (`could not be loaded` and `incompatibilityException`) name the value, which is why the case is still workable right now. A future extractor for `Duration`, `URI` or enums that simply reports `"is not a valid duration"` would, on a short form list, produce a message from which neither the position nor the value can be recovered. Hence a latent trap rather than a bug.

### Why it did not matter before

`MCRPropertyExtractor.toValue` is a plain `return value;`. Extraction could not fail for string lists, so the context handed to the short form loop had no observable effect. With `MCRClassPropertyExtractor` extraction can throw, which is what makes the difference between the two branches visible for the first time.

### Why `nested(index)` is not the fix

The obvious repair would be `context.nested(String.valueOf(i), ...)`, but no property `Foo.Values.0` exists in the configuration, so the message would point the operator at a key they cannot find in their `mycore.properties`. That is exactly the defect described in the `MCRValueMapSourceBase:67` comment, where `shortFormMap` is merged into `mapProperties` before the loop and short form entries therefore get a nested context for a key that does not exist.

The short form needs information that leaves `property` untouched. `MCRSourceContext` already carries the machinery for that: the `hints` field, which `description()` renders as `" (" + join(hints) + ")"`. It is currently never populated and `hints()` is never called. A `hinted("element " + i)` alongside `nested(...)` would yield:

> **Class list (element 2)**, configured in **Foo.Values** (and its sub-properties), ... has a class (a.b.Missing) that could not be loaded

Correct property, position included, and the same method would cover the map case.

### Severity

This is the weakest of the message related findings and not a blocker. Leaving it as is would be defensible. The part I would not leave unmarked is the inconsistency between two branches of the same method, so that the next extractor does not walk into it. The sibling finding is already addressed in 9fc6332060: `context.description() + " element"` on line 61 is precisely the derivation I meant.

mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRSource.java
sebhofmann: Fair points, and you are right on the parts that carry the most weight. Checked them against the code.

### Conceded

**My count was wrong.** The three abstract bases are one time infrastructure introduced by this PR, not marginal cost per type. Six is the fair number.

**The interface argument holds.** Annotation types have no common supertype, so some per annotation translation is unavoidable.

**The limiting principle is the important part.** If string to integer, string to enum and so on are handled by the conversion feature on the existing `MCRProperty` sources, the family grows per structurally distinct type, not per feature. That removes the concern I actually had, namely that `URI` or `Duration` would repeat the expansion. They will not.

### Where I would still push back

**"the latter can't be improved" reads too strong.** That annotations cannot implement an interface only means a projection from annotation to specification is required, and that is a lambda rather than a class. `MCRClassPropertySource` and `MCRPropertySource` are 93 lines each and, after renaming, differ in exactly four:

```
29  extends MCRValueSourceBase<Class<?>>                     vs  <String>
34  new MCRClassPropertyExtractor(annotation.valueClass())   vs  new MCRPropertyExtractor()
40  Type.CLASS_PROPERTY                                      vs  Type.PROPERTY
60  Class.class                                              vs  String.class
```

The other 89 lines (`name()`, `required()`, `absolute()`, `defaultName()`, `order()`, `allowedTargetTypes()`, `description()`, `allowsEmptyName()`) are identical delegation. The wiring point already exists: `Mapper` takes a `BiFunction<A, MCRAnnotationProvider, MCRSource>` in `MCRSource:134`, so it accepts an arbitrary factory. A generic `MCRValueSource<V>` parameterized with spec, type, extractor and description would need roughly five lines of projection per annotation instead of a 93 line class.

**The point the reply does not cover: hand written `type()` is a defect source.** That was the finding two thread. All three new sources returned the wrong enum constant, and it took a manual fix in 9fc6332060 because `Type.PROPERTY` and `Type.CLASS_PROPERTY` are the same Java type and the compiler cannot help.

`Type` already knows its annotation class through `Mapper.annotationClass()`, so a static reverse index would make `type()` implementable once, final, in `MCRSourceBase`:

```java
@Override
public final Type type() {
    return Type.fromAnnotation(annotationClass());
}
```

That does not fix the bug, it makes it unconstructible. Given that the bug actually occurred in this PR rather than being hypothetical, this is the part of the design question I find hardest to drop.

**`valueClass` is about to carry two meanings.** Per your reply on the `MCRClassProperty` thread, `@MCRProperty.valueClass` will be the conversion target while `@MCRClassProperty.valueClass` is the upper bound of the permitted classes. Across `MCRSource.valueClass()` the name consistently means "type of the injected value":

| Source | `valueClass()` |
|---|---|
| `MCRPropertySource` | `String.class` |
| `MCRPropertyListSource` | `List.class` |
| `MCRInstanceMapSource` | `Map.class` |
| `MCRInstanceSource` | `annotation.valueClass()` |
| `MCRClassPropertySource` | `Class.class` |

For `MCRInstance` the injected type and the bound coincide, which is why the same name works there. For `MCRClassProperty` they do not, and the two meanings collide inside a single class: line 34 reads `annotation.valueClass()` as the bound, line 60 returns `Class.class` as the value type, 26 lines apart.

This also resolves our disagreement on the `MCRClassProperty` thread. Your default is right on the merits; my objection came from the name. `bound() default Object.class` would be uncontroversial, because "no bound" is a natural default, whereas `valueClass() default Object.class` reads as "the injected value is an Object", which is not what happens. With the conversion feature not merged yet, this is a cheap moment to rename.

### Conclusion

The design question as I originally framed it, whether this explodes with every feature, is answered. Dropping it.

The two parts I would keep are independent of the annotation family and fit well as follow ups, in the same way you are already handling `MCRSimpleJobSelector` and `MCRValueMapSourceBase` in #3074:

1. Derive `type()` from `annotationClass()`. Small change, concrete defect prevention, evidenced by a real bug in this PR.
2. Rename the attribute on the `MCRClassProperty` annotations before the conversion feature gives `valueClass` a second meaning.

Neither should hold up this PR.

mycore-base/src/main/java/org/mycore/common/config/annotation/MCRClassProperty.java
sebhofmann: Agreed, your reasoning holds. `valueClass` here is an opt-in restriction rather than a claim about the injected type, so `Object.class` is the right no-op default and the comparison with `MCRInstance` was not apt.

What is left of my objection is the attribute name, not the default, and that belongs to the conversion feature rather than to this PR. I have written it up on the `MCRSource` thread. Resolving this one.

mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRValueMapSourceBase.java
sebhofmann: Thanks, tracking it in #3074 then. Resolving here.

mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRValueListSourceBase.java
toKrause: Well, having a good exception message that includes the offending value didn't happen just by luck, but I agree, we should not count on the hope, that future extractors also make an effort to create good exception messages. Also, unaccounted for unchecked exceptions may always be possible (as the `LinkageError`-example shows), so an adjusted context description is a good idea for short form lists and maps.

I will include this in https://github.com/MyCoRe-Org/mycore/pull/3074 as well (where the map source is refactored in a way that makes implementing this easy).

mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRSource.java
toKrause: Good suggestions. I don't like `bound` as a attribute name though; but I also have no really good name in mind. I'll replace the manually assigned return values of `type()` with a lookup as suggested.

Beyond than that, having one simple class per annotation class that spells out the behavior is the intended approach; there are other measures for simplicity than just counting LOCs. No need for additional abstraction layers here (imho). `Mapper` is just an internal clutch that facilitates the mapping from enum constants to correctly typed source implementations. It should not be burdened with additional tasks.

mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRValueListSourceBase.java
toKrause: See https://github.com/MyCoRe-Org/mycore/pull/3074/changes/9104ca6889225a6e67547cb28d7c9ba76c06ca33

mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRClassPropertyExtractor.java
```diff
@@ -0,0 +1,52 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with MyCoRe.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.mycore.common.config.instantiator.source;
+
+import org.mycore.common.MCRClassTools;
+
+/**
+ * A {@link MCRClassPropertyExtractor} is a {@link MCRPropertyExtractor} that uses
```
sebhofmann: Javadoc says the wrong supertype. The class implements `MCRValueExtractor<Class<?>>`, so it is not a `MCRPropertyExtractor`.

```java
 * A {@link MCRClassPropertyExtractor} is a {@link MCRValueExtractor} that uses
 * {@link MCRClassTools#forName(String)} to extract {@link Class} values.
```

Both link targets resolve, so javadoc does not fail, the statement is just not true.

mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRPropertyExtractor.java
```diff
@@ -0,0 +1,32 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with MyCoRe.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.mycore.common.config.instantiator.source;
+
+/**
+ * A {@link MCRClassPropertyExtractor} is a {@link MCRPropertyExtractor} that extracts
```
sebhofmann: Same copy and paste, but here the subject is wrong too: the sentence describes `MCRClassPropertyExtractor` while sitting on `MCRPropertyExtractor`.

```java
 * A {@link MCRPropertyExtractor} is a {@link MCRValueExtractor} that extracts
 * {@link String} values.
```

mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRSourceContext.java
```diff
@@ -93,13 +93,12 @@ public MCRConfigurationException configurationException(String exceptionMessage)
...
-    public MCRConfigurationException configurationException(String exceptionMessage, Exception exception) {
-        return new MCRConfigurationException(exceptionMessage(exceptionMessage), exception);
+    public MCRConfigurationException configurationException(String exceptionMessage, Throwable cause) {
+        return new MCRConfigurationException(exceptionMessage(exceptionMessage), cause);
     }
 
-    public MCRConfigurationException incompatibilityException(
-        Class<?> annotationValueClass, Object instance) {
-        return configurationException("has a class (" + instance.getClass().getName() + "),"
+    public MCRConfigurationException incompatibilityException(Class<?> annotationValueClass, Class<?> actualClass) {
```
sebhofmann, line 100: The wording did not follow the `superClass` rename in ce559322ab. The parameter is still called `annotationValueClass` and the message still reads *"incompatible with the annotated value class (X)"*, but there is no `valueClass` attribute on the class value annotations any more.

The method has two callers and the text now fits only one of them:

| Caller | argument | wording fits |
|---|---|---|
| `MCRSourceBase:123` | `configuration.valueClass()` | yes |
| `MCRClassPropertyExtractor:39` | `superClass` | no |

The error message is where an operator meets the attribute name for the first time, so this gives back part of what the rename bought. Two ways out: a separate method for the bound case, or neutral wording that names no attribute at all, for example *"is not assignable to (X)"*.

Note that `MCRInstantiatorClassPropertyBasicTest` asserts the current wording in three places (lines 72, 125 and 179), so a change touches those as well.

sebhofmann (commented):
Looked over ace2376987 and ce559322ab. Both do the job, and the `type()` default method on `MCRSource` is a better placement than what I suggested, since it also covers `MCRPostConstructionSource` and `MCRRawPropertiesSource`. I also checked the one thing that gets structurally weaker, namely that a forgotten `Type` constant is no longer a compile error: sources are only ever built through `MCRInstantiator:327` iterating `Type.values()`, so an unregistered annotation yields a source that is never constructed and `fromAnnotation` cannot realistically throw. The trade is clearly worth it.

`superClass` also reads correctly at the call site in `MCRRestV2App`: the absence of the attribute now says "no static bound, checked dynamically in `get()`", where `valueClass() default Object.class` used to claim "the injected value is an Object".

Three small things from the two commits, none of them blocking. Two are plain javadoc slips, the third is the error message wording that did not follow the rename.

Two optional nits on the lookup table that I did not put inline, since they are cosmetic: `TYPES_BY_ANNOTATION` could be made unmodifiable after the static block, and the block reads `type.mapper.annotationClass` directly while `Mapper.toSource` uses the `annotationClass()` accessor.

One item from my original review is still open and is not part of these two commits: `MCRInstanceConfiguration.ofClassName:90` still throws `MCRException("Failed to load class " + className)` without the property name, and neither it nor `MCRInstantiatorUtils.getClass:42` catches `LinkageError`. The fix from the extractor thread landed on one of the three class loading paths. It was a remark outside the diff in the first place, so a follow up ticket seems the right home for it.

sebhofmann (approved):

PR #3074, opened 2026-07-27
MCR-3787 Harmonize handling of instance and property annotations

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3787).

## Ticket & Documentation

## Testing

## MCR Conventions & Metadata

## Multi-Repo Considerations


sebhofmann (commented):
Reviewed the branch locally. All 1274 tests in `mycore-base` pass and Checkstyle is clean.

One finding blocks: the short form insertion point in `MCRValueListSourceBase` can throw `IndexOutOfBoundsException`. See the inline comment for the reproduction.

One point has no line to attach to. `MCRInstanceConfiguration.nested(Class, String)`, both `nestedMap(...)` overloads, `MCRSourceContext.incompatibilityException` and `MCRSourceContext.orderedKeys(Map)` are public API shipped in v2026.06.0 and are removed here, while neither commit message marks a breaking change. Nothing outside `mycore-base` uses them in my workspace, so this is about the marker, not about the removal itself. Along with them `MCRInstanceConfigurationTest` (352 lines) is dropped, which leaves the `nestedMap` edge cases (empty prefix, mixed nesting depth) uncovered.

The remaining inline comments are nits.

mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRValueListSourceBase.java
```diff
@@ -38,54 +45,48 @@ abstract sealed class MCRValueListSourceBase<Value> extends MCRSourceBase<List<V
...
+    protected final List<Value> getResult(MCRSourceContext context, Map<String, String> properties,
+        Map<String, String> fullProperties) {
+
+        List<Value> list = new ArrayList<>();
+
+        int negativeKeyCount = 0;
+        String elementDescription = context.description() + " element";
+        for (String key : context.orderedIntegerKeys(nextNestedKeys(properties))) {
+            if (key.charAt(0) == '-') {
+                negativeKeyCount++;
```
sebhofmann, line 57: `negativeKeyCount` counts keys, not inserted elements. A key starting with `-` is counted here even when the element is rejected by the sentinel or extracted to `null` below and therefore never reaches `list`. Line 81 then inserts the short form elements too far to the right, and once the count exceeds the list size it throws:

```
Foo.Values          = B,C
Foo.Values.-1.Extra = X
```
```
java.lang.IndexOutOfBoundsException: Index: 1, Size: 0
	at ...MCRValueListSourceBase.getResult(MCRValueListSourceBase.java:81)
```

Reproduced locally with a throwaway test. This is reachable whenever a list element is disabled by blanking `Foo.Values.-1=` while a sub-property below it stays in place, since `MCRConfiguration2` filters blank values but keeps the rest. Counting inside the `value != null` branch fixes it.

mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRSourceBase.java
```diff
@@ -82,64 +99,77 @@ public final Result get(MCRInstanceConfiguration<?> configuration, MCRTarget tar
...
+            reducedProperties.put(reducedKey, entry.getValue());
+        }
+        String directProperty = properties.get(prefix);
+        if (directProperty != null) {
+            reducedProperties.put("", directProperty);
+        }
+        return reducedProperties;
+    }
+
+    protected static Set<String> nextNestedKeys(Map<String, String> properties) {
```
sebhofmann, line 123: This turns intermediate segments of deeper keys into list and map keys. For lists that reaches `orderedIntegerKeys`, so

```
Foo.Values.1        = A
Foo.Values.Oops.Sub = X
```

now throws `has element with non-integer key Oops`, while on main the `indexOf('.', keyPrefixLength) == -1` filter dropped such keys silently. For instance lists the stricter behaviour looks right, for `@MCRPropertyList` it is a behaviour change that can surface on update in an application with a grown property file. If it is intended, it seems worth naming in the ticket and the documentation.

mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRSourceBase.java
```diff
@@ -28,8 +31,20 @@
...
-abstract sealed class MCRSourceBase<Result> implements MCRSource permits MCRInstanceListSource, MCRInstanceMapSource,
-    MCRInstanceSource, MCRValueListSourceBase, MCRValueMapSourceBase, MCRValueSourceBase {
+/**
+ * A {@link MCRSourceBase} is a base implementation of {@link MCRSource} that
+ * handles basic aspects for obtaining a value for annotation based injection from properties, i.e.
+ * <ul>
+ *   <li>whether annotation name are allowed</li>
+ *   <li>whether the annotation name is absolute or not,</li>
+ *   <li>whether falling back to a default value is necessary, if no value is configured</li>
+ *   <li>whether an exception is thrown id no value and no default value is configured.</li>
```
sebhofmann, line 41: Two typos in this list. "whether annotation name are allowed" does not parse, presumably it is about the empty annotation name. And "thrown id no value" should be "if".

mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRSourceBase.java
```diff
@@ -44,22 +59,24 @@ abstract sealed class MCRSourceBase<Result> implements MCRSource permits MCRInst
...
             throw MCRInstantiatorUtils.emptyNameException(target);
         }
 
-        if (absolute()) {
+        if (absoluteName()) {
             context = new MCRSourceContext(target, name, "absolute " + description());
-            result = getResult(context, configuration, configuration.fullProperties(), name);
-        } else if (name.isEmpty()) {
+            result = getResult(context, reduceProperties(fullProperties, name), fullProperties);
+        } else if (supportsEmptyName() && name.isEmpty()) {
```
sebhofmann, line 74: `supportsEmptyName()` is redundant here, the guard in line 67 already rejects `name.isEmpty() && !supportsEmptyName()`.

mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRSourceBase.java
```diff
@@ -82,64 +99,77 @@ public final Result get(MCRInstanceConfiguration<?> configuration, MCRTarget tar
 
     }
 
+    protected static Map<String, String> reduceProperties(Map<String, String> properties, String prefix) {
```
sebhofmann, line 102: This is line for line identical to the private `MCRInstanceConfiguration.reduceProperties` (line 194 there). One of the two should call the other.

mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRValueListSourceBase.java
```diff
@@ -19,16 +19,23 @@
...
-import org.mycore.common.config.instantiator.MCRInstanceConfiguration;
+import org.mycore.common.config.annotation.MCRSentinel;
 
+/**
+ * A {@link MCRValueListSourceBase} is a base implementation of {@link MCRSource} that
+ * obtains a {@link List} of values for annotation based injection from properties.
+ * It provides support for {@link MCRSentinel} for each list element and
+ * uses a {@link MCRValueExtractor} to obtain each list element value from the corresponding nested properties.
+ *
+ * @param <Value> the type of injected map values.
```
sebhofmann, line 35: Copied from the map class. This is the list base, so `@param <Value>` should describe the list element type.

mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRValueListSourceBase.java
```diff
@@ -38,54 +45,48 @@ abstract sealed class MCRValueListSourceBase<Value> extends MCRSourceBase<List<V
...
-                if (index == -1) {
-                    listProperties.put(key.substring(keyPrefixLength), value);
-                }
+    protected final List<Value> getResult(MCRSourceContext context, Map<String, String> properties,
+        Map<String, String> fullProperties) {
+
+        List<Value> list = new ArrayList<>();
+
+        int negativeKeyCount = 0;
+        String elementDescription = context.description() + " element";
```
sebhofmann, line 54: `context.description()` already contains the hints, and `nested()` passes the same hints along, so a hint on the parent context would show up twice in the nested description. Short form and long form do not mix today, so nothing surfaces, but the duplication is built in.

mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRValueListSourceBase.java
```diff
@@ -38,54 +45,48 @@ abstract sealed class MCRValueListSourceBase<Value> extends MCRSourceBase<List<V
...
-        List<String> keyList = context.orderedKeys(listProperties);
-        for (String key : keyList) {
-            MCRSourceContext nestedContext = context.nested(key, context.description() + " element");
-            if (!rejectedBySentinel(nestedContext, properties, keyPrefix + key + ".")) {
-                if (key.charAt(0) == '-') {
-                    headPropertyList.add(extractor.toValue(nestedContext, listProperties.get(key)));
-                } else {
-                    tailPropertyList.add(extractor.toValue(nestedContext, listProperties.get(key)));
+            MCRSourceContext nestedContext = context.nested(key, elementDescription);
+            Map<String, String> nestesProperties = reduceProperties(properties, key);
```
sebhofmann, line 60: Typo: `nestesProperties` should be `nestedProperties`, here and in lines 61 and 62.

mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRValueMapSourceBase.java
```diff
@@ -37,40 +45,40 @@ abstract sealed class MCRValueMapSourceBase<Value> extends MCRSourceBase<Map<Str
...
+        }
 
-        for (String key : mapProperties.keySet()) {
-            MCRSourceContext nestedContext = context.nested(key, context.description() + " entry");
-            if (!rejectedBySentinel(nestedContext, properties, keyPrefix + key + ".")) {
-                map.put(key, extractor.toValue(nestedContext, mapProperties.get(key)));
+        String entryDescription = context.description() + " entry";
+        for (String key : nextNestedKeys(properties)) {
+            MCRSourceContext nestedContext = context.nested(key, entryDescription);
+            Map<String, String> nestesProperties = reduceProperties(properties, key);
```
sebhofmann, line 67: Typo: `nestesProperties` should be `nestedProperties`, here and in lines 68 and 69.

mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRValueSourceBase.java
```diff
@@ -21,10 +21,17 @@
 import java.util.Map;
 
 import org.mycore.common.config.MCRConfigurationException;
-import org.mycore.common.config.instantiator.MCRInstanceConfiguration;
 
+/**
+ * A {@link MCRValueSourceBase} is a base implementation of {@link MCRSource} that
+ * obtains a single value for annotation based injection from properties.
+ * It provides support for {@link MCRSentinel} and
```
sebhofmann, line 28: `MCRSentinel` is not imported in this file, so the link stays unresolved. `MCRValueListSourceBase` and `MCRValueMapSourceBase` gained the import for exactly this reason.

mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRInstanceExtractor.java
```diff
@@ -0,0 +1,66 @@
...
+    MCRInstanceExtractor(Class<?> superClass) {
+        this.superClass = superClass;
+    }
+
+    @Override
+    public Object toValue(MCRSourceContext context, Map<String, String> properties,
+        Map<String, String> fullProperties) {
+
+        MCRInstanceConfiguration<?> configuration = MCRInstanceConfiguration.ofComponents(
+            superClass, MCRInstanceName.of(context.property()), properties, fullProperties);
```
sebhofmann, line 47: The name is recovered from the property string here, while the previous code built it structurally via `name.nested(prefix)`. `MCRInstanceName.of` strips a trailing `.Class`, so an instance map entry keyed `Class` ends up with a name short by one segment in messages and in the `[IMPLICIT]` log. Pathological, but avoidable.

mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRInstanceExtractor.java
```diff
@@ -0,0 +1,66 @@
...
+
+        if (!configuration.instantiatable()) {
+            if (LOGGER.isInfoEnabled()) {
+                LOGGER.info("[CLEAN-UP] Ignoring {}, configured in {} (and sub-properties thereof), " +
+                    "because {}.Class is missing or empty)", context.description(), context.property(),
```
sebhofmann, line 52: Stray closing parenthesis at the end of "because {}.Class is missing or empty)".

mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRClassPropertyExtractor.java
```diff
@@ -28,17 +30,26 @@ final class MCRClassPropertyExtractor implements MCRValueExtractor<Class<?>> {
 
     private final Class<?> superClass;
 
-    MCRClassPropertyExtractor(Class<?> superClass) {
-        this.superClass = superClass;
+    MCRClassPropertyExtractor(Class<?> valueClass) {
```
sebhofmann, line 33: The parameter is named `valueClass` and lands in a field named `superClass`, and `MCRInstanceExtractor` does the same with `annotation.valueClass()`. Both names denote the same class today, so this is only noise, but with the announced conversion feature for `@MCRProperty` the two meanings diverge. Cheap to settle here.

mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRValueListSourceBase.java
toKrause: Fixed.

mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRSourceBase.java
toKrause: I've added this as an example benefit of the harmonization in the ticket description.

mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRSourceBase.java
toKrause: Fixed.

mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRSourceBase.java
toKrause: Fixed.

mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRValueMapSourceBase.java
toKrause: Fixed.

mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRValueSourceBase.java
toKrause: Fixed.

mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRInstanceExtractor.java
toKrause: What?

mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRInstanceExtractor.java
toKrause: Fixed.

mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRClassPropertyExtractor.java
toKrause: Fixed.

mycore-base/src/main/java/org/mycore/common/config/instantiator/source/MCRSourceBase.java
toKrause: Both occurrences are going to be removed in [MCR-3703](https://github.com/MyCoRe-Org/mycore/pull/3075). I've added a copy temporarily, so I do not have make one method public and remove a public method in the next PR.

PR #3081, opened 2026-07-30
MCR-3710 Modernize MCRIIIFImageImpl and MCRIIIFPresentationImpl instantiation scheme

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3710).

## Ticket & Documentation

## Testing

## MCR Conventions & Metadata

## Multi-Repo Considerations

### DIFF
--- a/mycore-base/src/test/java/org/mycore/common/config/MCRInstantiatorProxyFineGrainedTest.java
+++ b/mycore-base/src/test/java/org/mycore/common/config/MCRInstantiatorProxyFineGrainedTest.java
@@ -1,0 +1,141 @@
+/*
+ * This file is part of ***  M y C o R e  ***
+ * See https://www.mycore.de/ for details.
+ *
+ * MyCoRe is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * MyCoRe is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with MyCoRe.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.mycore.common.config;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.mycore.common.config.instantiator.MCRInstanceConfiguration.ofName;
+
+import java.util.function.Supplier;
+
+import org.junit.jupiter.api.Test;
+import org.mycore.common.MCRTestConfiguration;
+import org.mycore.common.MCRTestProperty;
+import org.mycore.common.config.annotation.MCRConfigurationProxy;
+import org.mycore.common.config.annotation.MCRProperty;
+import org.mycore.test.MyCoReTest;
+
+@MyCoReTest
+public class MCRInstantiatorProxyFineGrainedTest {
+
+    @Test
+    @MCRTestConfiguration(
+        properties = {
+            @MCRTestProperty(key = "Foo.Class", classNameOf = TestClass.class),
+            @MCRTestProperty(key = "Foo.Property1", string = "Value1"),
+            @MCRTestProperty(key = "Foo.Property2", string = "Value2")
+        })
+    public void classWithProxy() {
+
+        TestClass instance = ofName(TestClass.class, "Foo").instantiate();
+
+        assertNotNull(instance);
+        assertEquals("Value1-Value2", instance.value());
+
+    }
+
+    @Test
+    @MCRTestConfiguration(
+        properties = {
+            @MCRTestProperty(key = "Foo.Class", classNameOf = ExtendedTestClass.class),
+            @MCRTestProperty(key = "Foo.Property1", string = "Value1"),
+            @MCRTestProperty(key = "Foo.Property2", string = "Value2"),
+            @MCRTestProperty(key = "Foo.AdditionalProperty", string = "AdditionalValue")
+        })
+    public void subClassWithProxy() {
+
+        ExtendedTestClass instance = ofName(ExtendedTestClass.class, "Foo").instantiate();
+
+        assertNotNull(instance);
+        assertEquals("Value1-Value2", instance.value());
+        assertEquals("AdditionalValue", instance.additionalValue());
+
+    }
+
+    @MCRConfigurationProxy(proxyClass = TestClass.Factory.class)
+    public static class TestClass {
+
+        private final String value;
+
+        public TestClass(String value) {
+            this.value = value;
+        }
+
+        public String value() {
+            return value;
+        }
+
+        public static abstract class FactoryBase {
+
+            @MCRProperty(name = "Property1")
+            public String value1;
+
+            @MCRProperty(name = "Property2")
+            public String value2;
+
+            public final String getValue() {
+                return value1 + "-" + value2;
+            }
+
+        }
+
+        public static final class Factory extends FactoryBase implements Supplier<TestClass> {
+
+            @Override
+            public TestClass get() {
+                return new TestClass(getValue());
+            }
+
+        }
+
+    }
+
+    @MCRConfigurationProxy(proxyClass = ExtendedTestClass.Factory.class)
+    public static class ExtendedTestClass extends TestClass {
+
+        private final String additionalValue;
+
+        public ExtendedTestClass(String value, String additionalValue) {
+            super(value);
+            this.additionalValue = additionalValue;
+        }
+
+        public String additionalValue() {
+            return additionalValue;
+        }
+
+        public static abstract class FactoryBase extends TestClass.FactoryBase {
+
+            @MCRProperty(name = "AdditionalProperty")
+            public String additionalValue;
+
+        }
+
+        public static final class Factory extends FactoryBase implements Supplier<ExtendedTestClass> {
+
+            @Override
+            public ExtendedTestClass get() {
+                return new ExtendedTestClass(getValue(), additionalValue);
+            }
+
+        }
+
+    }
+
+}
--- a/mycore-iiif/src/main/java/org/mycore/iiif/image/impl/MCRIIIFImageImpl.java
+++ b/mycore-iiif/src/main/java/org/mycore/iiif/image/impl/MCRIIIFImageImpl.java
@@ -19,15 +19,9 @@
 package org.mycore.iiif.image.impl;
 
 import java.awt.image.BufferedImage;
-import java.lang.reflect.Constructor;
-import java.lang.reflect.InvocationTargetException;
-import java.util.HashMap;
-import java.util.Map;
 
 import org.mycore.access.MCRAccessException;
-import org.mycore.common.MCRException;
 import org.mycore.common.config.MCRConfiguration2;
-import org.mycore.common.config.MCRConfigurationException;
 import org.mycore.iiif.image.model.MCRIIIFImageInformation;
 import org.mycore.iiif.image.model.MCRIIIFImageProfile;
 import org.mycore.iiif.image.model.MCRIIIFImageQuality;
@@ -39,7 +33,7 @@ public abstract class MCRIIIFImageImpl {
 
     private static final String MCR_IIIF_IMAGE_CONFIG_PREFIX = "MCR.IIIFImage.";
 
-    private static final Map<String, MCRIIIFImageImpl> IMPLHOLDER = new HashMap<>();
+    protected static final String DEFAULT_PROPERTY_PREFIX = "MCR.Default.IIIFImage.";
 
     private final String implName;
 
@@ -47,43 +41,19 @@ public abstract class MCRIIIFImageImpl {
         this.implName = implName;
     }
 
-    public static synchronized MCRIIIFImageImpl obtainInstance(String implNameParameter) {
-        String implName = (implNameParameter == null || implNameParameter.isBlank())
+    public static MCRIIIFImageImpl obtainInstance(String implName) {
+
+        String checkedImplName = (implName == null || implName.isBlank())
             ? MCRConfiguration2.getStringOrThrow("MCR.IIIFImage.Default")
-            : implNameParameter;
+            : implName;
 
-        if (IMPLHOLDER.containsKey(implName)) {
-            return IMPLHOLDER.get(implName);
-        }
+        String implPropertyName = MCR_IIIF_IMAGE_CONFIG_PREFIX + checkedImplName;
+        return MCRConfiguration2.getSingleInstanceOfOrThrow(MCRIIIFImageImpl.class, implPropertyName);
 
-        String classPropertyName = MCR_IIIF_IMAGE_CONFIG_PREFIX + implName + ".Class";
-        Class<? extends MCRIIIFImageImpl> classObject = MCRConfiguration2.<MCRIIIFImageImpl>getClass(classPropertyName)
-            .orElseThrow(() -> MCRConfiguration2.createConfigurationException(classPropertyName));
-
-        try {
-            Constructor<? extends MCRIIIFImageImpl> constructor = classObject.getConstructor(String.class);
-            MCRIIIFImageImpl imageImpl = constructor.newInstance(implName);
-            IMPLHOLDER.put(implName, imageImpl);
-            return imageImpl;
-        } catch (NoSuchMethodException e) {
-            throw new MCRConfigurationException(
-                "Configurated class (" + classObject.getName() + ") needs a string constructor: " + classPropertyName,
-                e);
-        } catch (IllegalAccessException | InstantiationException | InvocationTargetException e) {
-            throw new MCRException(e);
-        }
     }
 
     public String getImplName() {
         return implName;
-    }
-
-    protected final Map<String, String> getProperties() {
-        return MCRConfiguration2.getSubPropertiesMap(getConfigPrefix());
-    }
-
-    protected String getConfigPrefix() {
-        return MCR_IIIF_IMAGE_CONFIG_PREFIX + implName + ".";
     }
 
     public abstract BufferedImage provide(String identifier,
--- a/mycore-iiif/src/main/java/org/mycore/iiif/image/impl/MCRIIIFImageImpl.java
+++ b/mycore-iiif/src/main/java/org/mycore/iiif/image/impl/MCRIIIFImageImpl.java
@@ -22,6 +22,7 @@ import java.awt.image.BufferedImage;
 
 import org.mycore.access.MCRAccessException;
 import org.mycore.common.config.MCRConfiguration2;
+import org.mycore.common.config.annotation.MCRPostConstruction;
 import org.mycore.iiif.image.model.MCRIIIFImageInformation;
 import org.mycore.iiif.image.model.MCRIIIFImageProfile;
 import org.mycore.iiif.image.model.MCRIIIFImageQuality;
@@ -69,5 +70,16 @@ public abstract class MCRIIIFImageImpl {
         throws MCRIIIFImageNotFoundException, MCRIIIFImageProvidingException, MCRAccessException;
 
     public abstract MCRIIIFImageProfile getProfile();
+
+    public static abstract class FactoryBase {
+
+        public String implName;
+
+        @MCRPostConstruction(MCRPostConstruction.Value.TRAILING_NAME)
+        public void setImplName(String implName) {
+            this.implName = implName;
+        }
+
+    }
 
 }
--- a/mycore-iiif/src/main/java/org/mycore/iiif/image/impl/MCRIIIFImageImpl.java
+++ b/mycore-iiif/src/main/java/org/mycore/iiif/image/impl/MCRIIIFImageImpl.java
@@ -19,16 +19,10 @@
 package org.mycore.iiif.image.impl;
 
 import java.awt.image.BufferedImage;
-import java.lang.reflect.Constructor;
-import java.lang.reflect.InvocationTargetException;
-import java.util.HashMap;
-import java.util.Map;
 
 import org.mycore.access.MCRAccessException;
-import org.mycore.common.MCRException;
 import org.mycore.common.config.MCRConfiguration2;
-import org.mycore.common.config.MCRConfigurationBase;
-import org.mycore.common.config.MCRConfigurationException;
+import org.mycore.common.config.annotation.MCRPostConstruction;
 import org.mycore.iiif.image.model.MCRIIIFImageInformation;
 import org.mycore.iiif.image.model.MCRIIIFImageProfile;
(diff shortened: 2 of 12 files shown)


mycore-iview2/src/main/resources/components/iview2/config/mycore.properties
```diff
@@ -35,6 +35,9 @@ MCR.LayoutTransformerFactory.Default.Ignore=%MCR.LayoutTransformerFactory.Defaul
 
 # the default impl is Iview
 MCR.IIIFImage.Default=Iview
+# 1024*1024
+MCR.IIIFImage.Default.IdentifierSeparator=1048576
+MCR.IIIFImage.Default.MaxImageBytes=/
```
sebhofmann: **Blocker: the two values are swapped.**

This is not just cosmetic, both properties are actually reached at runtime:

* Neither `Iview` nor `thumbnail` defines `IdentifierSeparator`, so the default always applies. `MCRIVIEWIIIFImageImpl#createTileInfo` uses it as a regex in `id.split(identifierSeparator, 2)`. With `1048576`, an identifier such as `mir_derivate_00000001/image.jpg` never splits, `splittedIdentifier.length == 1`, and the resulting `MCRTileInfo` carries `derivate == null`. Image delivery breaks for every derivate qualified identifier. The previous code hardcoded `"/"` as the fallback.
* `MaxImageBytes=/` makes `Long.parseLong` throw for any implementation that does not set the property itself. `Iview`, `thumbnail` and MIR all set it, but custom implementations relying on the default would fail on instantiation.

Suggested fix, with the comment moved to the line it belongs to:

```suggestion
MCR.IIIFImage.Default.IdentifierSeparator=/
# 1024*1024
MCR.IIIFImage.Default.MaxImageBytes=1048576
```

mycore-iiif/src/main/java/org/mycore/iiif/image/impl/MCRIIIFImageImpl.java
```diff
@@ -39,53 +34,29 @@ public abstract class MCRIIIFImageImpl {
 
     private static final String MCR_IIIF_IMAGE_CONFIG_PREFIX = "MCR.IIIFImage.";
 
-    private static final Map<String, MCRIIIFImageImpl> IMPLHOLDER = new HashMap<>();
+    protected static final String MCR_IIIF_IMAGE_DEFAULT_CONFIG_PREFIX = MCR_IIIF_IMAGE_CONFIG_PREFIX + "Default.";
```
sebhofmann: `MCR.IIIFImage.Default` now carries two meanings: it is the property holding the name of the default implementation (`MCR.IIIFImage.Default=Iview`), and it is the prefix under which default values for all implementations live (`MCR.IIIFImage.Default.MaxImageBytes`).

An implementation actually named `Default` would serve both roles at once, and even without that the two are easy to confuse when reading a properties file. A distinct prefix such as `MCR.IIIFImage.Defaults.` would keep the namespaces apart.

mycore-iiif/src/main/java/org/mycore/iiif/image/impl/MCRIIIFImageImpl.java
```diff
@@ -39,53 +34,29 @@ public abstract class MCRIIIFImageImpl {
...
 
     private final String implName;
 
     public MCRIIIFImageImpl(final String implName) {
         this.implName = implName;
     }
 
-    public static synchronized MCRIIIFImageImpl obtainInstance(String implNameParameter) {
-        String implName = (implNameParameter == null || implNameParameter.isBlank())
+    public static synchronized MCRIIIFImageImpl obtainInstance(String implName) {
```
sebhofmann: Nit: `synchronized` guarded the removed `IMPLHOLDER` map. Caching is now handled by `MCRConfiguration2.getSingleInstanceOfOrThrow`, so the modifier can go.

ruleset.xml
```diff
@@ -15,7 +15,11 @@
   <rule ref="category/xsl/codestyle.xml/UseConcatOnce"/>
 
   <!-- Best Practices Rules -->
-  <rule ref="category/java/bestpractices.xml/AbstractClassWithoutAbstractMethod"/>
+  <rule ref="category/java/bestpractices.xml/AbstractClassWithoutAbstractMethod">
+    <properties>
+      <property name="violationSuppressXPath" value="//ClassDeclaration[matches(@SimpleName, 'FactoryBase') or matches(@BinaryName, 'FactoryBase')]" />
```
sebhofmann: This suppression is broader than intended. I verified it with a probe class in `mycore-iview2`, and PMD stays silent for both of these:

```java
// suppressed, because matches() is unanchored
public abstract static class SomeFactoryBaseHelper { public int x; }

// suppressed, because @BinaryName contains "FactoryBase"
public abstract static class FactoryBase {
    public abstract static class Inner { public int z; }
}
```

So any class whose simple name merely *contains* `FactoryBase`, and anything nested inside a `FactoryBase`, silently loses both rules.

Two options:

1. Anchor the expression, e.g. `matches(@SimpleName, '^FactoryBase$')`, and either anchor the `@BinaryName` clause to `\.FactoryBase$` or drop it.
2. Drop the ruleset change entirely and put `@SuppressWarnings("PMD.AbstractClassWithoutAbstractMethod")` on the three factory base classes. Local, explicit, and it does not turn a naming convention into a project wide rule opt out.

I would lean towards option 2.

ruleset.xml
```diff
@@ -115,7 +119,11 @@
   <rule ref="category/java/codestyle.xml/UseUnderscoresInNumericLiterals"/>
 
   <!-- Design Rules -->
-  <rule ref="category/java/design.xml/AbstractClassWithoutAnyMethod"/>
+  <rule ref="category/java/design.xml/AbstractClassWithoutAnyMethod">
+    <properties>
+      <property name="violationSuppressXPath" value="//ClassDeclaration[matches(@SimpleName, 'FactoryBase') or matches(@BinaryName, 'FactoryBase')]" />
```
sebhofmann: Same over broad suppression as for `AbstractClassWithoutAbstractMethod` above.

mycore-base/src/test/java/org/mycore/common/config/MCRInstantiatorProxyFineGrainedTest.java
```diff
@@ -0,0 +1,144 @@
...
+
+            @MCRProperty(name = "AdditionalProperty")
+            public String additionalValue;
+
+        }
+
+        public static final class Factory extends FactoryBase implements Supplier<ExtendedTestClass> {
+
+            @MCRProperty(name = "AdditionalProperty")
+            public String additionalValue;
```
sebhofmann: `ExtendedTestClass.FactoryBase` already declares `additionalValue` with the same `@MCRProperty`. Redeclaring it here shadows the inherited field, and since `Class#getFields()` returns both, the instantiator injects both.

It happens to work, but it undermines exactly what this test is meant to demonstrate, namely that the properties live in the `FactoryBase` and the `Factory` only wires the constructor call. Removing these two lines keeps the test honest.

mycore-iview2/src/main/java/org/mycore/iview2/iiif/MCRIVIEWIIIFImageImpl.java
```diff
@@ -61,38 +65,50 @@
 import org.mycore.iview2.backend.MCRTileInfo;
 import org.mycore.iview2.services.MCRIView2Tools;
 
+@MCRConfigurationProxy(proxyClass = MCRIVIEWIIIFImageImpl.Factory.class )
```
sebhofmann: The Eclipse formatter (`mycore-javastyle.xml`) has not been run on this file. It changes three spots, none of which Checkstyle catches:

* here: stray space before the closing parenthesis of the annotation
* line 99: `private final   String identifierSeparator;`
* line 110: `protected final  MCRTileFileProvider tileFileProvider() {`

mycore-iview2/src/main/java/org/mycore/iview2/iiif/MCRIVIEWIIIFImageImpl.java
```diff
@@ -61,38 +65,50 @@
...
-                MCRTileFileProvider.class, getConfigPrefix() + TILE_FILE_PROVIDER_PROPERTY);
-        }
+        this.tileFileProvider = tileFileProvider;
+        this.transparentFormats = transparentFormats;
+        this.maxImageSize = maxImageSize;
+        this.identifierSeparator = identifierSeparator;
+    }
 
-        transparentFormats = Arrays.asList(properties.get("TransparentFormats").split(","));
+    protected final  MCRTileFileProvider tileFileProvider() {
```
sebhofmann: Naming nit: this is `tileFileProvider()` while the factory right below uses `getTileFileProvider()`, and the rest of the class uses `getXxx()`. Suggest `getTileFileProvider()` for consistency.

mycore-iview2/src/main/java/org/mycore/iview2/iiif/MCRIVIEWIIIFImageImpl.java
```diff
@@ -352,4 +364,35 @@ protected boolean checkPermission(String identifier, MCRTileInfo tileInfo) {
...
+        @MCRPropertyList(name = TRANSPARENT_FORMATS_PROPERTY, required = false)
+        public List<String> transparentFormats;
+
+        @MCRProperty(name = MAX_BYTES_PROPERTY, defaultName = MAX_BYTES_DEFAULT_PROPERTY)
+        public String maxImageSize;
+
+        @MCRProperty(name = IDENTIFIER_SEPARATOR_PROPERTY, defaultName = IDENTIFIER_SEPARATOR_DEFAULT_PROPERTY)
+        public String identifierSeparator;
+
+        public MCRTileFileProvider getTileFileProvider() {
```
sebhofmann, line 382: Minor asymmetry: `MaxImageBytes` and `IdentifierSeparator` get their defaults from `MCR.IIIFImage.Default.*`, but the `MCRTileFileProvider` default is hardcoded here. That is forced by `@MCRInstance` not offering a `defaultName`, so nothing to fix right now, but it is worth a comment in the code so the next reader does not go looking for the missing property. `MCR.IIIFImage.thumbnail` is the case that actually hits this path.

mycore-iview2/src/main/java/org/mycore/iview2/iiif/MCRIVIEWIIIFImageImpl.java
```diff
@@ -352,4 +364,35 @@ protected boolean checkPermission(String identifier, MCRTileInfo tileInfo) {
...
+        }
+
+    }
+
+    public static final class Factory extends FactoryBase implements Supplier<MCRIVIEWIIIFImageImpl> {
+
+        @Override
+        public MCRIVIEWIIIFImageImpl get() {
+            return new MCRIVIEWIIIFImageImpl(implName, getTileFileProvider(), transparentFormats,
+                Long.parseLong(maxImageSize), identifierSeparator);
```
sebhofmann, line 393: A misconfigured `MaxImageBytes` surfaces as a raw `NumberFormatException` rather than an `MCRConfigurationException`. The instantiator can only inject `String`, so the field type is fine, but wrapping the parse and naming the offending property in the message would be a lot more useful for operators. See the swapped default value in `mycore.properties` for a case where this actually triggers.

sebhofmann (changes requested):
Reviewed locally in a worktree at `cacba2d`. `mvn install -DskipTests` is green, Checkstyle reports 0 violations, PMD passes. Tests were not executed.

The direction of this change is good: replacing the reflective `getConstructor(String.class)` scheme plus the manual `IMPLHOLDER` cache with a configuration proxy removes real boilerplate, and the `FactoryBase` split is a neat answer to the `Supplier<T>` covariance problem documented in `MCRInstantiatorProxyTest`.

One blocker, though: the two new default properties have their values swapped, which breaks image delivery at runtime. Details inline.

### Summary of findings

**Blocker**
1. `MCR.IIIFImage.Default.IdentifierSeparator` and `MCR.IIIFImage.Default.MaxImageBytes` hold each other's values.

**Important**
2. `MCR.IIIFImage.Default` is now both the name of the default implementation and the prefix for default values.
3. The new PMD suppressions are broader than intended (verified empirically, see inline).
4. Duplicated `additionalValue` field in the new fine grained test.

**Minor**
5. Eclipse formatter was not run on `MCRIVIEWIIIFImageImpl`.
6. `Long.parseLong` surfaces a `NumberFormatException` instead of an `MCRConfigurationException`.
7. Accessor naming is inconsistent (`tileFileProvider()` vs `getTileFileProvider()`).
8. `obtainInstance` no longer needs to be `synchronized`.
9. The `MCRTileFileProvider` default is hardcoded in Java while the other defaults live in properties.

### Verified as correct

* The breaking changes are correctly marked with `!`. Removed or narrowed: `MCRIVIEWIIIFImageImpl(String)`, `getProperties()`, `getConfigPrefix()`, `MAX_BYTES_PROPERTY` (public to protected), `tileFileProvider` (protected to private), `DERIVATE_TYPES` renamed. No consumer inside this repository is affected, and MIR only overrides properties without subclassing.
* `@MCRPropertyList(required = false)` yields an empty list rather than `null` (`MCRSourceBase#nullResultReplacement`), so `transparentFormats.contains(...)` cannot NPE.
* `MCRPostConstruction.Value.TRAILING_NAME` resolves `MCR.IIIFImage.Iview.Class` to `Iview`, matching the semantics of the old string constructor.
* `distinct()` is preserved via `LinkedHashSet`, and iteration order stays stable for `createTileInfoForMCRObject`.
* The `deprecated.properties` mappings still apply.

### Out of scope, possible follow up

`MCRIIIFPresentationImpl#obtainInstance` still uses the old reflection scheme. Worth its own ticket so both IIIF entry points end up consistent.

ruleset.xml
```diff
@@ -15,7 +15,11 @@
   <rule ref="category/xsl/codestyle.xml/UseConcatOnce"/>
 
   <!-- Best Practices Rules -->
-  <rule ref="category/java/bestpractices.xml/AbstractClassWithoutAbstractMethod"/>
+  <rule ref="category/java/bestpractices.xml/AbstractClassWithoutAbstractMethod">
+    <properties>
+      <property name="violationSuppressXPath" value="//ClassDeclaration[matches(@SimpleName, '^FactoryBase$')" />
```
sebhofmann: The XPath is missing its closing `]`, which silently disables this rule on every file it touches.

Because the expression only lives in an XML attribute value, the file stays well formed and nothing in the build validates it up front. PMD compiles the expression lazily, the first time it has to decide whether a violation is suppressed, and then fails:

```
PmdXPathException: net.sf.saxon.trans.XPathException: expected "]", found "<eof>"
Exception Context:
	[1:Kind=Syntax error]
	[2:Phase=INITIALIZATION]
	[3:Expression=//ClassDeclaration[matches(@SimpleName, '^FactoryBase$')]
```

PMD then logs `Exception applying rule AbstractClassWithoutAbstractMethod on file ..., continuing with next rule` and drops the rule for that compilation unit. So the violation is not suppressed, it is swallowed together with the rule, and the same happens for `AbstractClassWithoutAnyMethod` below. The affected files are `MCRIIIFImageImpl`, `MCRIIIFPresentationImpl`, `MCRIVIEWIIIFImageImpl`, `MCRThumbnailImageImpl` and `MCRMetsIIIFPresentationImpl`.

The maven build only reports this as a `[WARNING] There are N PMD processing errors` and still ends in `BUILD SUCCESS`, so CI will not catch it. Reproduced on `1cf3a3ed2f` with `mvn -DskipTests -pl mycore-iiif,mycore-iview2,mycore-mets install`: 3 processing errors, build green.

```suggestion
      <property name="violationSuppressXPath" value="//ClassDeclaration[matches(@SimpleName, '^FactoryBase$')]" />
```

The anchoring itself is exactly what was needed: `^FactoryBase$` plus dropping `@BinaryName` resolves the previous round's finding about `SomeFactoryBaseHelper` and about nested classes.

ruleset.xml
```diff
@@ -115,7 +119,11 @@
   <rule ref="category/java/codestyle.xml/UseUnderscoresInNumericLiterals"/>
 
   <!-- Design Rules -->
-  <rule ref="category/java/design.xml/AbstractClassWithoutAnyMethod"/>
+  <rule ref="category/java/design.xml/AbstractClassWithoutAnyMethod">
+    <properties>
+      <property name="violationSuppressXPath" value="//ClassDeclaration[matches(@SimpleName, '^FactoryBase$')" />
```
sebhofmann: Same missing `]` here, with the same consequence for `AbstractClassWithoutAnyMethod`.

```suggestion
      <property name="violationSuppressXPath" value="//ClassDeclaration[matches(@SimpleName, '^FactoryBase$')]" />
```

mycore-iview2/src/main/resources/components/iview2/config/mycore.properties
```diff
@@ -35,6 +35,10 @@ MCR.LayoutTransformerFactory.Default.Ignore=%MCR.LayoutTransformerFactory.Defaul
 
 # the default impl is Iview
 MCR.IIIFImage.Default=Iview
+# 1024*1024
+MCR.Default.IIIFImage.MaxImageBytes=1048576
+MCR.Default.IIIFImage.IdentifierSeparator=/
+MCR.Default.IIIFPresentation.StoreMetsOnGenerate=%MCR.Mets.storeMetsOnGenerate%
```
sebhofmann: This default belongs into `mycore-mets`, not into `mycore-iview2`.

Its only consumer is `MCRMetsIIIFPresentationImpl.FactoryBase#storeMetsOnGenerate`, and `mycore-mets` does not depend on `mycore-iview2`: its MyCoRe dependencies are `mycore-base`, `mycore-iiif` and `mycore-solr`. In an application that ships `mycore-mets` without `mycore-iview2` and enables the IIIF presentation impl, neither `MCR.IIIFPresentation.Mets.StoreMetsOnGenerate` nor this default property exists. `MCRProperty#required` defaults to `true`, so `MCRSourceBase#get` (line 71) throws an `MCRConfigurationException` and the impl cannot be instantiated at all.

Moving the line to `mycore-mets/src/main/resources/components/mets/config/mycore.properties`, right next to `MCR.Mets.storeMetsOnGenerate`, fixes that and also removes the cross component `%...%` reference.

The two `MCR.Default.IIIFImage.*` lines above are fine, `MCRIVIEWIIIFImageImpl` and `MCRThumbnailImageImpl` both live in this component.

sebhofmann (changes requested):
Re-review after the force push, based on `1cf3a3ed2f`.

Thanks, most of the previous round is addressed: the blocker (swapped `IdentifierSeparator` / `MaxImageBytes` values) is fixed, the `MCR.IIIFImage.Default` namespace collision is gone thanks to the new `MCR.Default.IIIFImage.` / `MCR.Default.IIIFPresentation.` prefix, the duplicated field in the new test is gone, the formatter now runs clean over all touched files, `tileFileProvider()` became `getTileFileProvider()`, and the redundant `synchronized` on both `obtainInstance` methods is gone. Pulling `MCRIIIFPresentationImpl` into the scope is a nice bonus.

Three issues from this round. Two of them are inline, the third one cannot be anchored because the file is not part of the diff.

### Commented example config in `mycore-mets` was not updated

`mycore-mets/src/main/resources/components/mets/config/mycore.properties`, lines 47 to 49:

```properties
# MCR.IIIFPresentation.Default=Mets
# MCR.IIIFPresentation.Mets=org.mycore.mets.iiif.MCRMetsIIIFPresentationImpl
# MCR.IIIFPresentation.Mets.Transformer=mets-dfg
```

`MCRIIIFPresentationImpl#obtainInstance` now delegates to `MCRConfiguration2.getSingleInstanceOfOrThrow(MCRIIIFPresentationImpl.class, "MCR.IIIFPresentation." + checkedImplName)`, and that method resolves the class from `name + CLASS_SUFFIX` (`MCRConfiguration2:194`). The old code used the bare property name as the class property, so the second line has to become:

```properties
# MCR.IIIFPresentation.Mets.Class=org.mycore.mets.iiif.MCRMetsIIIFPresentationImpl
```

This block is the only documentation of how to wire up `MCRMetsIIIFPresentationImpl`, so it should move with the breaking change. While you are there: `MCRMetsIIIFPresentationImpl.STORE_METS_ON_GENERATE` was a public constant and is now removed, which is worth an explicit mention in the PR description next to the `!` marker.

Everything else I checked on this revision is fine: Checkstyle reports 0 violations across `mycore-base`, `mycore-iiif`, `mycore-iview2` and `mycore-mets`, the Eclipse formatter produces no diff on any of the eight touched Java files, and the `MCR.Default.IIIFImage.*` defaults are consumed only by classes that live in `mycore-iview2`, so those are in the right component.

Still open from the previous round, both minor and both non blocking: `Long.parseLong(maxImageSize)` (and now `Boolean.parseBoolean(storeMetsOnGenerate)`) turn a configuration mistake into a `NumberFormatException` respectively into a silent `false` instead of an `MCRConfigurationException`, and `FactoryBase#getTileFileProvider` still hardcodes `new MCRDefaultTileFileProvider()`. With the new default prefix in place the latter is now a one liner: `@MCRInstance(name = TILE_FILE_PROVIDER_PROPERTY, valueClass = MCRTileFileProvider.class, defaultName = DEFAULT_PROPERTY_PREFIX + TILE_FILE_PROVIDER_PROPERTY)` plus a `MCR.Default.IIIFImage.TileFileProvider.Class` entry. `MCRInstanceSource` supports `defaultName`.

toKrause:
Fixed configuration issues.

`@MCRInstance` does not support a default name. This will, however, be added in MCR-3790. `Long#parseLong` and `Boolean#parseBoolean` are used to keep the configuration aligned with `MCRConfiguration2#getLong` and `MCRConfiguration2#getBoolean`.

sebhofmann (dismissed):

mycore-iview2/src/main/resources/components/iview2/config/mycore.properties
```diff
@@ -35,6 +35,9 @@ MCR.LayoutTransformerFactory.Default.Ignore=%MCR.LayoutTransformerFactory.Defaul
 
 # the default impl is Iview
 MCR.IIIFImage.Default=Iview
+# 1024*1024
+MCR.Default.IIIFImage.MaxImageBytes=1048576
```
rsteph-de, line 39: Can/Should the 'Default' be moved in the property key?
All properties should start with`MCR.IIIIFImage.`.

This will probably have the side effect, that `Default` is treated as Implementation-ID - maybe use `_Default` - how is this dealt with elsewhere in MyCoRe.

mycore-iview2/src/main/java/org/mycore/iview2/iiif/MCRIVIEWIIIFImageImpl.java
rsteph-de: For me the NumberformatException would be O.K. - this is very rare problem.

mycore-iview2/src/main/resources/components/iview2/config/mycore.properties
rsteph-de: after discussion we keep `MCR.Default`

rsteph-de (changes requested):

rsteph-de (approved):
all open questions have been discussed and fixed

PR #3096, opened 2026-08-06
MCR-3388 Duplicate Detection and handling

sebhofmann:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3388).

Adds duplicate detection and handling to MyCoRe as two new components plus the MODS specific criteria.

## What is added

**`mycore-dedup`** (new component)
* `MCRDeDupCriterionBuilder` builds normalized deduplication keys from an object. Builders are registered per object type via `MCR.DeDup.CriterionBuilder.<type>.<name>.Class`, `MCRDeDupCriteriaProvider` aggregates them.
* `MCRDeDupKeyManager` persists the keys, the no-duplicate markings and the display titles (`MCRDeDupKey`, `MCRDeDupNoDuplicate`, `MCRDeDupTitle`). Two objects sharing a key are a possible duplicate unless a no-duplicate marking exists for the pair.
* `MCRDeDupEventHandler` (`MCR.EventHandler.MCRObject.085.Class`) keeps keys, titles and markings in sync on create, update, repair and delete.
* `MCRDeDupNoDuplicateFlagEventHandler` (`MCR.EventHandler.MCRObject.015D.Class`) turns service flags of the type `MCR.DeDup.NoDuplicateFlagType` into permanent no-duplicate markings and removes them before the metadata is written. This exists because an object that is about to be created has no id yet, so a user confirmation given in the editor can only travel as a flag.
* `MCRDeDupURIResolver` exposes the possible duplicates of an object as XML (`dedup:duplicates:{objectId}`, `dedup:duplicates-for-session:{sessionKey}`), the latter restricted to the session key prefixes in `MCR.URIResolver.ModuleResolver.dedup.SessionKeyPrefixes`.
* REST API under `/api/dedup` (`MCRDeDupResource`) listing possible duplicates and no-duplicate markings and creating or deleting markings.

**`mycore-dedup-gui`** (new component)
* Vue 3 based management interface served under `/dedup`, built with the frontend-maven-plugin like the other Vue based tools.
* Tab "Possible duplicates" lists the pairs found by the criteria and lets a user confirm a pair as two different publications. Tab "Confirmed non duplicates" lists the existing markings and lets a user revoke them.

**`mycore-mods`**
* `MCRMODSIdentifierDeDupCriterionBuilder`, `MCRMODSTitleAuthorDeDupCriterionBuilder` and `MCRMODSShelfmarkDeDupCriterionBuilder` as the MODS criteria, plus `MCRMODSDeDupTitleResolver` for the display title.

## Design decisions

* Keys are stored instead of computed on demand so that finding duplicates is a database join and does not need to touch the metadata store.
* A no-duplicate marking is stored for the pair of ids, not on the object, so that revoking it in the GUI does not have to touch the metadata. The pair is normalized and the insert is idempotent, so marking the same pair twice does not fail.
* Criterion builders and title resolvers are configured per object type. `mycore-dedup` itself contains no MODS knowledge, the metadata model contributes it.

## How to test

1. Add `mycore-dedup` and `mycore-dedup-gui` to the application (MIR already has both in the branch mentioned below).
2. Import or create two MODS objects that share an identifier or title and author.
3. Open `/dedup`. The pair shows up under "Possible duplicates". Mark it, it moves to "Confirmed non duplicates". Delete the marking, it comes back.

---

## Ticket & Documentation

- [ ] New features or migrations are documented. (documentation in `mycore-website` is still missing)
  - [x] Yes: Is a migration required? If yes, describe it. (only for applications that add the new components: the tables `MCRDeDupKey`, `MCRDeDupNoDuplicate` and `MCRDeDupTitle` come from the JPA mapping, existing objects get their keys by running a repair of the metadata)
  - [ ] Breaking change is marked in the **commit message**. (no breaking change, all additions are new components)

## Bugfix-Specific Checks
Not applicable, this is a new feature.

## Testing
  - [ ] Yes: explain the changes for reviewers. (no existing test was modified, only new ones added)

## MCR Conventions & Metadata
  - [x] If not, no action needed. (nothing removed or changed, only new classes)
- [x] No default properties are hardcoded, all set via `mycore.properties`.

## Multi-Repo Considerations
  - [ ] If yes, is it already created? (MIR needs the import workflow that writes the no-duplicate flags, PR still to be opened)


mycore-dedup-gui/src/main/resources/components/dedup-gui/config/messages_de.properties
```diff
@@ -0,0 +1,24 @@
+dedup.gui.title=Dublettenverwaltung
+dedup.gui.tab.duplicates=Mögliche Dubletten
```
erodde, line 2: Umlauts okay here?

mycore-dedup/src/test/java/org/mycore/dedup/MCRDeDupKeyManagerTest.java
```diff
@@ -0,0 +1,309 @@
...
+import org.mycore.dedup.backend.MCRDeDupTitle;
+import org.mycore.test.MCRJPAExtension;
+import org.mycore.test.MyCoReTest;
+
+@MyCoReTest
+@ExtendWith(MCRJPAExtension.class)
+@MCRTestConfiguration(properties = {
+    @MCRTestProperty(key = "MCR.Metadata.Type.test", string = "true")
+})
+public class MCRDeDupKeyManagerTest {
```
erodde, line 45: 100% Test coverage, thats great! :)

mycore-dedup/src/main/java/org/mycore/dedup/MCRDeDupKeyManager.java
```diff
@@ -0,0 +1,362 @@
...
+                new MCRDeDupCriterion((String) row[1], (String) row[2])))
+            .distinct()
+            .toList();
+    }
+
+    /**
+     * Finds all existing objects that share at least one of the given criteria.
+     * <p>
+     * Criterion values are truncated to {@link MCRDeDupKey#MAX_VALUE_LENGTH} characters before
+     * matching, exactly as in {@link #storeKeys(MCRObjectID, Set)}.
```
erodde, line 226: Add in javadoc that noDuplicates are not considered here?

mycore-dedup/src/main/java/org/mycore/dedup/MCRDeDupKeyManager.java
```diff
@@ -0,0 +1,362 @@
...
+     *
+     * @param criteria the criteria to match, may be empty
+     * @return matching objects mapped to the criteria they share
+     */
+    public Map<MCRObjectID, Set<MCRDeDupCriterion>> findDuplicates(Set<MCRDeDupCriterion> criteria) {
+        List<MCRDeDupCriterion> truncated = new ArrayList<>(truncate(criteria));
+        if (truncated.isEmpty()) {
```
erodde, line 233: As truncation shouldn't change the amount of given criteria, the empty check and return could be directly executed onto the criteria as first method step.

mycore-dedup/src/main/java/org/mycore/dedup/MCRDeDupKeyManager.java
```diff
@@ -0,0 +1,362 @@
...
+            .getResultList();
+
+        Set<ObjectIdPair> noDuplicatePairs = loadNoDuplicatePairs(em);
+        List<MCRPossibleDuplicate> result = new ArrayList<>();
+
+        int index = 0;
+        while (index < rows.size()) {
+            String type = (String) rows.get(index)[1];
+            String value = (String) rows.get(index)[2];
+            List<String> members = new ArrayList<>();
```
erodde, line 289: rename `members` for better understandibility, for example `currentCriterionMembers`

mycore-dedup/src/main/java/org/mycore/dedup/MCRDeDupKeyManager.java
```diff
@@ -0,0 +1,362 @@
...
+     * Finds all pairs of objects that are possible duplicates of each other, i.e. that share at least
+     * one deduplication key and that are not marked as no-duplicates of each other.
+     * <p>
+     * Instead of a self-join over the whole key table this loads only the keys that actually have a
+     * matching partner (via an {@code EXISTS} semi-join) and groups them in memory, which avoids the
+     * combinatorial blow-up of a plain self-join.
+     *
+     * @return the possible duplicate pairs, one entry per pair and matching criterion
+     */
+    public List<MCRPossibleDuplicate> findAllDuplicates() {
```
erodde, line 272: This method was kind of hard to understand for me, especially the four loops. Is there a way to explain a bit per comments or improve readability?

mycore-dedup/src/main/java/org/mycore/dedup/resources/MCRDeDupResource.java
```diff
@@ -0,0 +1,157 @@
...
+    public static final String PERMISSION = "manage-deduplication";
+
+    private static final Gson GSON = new Gson();
+
+    @GET
+    @Path("duplicates")
+    @MCRCacheControl(noStore = true,
+        codeIsPrivate = @MCRCacheControl.FieldArgument(active = true),
+        noCache = @MCRCacheControl.FieldArgument(active = true))
+    public Response listAllDuplicates() {
```
erodde, line 77: Here and following methods:
1. should the endpoint have error handling? (not giving 500 errors back). Database operations can always lead to exceptions
2. Asking cause I honestly dont know at this point: Is it okay to call obtainInstance() every time or should instance be loaded once and reference kept here?

erodde (changes requested):
Only reviewed Java part, I am not able to review the Vue component

mycore-dedup-gui/src/main/resources/components/dedup-gui/config/messages_de.properties
sebhofmann: Changed to `\uXXXX` escapes, the file is us-ascii now.

The reason is how the bundles are loaded: `MCRCombinedResourceBundleControl` concatenates the `messages_de.properties` of all components into one stream via `MCRConfigurationInputStream` and hands that single stream to `PropertyResourceBundle`. The stream is decoded as UTF-8 with a fallback to ISO-8859-1 that applies to the whole stream. So one ISO-8859-1 encoded bundle in any other component or in an application overlay would have turned our UTF-8 umlauts into mojibake. All existing bundles use escapes as well.

(60e2a02e60)

mycore-dedup/src/test/java/org/mycore/dedup/MCRDeDupKeyManagerTest.java
sebhofmann: Thanks :) No change needed here.

mycore-dedup/src/main/java/org/mycore/dedup/MCRDeDupKeyManager.java
sebhofmann: Added. The javadoc of `findDuplicates(Set<MCRDeDupCriterion>)` now states that no-duplicate markings are not taken into account, as the criteria are not bound to a specific object: all objects sharing a criterion are returned.

(60e2a02e60)

mycore-dedup/src/main/java/org/mycore/dedup/MCRDeDupKeyManager.java
sebhofmann: Done, the empty check now runs on the given criteria as the first statement of the method, before the truncation.

(60e2a02e60)

mycore-dedup/src/main/java/org/mycore/dedup/MCRDeDupKeyManager.java
sebhofmann: Renamed to `currentCriterionMembers`.

(60e2a02e60)

mycore-dedup/src/main/java/org/mycore/dedup/MCRDeDupKeyManager.java
sebhofmann: Reworked. The four nested loops are now one loop over the criterion blocks plus three small helpers: `findEndOfCriterionBlock`, `addPairs` and `toCriterion`.

The javadoc now also names the idea behind it: the query orders the keys by criterion, so all objects sharing one criterion form one consecutive block of rows. Each block is read as a whole and then expanded into the pairs of its members.

(60e2a02e60)

mycore-dedup/src/main/java/org/mycore/dedup/resources/MCRDeDupResource.java
sebhofmann: 1. Added. `MCRDeDupApp` now registers `MCRExceptionMapper`, the same mapper `MCRRestV2App` uses. A database failure is genuinely a server error and stays a 500, but clients now get the machine readable error document of the REST API instead of a container error page. The Vue views already check for a status other than 200 resp. 201.

2. `obtainInstance()` per call is fine here. The manager is obtained through `MCRConfiguration2.getSingleInstanceOfOrThrow`, which caches the instance at configuration level, so every call returns the same object. Keeping a static reference in the resource would gain nothing and would break configuration reload and test isolation.

(60e2a02e60)

erodde (approved):
Java part is now fine.

mycore-mods/src/main/java/org/mycore/mods/dedup/MCRMODSIdentifierDeDupCriterionBuilder.java
```diff
@@ -0,0 +1,60 @@
...
+    /**
+     * Builds a deduplication criterion for a given identifier type and value. The value is normalized
+     * by removing hyphens so that for example {@code 978-1-56619-909-4} and {@code 9781566199094} match.
+     *
+     * @param type  the identifier type, e.g. {@code doi}, {@code isbn}, {@code issn}, {@code urn}
+     * @param value the identifier value
+     * @return the deduplication criterion
+     */
+    public MCRDeDupCriterion buildFromIdentifier(String type, String value) {
+        String normalizedValue = value.replace("-", "");
```
yagee-de: This normalization only makes sense for ISBN or ISSN where "-" is inserted for readability in URN or DOI "-" is part of the key and shouldn't be removed

PR #2972, opened 2026-06-08
MCR-3719 Transition property.xsl to a pure XSLT3 environment

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3719).

# Pull Request Checklist (Author)

Please go through the following checklist before assigning the PR for review:

## Ticket & Documentation
- [x] The issue in the ticket is clearly described and the solution is documented.
- [x] Design decisions (if any) are explained.
- [x] The ticket references the correct source and target branches.
- [x] The `fixed-version` is correctly set in the ticket and matches the PR's target branch (`main`).

## Feature & Improvement Specific Checks
- [ ] Instructions on how to test or use the feature are included or linked (e.g. to documentation).
- [ ] For UI changes: before & after screenshots are attached.
- [ ] New features or migrations are documented.
- [x] Does this change affect existing applications, data, or configurations?
  - [x] Yes: Is a migration required? If yes, describe it.
  - [x] Breaking change is marked in the **commit message**.

usage of `mcrproperty:one` needs to be replaced with `mcrproperty:get`
usage of `mcrproperty:all` needs to be replaced with `mcrproperty:map`

## Testing
- [x] I have tested the changes locally.
- [x] The feature behaves as described in the ticket.
- [ ] Were existing tests modified?
  - [ ] Yes: explain the changes for reviewers.

## MCR Conventions & Metadata
- [x] [MCR naming conventions](https://www.mycore.de/documentation/developer/conventions/) are followed
- [x] If the **public API** has changed:
  - [x] Old API is deprecated or a migration is documented.
  - [ ] If not, no action needed.
- [x] Java license headers are added where necessary.
- [x] Javadoc is written for non-self-explanatory classes/methods (Clean Code).
- [ ] All configuration options are documented in Javadoc and `mycore.properties`.
- [ ] No default properties are hardcoded — all set via `mycore.properties`.

## Multi-Repo Considerations
- [x] Is an equivalent PR in `MIR` required?
  - [x] If yes, is it already created?

**To be used in conjunction with https://github.com/MyCoRe-Org/mir/pull/1387.**

PR #3100, opened 2026-08-11
MCR-3719 (2) Transition property.xsl to a pure XSLT3 environment

toKrause:
[Link to jira](https://mycore.atlassian.net/browse/MCR-3719).

Small follow-up to https://github.com/MyCoRe-Org/mycore/pull/2972 and https://github.com/MyCoRe-Org/mycore/pull/3067.


rsteph-de (approved):
good improvement - is there doku, that needs an update?

PR #3131, opened 2026-09-07
MCR-3830 Share XSL transformer factories

yagee-de:
* Saxon becomes the explicit default, replacing automatic JAXP provider discovery. Default selection is now resolved at call time.
* Configuration uses named factory IDs (Saxon, Xalan, SlowXalan) instead of class names. Existing module configurations are migrated; class-based APIs and properties remain supported but deprecated.
* Factories can have individual configuration, including an optional Saxon XML configuration file.
* Concurrency handling is explicit: factory operations are serialized for Xalan by default, while Saxon permits concurrent access.
* Transformer caching is corrected: XML transformer cache keys now include the factory, stylesheet combinations cannot collide, and concurrent cache misses avoid duplicate creation.
* Template compilation stops changing the JVM-wide JAXP property and uses the configured Xalan factory.
* Regression tests cover concurrency, configuration, legacy compatibility, caching, and XEditor processor selection.

[Link to jira](https://mycore.atlassian.net/browse/MCR-3830).

## Ticket & Documentation

## Bugfix-Specific Checks
- [ ] Affected version is listed in the ticket.
- [ ] This PR truly fixes **only** the reported bug.

## Testing

## MCR Conventions & Metadata
  - [x] Old API is deprecated or a migration is documented.

## Multi-Repo Considerations

### DIFF
--- a/mycore-base/src/main/java/org/mycore/common/xsl/MCRLegacyTransformerFactorySupport.java
+++ b/mycore-base/src/main/java/org/mycore/common/xsl/MCRLegacyTransformerFactorySupport.java
@@ -1,0 +1,121 @@
+/*
+ * This file is part of ***  M y C o R e  ***
+ * See https://www.mycore.de/ for details.
+ *
+ * MyCoRe is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * MyCoRe is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with MyCoRe.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.mycore.common.xsl;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
+import javax.xml.transform.TransformerFactory;
+
+import org.mycore.common.MCRClassTools;
+import org.mycore.common.config.MCRConfiguration2;
+
+/**
+ * Compatibility support for class-based transformer factory selection.
+ * <p>
+ * Resolves legacy classes to a unique configured ID, preferring exact matches over subclasses.
+ * Classes without a unique match receive an internal {@code legacy:} ID. Their factories are created
+ * lazily, shared within the owning registry and always accessed serially.
+ * The class-to-ID registration is independent of registry initialization so legacy properties can be
+ * resolved while the registry is being constructed.
+ */
+final class MCRLegacyTransformerFactorySupport {
+
+    private static final String CLASS_PROPERTY_SUFFIX = ".Class";
+
+    private static final String LEGACY_FACTORY_ID_PREFIX = "legacy:";
+
+    private static final ConcurrentMap<String, Class<? extends TransformerFactory>> LEGACY_FACTORY_CLASSES =
+        new ConcurrentHashMap<>();
+
+    private final ConcurrentMap<String, MCRSAXTransformerFactoryManager> factories = new ConcurrentHashMap<>();
+
+    /**
+     * Returns the shared legacy factory, or {@code null} if the ID has not been registered for a class.
+     */
+    MCRSAXTransformerFactoryManager get(String id) {
+        Class<? extends TransformerFactory> factoryClass = LEGACY_FACTORY_CLASSES.get(id);
+        return factoryClass == null ? null : factories.computeIfAbsent(id, _ -> createLegacyFactory(id, factoryClass));
+    }
+
+    static String getFactoryId(Class<? extends TransformerFactory> factoryClass,
+        Map<String, MCRSAXTransformerFactoryManager> configuredFactories) {
+        Map<String, Class<? extends TransformerFactory>> factoryClasses = new HashMap<>();
+        configuredFactories.forEach((id, factory) -> factoryClasses.put(id, factory.getFactoryClass()));
+        return resolveFactoryId(factoryClasses, factoryClass);
+    }
+
+    /**
+     * Resolves an ID directly from properties without initializing the shared registry.
+     * This keeps legacy-property migration out of the registry's construction path.
+     */
+    static String getFactoryId(Class<? extends TransformerFactory> factoryClass) {
+        String propertyPrefix = MCRTransformerFactoryRegistry.CONFIGURATION_PREFIX + ".";
+        Map<String, Class<? extends TransformerFactory>> factoryClasses = new HashMap<>();
+        MCRConfiguration2.getSubpropertiesMap(propertyPrefix).keySet().stream()
+            .filter(key -> key.endsWith(CLASS_PROPERTY_SUFFIX))
+            .filter(key -> key.indexOf('.') == key.lastIndexOf('.'))
+            .map(key -> key.substring(0, key.length() - CLASS_PROPERTY_SUFFIX.length()))
+            .forEach(id -> MCRConfiguration2.<TransformerFactory>
+                getClass(propertyPrefix + id + CLASS_PROPERTY_SUFFIX)
+                .ifPresent(configuredClass -> factoryClasses.put(id, configuredClass)));
+        return resolveFactoryId(factoryClasses, factoryClass);
+    }
+
+    private static String resolveFactoryId(Map<String, Class<? extends TransformerFactory>> factoryClasses,
+        Class<? extends TransformerFactory> factoryClass) {
+        List<String> exactMatches = factoryClasses.entrySet().stream()
+            .filter(entry -> entry.getValue().equals(factoryClass))
+            .map(Map.Entry::getKey)
+            .toList();
+        if (!exactMatches.isEmpty()) {
+            return getConfiguredOrLegacyId(exactMatches, factoryClass);
+        }
+        List<String> assignableMatches = factoryClasses.entrySet().stream()
+            .filter(entry -> factoryClass.isAssignableFrom(entry.getValue()))
+            .map(Map.Entry::getKey)
+            .toList();
+        return getConfiguredOrLegacyId(assignableMatches, factoryClass);
+    }
+
+    private static String getConfiguredOrLegacyId(List<String> matches,
+        Class<? extends TransformerFactory> factoryClass) {
+        if (matches.size() == 1) {
+            return matches.getFirst();
+        }
+        String legacyFactoryId = LEGACY_FACTORY_ID_PREFIX + factoryClass.getName();
+        LEGACY_FACTORY_CLASSES.putIfAbsent(legacyFactoryId, factoryClass);
+        return legacyFactoryId;
+    }
+
+    static boolean isLegacyFactoryId(String factoryId) {
+        return factoryId.startsWith(LEGACY_FACTORY_ID_PREFIX);
+    }
+
+    private MCRSAXTransformerFactoryManager createLegacyFactory(String id,
+        Class<? extends TransformerFactory> factoryClass) {
+        TransformerFactory factory = TransformerFactory.newInstance(factoryClass.getName(),
+            MCRClassTools.getClassLoader());
+        return new MCRSAXTransformerFactoryManager(id, factory, true);
+    }
+
+}
--- a/mycore-base/src/main/java/org/mycore/common/xsl/MCRLegacyXSLCheck.java
+++ b/mycore-base/src/main/java/org/mycore/common/xsl/MCRLegacyXSLCheck.java
@@ -1,0 +1,80 @@
+/*
+ * This file is part of ***  M y C o R e  ***
+ * See https://www.mycore.de/ for details.
+ *
+ * MyCoRe is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * MyCoRe is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with MyCoRe.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.mycore.common.xsl;
+
+import java.util.Map;
+import java.util.stream.Collectors;
+
+import org.mycore.common.config.MCRConfiguration2;
+import org.mycore.common.config.MCRConfigurationException;
+import org.mycore.common.events.MCRStartupHandler.AutoExecutable;
+
+import jakarta.servlet.ServletContext;
+
+/**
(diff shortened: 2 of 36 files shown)


mycore-base/src/main/java/org/mycore/common/xsl/MCRTransformerFactorySelector.java
```diff
@@ -0,0 +1,118 @@
...
+     */
+    public static String getFactoryId(String property, String legacyProperty, String defaultId) {
+        return getConfiguredFactoryId(property, legacyProperty).orElse(defaultId);
+    }
+
+    private static Optional<String> getConfiguredFactoryId(String property, String legacyProperty) {
+        return MCRConfiguration2.getString(property).map(MCRTransformerFactorySelector::getFactoryId)
+            .or(() -> MCRConfiguration2
+                .<TransformerFactory>getClass(legacyProperty)
+                .map(factoryClass -> migrateLegacyProperty(property, legacyProperty, factoryClass)));
```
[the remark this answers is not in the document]
yagee-de: addressed in https://github.com/MyCoRe-Org/mycore/pull/3131/commits/5467491f1fd18f884d8e3cc9fb181cb714d2ed09

mycore-base/src/main/java/org/mycore/common/xsl/MCRSAXTransformerFactoryManager.java
```diff
@@ -0,0 +1,234 @@
...
+
+    MCRSAXTransformerFactoryManager(String id, TransformerFactory factory, boolean serializeAccess) {
+        this(id, factory, serializeAccess, MCRURIResolver::obtainInstance);
+    }
+
+    MCRSAXTransformerFactoryManager(String id, TransformerFactory factory, boolean serializeAccess,
+        Supplier<URIResolver> uriResolverSupplier) {
+        this.id = Objects.requireNonNull(id);
+        Objects.requireNonNull(factory);
+        if (factory instanceof SAXTransformerFactory saxTransformerFactory) {
```
toKrause, line 76: If only `SAXTransformerFactory` instances are allowed, the constructor should take a `SAXTransformerFactory` and not a `TransformerFactory`

mycore-base/src/main/java/org/mycore/common/xsl/MCRSAXTransformerFactoryManager.java
```diff
@@ -0,0 +1,234 @@
...
+import org.mycore.common.xsl.uriresolver.MCRURIResolver;
+
+/**
+ * Manages access to one application-scoped {@link SAXTransformerFactory}.
+ * <p>
+ * JAXP does not guarantee that a factory supports concurrent operations. Access can therefore be serialized for
+ * providers that require it, while the returned {@link Templates}, {@link Transformer}, and
+ * {@link TransformerHandler} instances can follow their individual JAXP lifecycle rules.
+ */
+public final class MCRSAXTransformerFactoryManager {
```
toKrause, line 48: Naming: The name suggests, correctly, that this class works only with `SAXTransformerFactory` instances. If that is so important, that it is communicated in the class name, all the supporting classes (`...FactoryRegistry`, `...FactorySelector`, `...Legacy...Support`) should have the `MCRSAX...` prefix. If it not that important, this class should be renamed to `MCRTransformerFactoryManager`

mycore-base/src/main/java/org/mycore/common/xsl/MCRLegacyTransformerFactorySupport.java
```diff
@@ -0,0 +1,121 @@
...
+/**
+ * Compatibility support for class-based transformer factory selection.
+ * <p>
+ * Resolves legacy classes to a unique configured ID, preferring exact matches over subclasses.
+ * Classes without a unique match receive an internal {@code legacy:} ID. Their factories are created
+ * lazily, shared within the owning registry and always accessed serially.
+ * The class-to-ID registration is independent of registry initialization so legacy properties can be
+ * resolved while the registry is being constructed.
+ */
+final class MCRLegacyTransformerFactorySupport {
```
toKrause, line 41: I do not like this class name. It is very abstruse. It is not a simple "helper" or "utility class". The ultimate purpose of this class seems to be to provide legacy transformer factories that are not registered in the transformer registry. I suggest `MCRLegacyTransformerFactoryProvider`.

mycore-base/src/main/java/org/mycore/common/xsl/MCRLegacyTransformerFactorySupport.java
```diff
@@ -0,0 +1,121 @@
...
+import org.mycore.common.MCRClassTools;
+import org.mycore.common.config.MCRConfiguration2;
+
+/**
+ * Compatibility support for class-based transformer factory selection.
+ * <p>
+ * Resolves legacy classes to a unique configured ID, preferring exact matches over subclasses.
+ * Classes without a unique match receive an internal {@code legacy:} ID. Their factories are created
+ * lazily, shared within the owning registry and always accessed serially.
+ * The class-to-ID registration is independent of registry initialization so legacy properties can be
```
toKrause, line 38: class-to-ID-mapping might be a bit easier to understand

mycore-base/src/main/java/org/mycore/common/xsl/MCRTransformerFactoryRegistry.java
```diff
@@ -0,0 +1,122 @@
...
+ * {@code MCR.TransformerFactory.{id}.Configuration.Class} is applied before the common MyCoRe configuration.
+ * Access to a factory is serialized by default and can be disabled with
+ * {@code MCR.TransformerFactory.{id}.SerializeAccess=false} for thread-safe providers.
+ * A factory ID is one property-name segment and therefore must not contain a dot.
+ */
+public final class MCRTransformerFactoryRegistry {
+
+    public static final String CONFIGURATION_PREFIX = "MCR.TransformerFactory";
+
+    private static final String SERIALIZE_ACCESS_PROPERTY = "SerializeAccess";
```
toKrause, line 45: I do not like this property name. While it is technically correct, it might not be as easy to comprehend for casual readers. May I suggest `SupportsConcurrency`.

mycore-base/src/main/java/org/mycore/common/xsl/MCRTransformerFactoryRegistry.java
```diff
@@ -0,0 +1,122 @@
...
+
+    /**
+     * Resolves a legacy class to a configured or internal legacy factory.
+     */
+    MCRSAXTransformerFactoryManager get(Class<? extends TransformerFactory> factoryClass) {
+        return get(MCRLegacyTransformerFactorySupport.getFactoryId(factoryClass, factories));
+    }
+
+    @SuppressWarnings("unchecked")
+    private void applyConfiguration(String configurationPrefix, String id, TransformerFactory factory) {
```
toKrause, line 108: Users may need to set additional factory attributes or may need to perform other operations on the factory, as well as setting the configuration file. It would be better, if we could have a list of "configurations", not just a single configuration.

mycore-base/src/main/resources/config/mycore.properties
```diff
@@ -52,6 +52,16 @@ SLOWXALAN=org.apache.xalan.processor.TransformerFactoryImpl
 XALAN=org.mycore.common.xsl.MCRXalanTransformerFactory
 SAXON=net.sf.saxon.TransformerFactoryImpl
 
+MCR.TransformerFactory.Xalan.Class=%XALAN%
```
toKrause, line 55: We do not have a strict rule for this, and currently it is done very inconsistently, but I would suggest to follow this rule (an may be make it a design rule for MyCoRe):

Use  `TitleCase` for static configuration key segments and `lowerCamelCase` for variable configuration key segments, i.e.

```
MCR.TransformerFactory.xalan.Class=%XALAN%
MCR.TransformerFactory.xalan.SupportsConcurrency=false
...
```

mycore-base/src/main/java/org/mycore/common/xsl/MCRTransformerFactoryRegistry.java
```diff
@@ -0,0 +1,122 @@
...
+/**
+ * Registry of named, application-scoped JAXP transformer factories.
+ * <p>
+ * Entries are injected from {@code MCR.TransformerFactory.{id}.Class}. An optional consumer configured below
+ * {@code MCR.TransformerFactory.{id}.Configuration.Class} is applied before the common MyCoRe configuration.
+ * Access to a factory is serialized by default and can be disabled with
+ * {@code MCR.TransformerFactory.{id}.SerializeAccess=false} for thread-safe providers.
+ * A factory ID is one property-name segment and therefore must not contain a dot.
+ */
+public final class MCRTransformerFactoryRegistry {
```
toKrause, line 41: Better class name would be `MCRTransformerFactoryManagerRegistry`, because it ultimately holds a map of `MCRTransformerFactoryManager` instances,

mycore-base/src/main/java/org/mycore/common/xsl/MCRTransformerFactoryRegistry.java
```diff
@@ -0,0 +1,122 @@
...
+        this.configuredFactories = new HashMap<>(configuredFactories);
+    }
+
+    /**
+     * Applies the optional per-factory configuration and creates the shared holders.
+     *
+     * @param configurationPrefix canonical property prefix supplied by the MyCoRe instantiator
+     */
+    @MCRPostConstruction
+    public void initialize(String configurationPrefix) {
```
toKrause, line 69: This approach is a hard code smell. Classes that use annotation based configuration for some parts and then revert to manual extraction and parsing of configuration properties is exactly what we try to get rid of. This allows places like this to profit from future improvements to the configuration API.

In addition, the way it is done here has a problem, because it contaminates the property namespace of the instantiated class.

```
MCR.TransformerFactory.slowXalan.Class=org.mycore.common.xsl.MCRXalanTransformerFactory
MCR.TransformerFactory.slowXalan.SerializeAccess=true
...
```

This reads as if `SerializeAccess` is a property implemented in `org.mycore.common.xsl.MCRXalanTransformerFactory`, but that is not the case here. The property is implemented in `MCRTransformerFactoryRegistry`.

We should not smear the property bounderies in this way.

`SerializeAccess` should, normally, be a common property of all potential implementations, i.e., implemented in a common abstract base class.

This case is a bit special though, because the instances we want are not part of the MyCoRe code base.

So, on one hand:
- we want to inject instances of `SAXTransformerFactory`, that we do not necessarily control (the `TransformerFactoryImpl` classes of Saxon and Xalan)
- we can not create an abstract base class for (and in case of Saxon, can not create an inherited class of because of licensing reasons)

On the other hand:
- we want to provide common behavior (like the `SerializeAccess` and `Configuration`)


I suggest adding a small class implicit for registry entries, that holds the actual factory and all common properties that are used to influence or modify the otherwise externally controlled factory implementations. 

The only cost is, that the property names would be a bit less sleek, but I think that that is a price well worth paying.

Instead of
```
MCR.TransformerFactory.xalan.Class=%XALAN%
```
we would have sth. like 
```
MCR.TransformerFactory.xalan.Factory.Class=%XALAN%
```
This obviously reads and sounds wrong, but if we adjust the prefix, this can be countered, for example:
```
MCR.TransformerRegistry.xalan.Factory.Class=%XALAN%
```

toKrause (changes requested):

mycore-base/src/main/resources/config/mycore.properties
```diff
@@ -52,6 +52,16 @@ SLOWXALAN=org.apache.xalan.processor.TransformerFactoryImpl
 XALAN=org.mycore.common.xsl.MCRXalanTransformerFactory
 SAXON=net.sf.saxon.TransformerFactoryImpl
```
toKrause, line 53: Since the need for these short hand notations is on the way out, I would swap the definitions ad keep all relevant information in the properties below, i.e.:

```
MCR.TransformerRegistry.Entries.xalan.Factory.Class=org.mycore.common.xsl.MCRXalanTransformerFactory

...

XALAN=%MCR.TransformerRegistry.Entries.xalan.Factory.Class%
```

The shorthands can then be removed along with all the deprecated code in `2027.06`.

toKrause:
@yagee-de I've created a [non-polished branch with the suggestions](https://github.com/MyCoRe-Org/mycore/tree/issues/MCR-3830-share-transformer-factories-SUGGESTIONS), esp. a rework of `MCRTransformerFactoryManagerRegistry`

mycore-base/src/main/java/org/mycore/common/xsl/MCRTransformerFactoryRegistry.java
```diff
@@ -0,0 +1,122 @@
...
+    private Map<String, MCRSAXTransformerFactoryManager> factories = Map.of();
+
+    private final MCRLegacyTransformerFactorySupport legacyFactories = new MCRLegacyTransformerFactorySupport();
+
+    /**
+     * Receives the transformer factories configured directly below {@link #CONFIGURATION_PREFIX}.
+     *
+     * @param configuredFactories transformer factories by configured ID
+     */
+    @MCRInstanceMap(valueClass = TransformerFactory.class)
```
toKrause, line 58: One more thought:

While `MCRInstanceMap` supports configurations without a `name` attribute, that design decision should be approached with caution, because it prevents the registry from having any other configurable properties in the future.

Might be better do add a name and go for, for example:
```
MCR.TransformerRegistry.Entries.xalan.Factory.Class=%XALAN%
```

As with the `.Factory.Class`-suffix, the trade-off here is sleek property names vs. composability and future extensibility.

mycore-base/src/main/java/org/mycore/common/xsl/MCRLegacyXSLCheck.java
```diff
@@ -0,0 +1,80 @@
...
+        String conflicts = properties.keySet().stream()
+            .filter(key -> key.endsWith(LEGACY_SUFFIX))
+            .filter(key -> properties.containsKey(currentProperty(key)))
+            .sorted()
+            .map(key -> migrationMessage(key, properties))
+            .collect(Collectors.joining("\n\n"));
+        if (!conflicts.isEmpty()) {
+            throw new MCRConfigurationException("Conflicting legacy XSL transformer factory properties. "
+                + "Migrate the following settings before starting the application:\n\n" + conflicts
+                + "\n\nFactory IDs are configured below '" + MCRTransformerFactoryRegistry.CONFIGURATION_PREFIX
```
toKrause, line 63: For hierarchical structures (like the propriety names), "configured under" is, afaik, more idiomatic.

toKrause:
I've had a another thought, ....

One thing that made it quite hard to understand what is going on, is that `MCRXSLTransformerFactoryManager` creates and holds on to an instance of the registry which in tun creates instances of `MCRXSLTransformerFactoryManager`  and when calling `MCRXSLTransformerFactoryManager#obtainInstance` the registry is used to perform that operation.

Also, the `MCRXSLTransformerFactoryManager` only provides methods that are provided by `XSLTransformerFactory`, not the factories itself. Typically, in MyCoRe a XXX-manager provides access to XXX instances.

Some suggestions:

- Iif `MCRXSLTransformerFactoryManager` is a facade for `XSLTransformerFactory` instances, why not call it just `MCRTransformerFactory`. That's pretty simple and concise and avoids the dreaded `FactoryManager`-suffix.
- If the registry is the object that is actually configured, let it be the instance that is obtained.
  - Drop the instance holder in the factory manager and just use `MCRConfiguration2#getSingleInstanceOfOrThrow` in a static `obtainInstance` method
     - The composition than feels very natural and is very similar to other classes in MyCoRe
     -  My suggested naming theme for the properties (`MCR.TransformerRegistry....`) than fits pretty well. 
     - We can still provide convenience accessors for (shared) factories in `MCRTransformerFactory`
  - Let the registry have public methods that provide the registered shared factories or legacy factories that are created on the fly 
  - Let the method names reflect if a shared factory is returned or not.

toKrause:
I've updated the suggestion branch accordingly

